PR #18, opened 2018-03-13
added ability to build platform specific wheels

OmegaDroid:
This should solve the wheel building issue.

It should also support multiple distribution platforms but will need to be build on each.

To publish the source dist run `python setup.py publish --sdist`.
To publish the wheel run `python setup.py publish --wheel` on the specific platform.

If there isn't a wheel available for the specific platform it should revert to the source dist which will compile ktools on install (assuming all the build dependencies are present).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,10 @@ Alternatively you can install the latest development version using::
 
 (over SSH).
 
+You can also install from a specific branch ``<branch name>`` using::
+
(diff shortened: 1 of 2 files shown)


sr-murthy:
@mpinkerton-oasis I've added package publication instructions to the repository README

https://github.com/OasisLMF/OasisLMF#publishing

PR #35, opened 2018-04-24
Add concurrency utils module

sr-murthy:
The main contributions are:

* `Task` class to represent tasks as `(<function>, <args>, <result key>)` triples
* a thread signal handler
* a concurrent task execution method `aggregate` - accepts an iterable of `Task` objects, with an optional `pool_size` parameter for setting the size of the thread pool (by default equal to the number of tasks)

### DIFF
--- a/oasislmf/utils/concurrency.py
+++ b/oasislmf/utils/concurrency.py
@@ -1,0 +1,135 @@
+# -*- coding: utf-8 -*-
+
+import queue
+import signal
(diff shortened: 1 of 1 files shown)


sr-murthy:
[Add concurrency utils to MDK (`dev`)](https://trello.com/c/KYCFgyqZ/377-add-concurrency-utils-to-mdk-dev)

PR #59, opened 2018-07-06
Fix/ktools install

sambles:


setup.py
```diff
@@ -101,14 +102,30 @@ def unpack_tar(self, tar_location, extract_location):
...
+        for ktools_bin in ktools_bin_subset:
+            if find_executable(ktools_bin) is None:
+                return False
+        return True        
+
     def build_ktools(self, extract_location):
         self.announce('Building ktools', INFO)
         build_dir = os.path.join(extract_location, 'ktools-OASIS_{}'.format(KTOOLS_VERSION))
 
         exit_code = os.system('cd {build_dir} && ./autogen.sh && ./configure && make && make check'.format(build_dir=build_dir))
```
sr-murthy: This is not a request, just a comment but it's probably better to replace `os.system` by `subprocess.check_call` (https://docs.python.org/2/library/subprocess.html#subprocess.check_call), which will raise `subprocess.CalledProcessError` if the return code is non-zero - the return code will be the value of the attribute `returncode` in the error object. And all this can be wrapped in a `try`-`except` block. So perhaps line 123 could be replaced by the following:

    try:
        subprocess.check_call(['cd', build_dir, '&&', './autogen.sh', '&&', './configure', '&&', 'make', '&&', 'make', 'check'])
    except subprocess.CalledProcessError as e:
        # as you're in the `except` block you know the return code for the attempted command was non-zero
        self.announce('Ktools build failed.\n', WARN)
        if (not self.ktools_inpath()):
            ...
            sys.exit(e.returncode)
        
    # after `try`-`except` block
    return build_dir

sr-murthy: Actually instead of `sys.exit(e.returncode)` inside the `except` block, you can just raise an `OasisException`(from `oasislmf.utils.exceptions`), which will interrupt the original command, i..e

    try:
        ...
    except subprocess.CalledProcessError as e:
        ...
        if not self.ktools_inpath():
            ...
            raise OasisException(e)

    return build_dir

PR #60, opened 2018-07-06
[WIP] Feature/csv multiprocces

sambles:


oasislmf/cmd/model.py
```diff
@@ -212,7 +212,7 @@ def action(self, args):
 
         self.logger.info('Generating a canonical {} file {} from source {} file {}'.format(_sft, output_file_path, _sft, source_file_path))
 
-        translator = Translator(source_file_path, output_file_path, xsd_validation_file_path, xslt_transformation_file_path, append_row_nums=True)
+        translator = Translator(source_file_path, output_file_path, xslt_transformation_file_path, xsd_validation_file_path, append_row_nums=True)
```
sr-murthy: Looks like the only difference here is you've swapped the order of the transformation and validation file path arguments?

oasislmf/cmd/model.py
```diff
@@ -280,7 +280,7 @@ def action(self, args):
 
         self.logger.info('Generating a model exposures file {} from canonical exposures file {}'.format(output_file_path, canonical_exposures_file_path))
 
-        translator = Translator(canonical_exposures_file_path, output_file_path, xsd_validation_file_path, xslt_transformation_file_path, append_row_nums=True)
+        translator = Translator(canonical_exposures_file_path, output_file_path, xslt_transformation_file_path, xsd_validation_file_path ,append_row_nums=True)
```
sr-murthy: Same as above

oasislmf/exposures/csv_trans.py
```diff
@@ -6,65 +6,132 @@
...
     def __call__(self):
         csv_reader = pd.read_csv(self.fpath_input, iterator=True, dtype=object, encoding='utf-8')
-        for data, first_row_number, last_row_number in self.next_file_slice(csv_reader):
-            self.logger.debug('--- lines[%d .. %d] ---', first_row_number, last_row_number)
 
-            # Convert CSV -> XML
-            xml_input_slice = self.csv_to_xml(
-                self.row_header_in,
-                data
+        task_list = []
```
sr-murthy: Just an observation but as the chunk processing doesn't seem to involve any filesystem access you could use `multiprocess` which uses processes rather than `multithread` which uses plain threads. It'll probably be quicker.

oasislmf/exposures/csv_trans.py
```diff
@@ -6,65 +6,132 @@
...
+        for data, first_row, last_row in self.next_file_slice(csv_reader):
+            task_list.append(Task(self.process_chunk, (data,first_row,last_row)))
+        # run is process file chunk
+        for task in multithread(task_list):
+            pass
+
+        # write output
```
sr-murthy: If you assigned IDs to the chunks (enumerating `Translator.next_file_slice` will give you a pair `(i, it)` where `i` is a counter value and `it` is the next item generated) you might be able to order the chunks using the chunk ID. So you could also use `multiprocess` to create tasks for writing the chunks in the output.

This is a just a suggestion.

PR #65, opened 2018-07-25
Feature/multi peril and coverage type model support

sr-murthy:
* Enable lookup framework and exposure manager to support multi-peril and multi-coverage type models


sr-murthy:
[Add multi-peril and multi-coverage type support to oasislmf (new lookup framework + exposure manager)](https://trello.com/c/qrf45bZn/433-add-multi-peril-and-multi-coverage-type-support-to-oasislmf-new-lookup-framework-exposure-manager)

PR #82, opened 2018-10-09
Fix/test timeout

sambles:


tests/exposures/test_csv_trans.py
```diff
@@ -20,6 +24,7 @@
 
 class CsvTrans(unittest.TestCase):
     @given(integers(min_value=1, max_value=10))
```
sr-murthy: This is just a suggestion but I think that any arguments in test cases whose data are generated by `hypothesis.given` but should be named, e.g.
    ...
    @given(chunk_size=integers(min_value=1, max_value=10))
    def test_source_to_canonical(self, chunk_size):
    ...

It is a bit easier to read the test case that way as you know which arguments are using which data generation strategy.

Also I would put the `settings` decorator at the top, above `given`.

tests/model_execution/test_bin.py
```diff
@@ -99,6 +103,7 @@ def test_subprocess_raises___oasis_exception_is_raised(self):
                     create_binary_files(csv_dir, bin_dir, do_il=True)
 
     @given(standard_input_files(min_size=1), il_input_files(min_size=1))
```
sr-murthy: Where are these custom methods `standard_input_files`, `il_input_files` defined? If you're going to use them in other places perhaps you could put them in the tests data generation module `tests/data.py` (assume that's part of the `master`, but it is certainly something I'm using in the FM PR extensively for the tests).

tests/exposures/test_csv_trans.py
```diff
@@ -20,6 +24,7 @@
...
+    @settings(deadline=800, suppress_health_check=[HealthCheck.too_slow])
```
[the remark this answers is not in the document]
sambles: done

tests/model_execution/test_bin.py
sambles: and done

PR #88, opened 2018-10-24
Feature/reinsurance tests

mpinkerton-oasis:
Moved over reinsurance tests from the ReinsuranceTestTool.


tests/exposures/test_reinsurance.py
```diff
@@ -0,0 +1,227 @@
...
+                losses_df = direct_layer.apply_fm(
+                    loss_percentage_of_tiv=loss_factor, net=False)
+                net_losses['Direct'] = losses_df
+
+                oed_validator = oed.OedValidator()
+                if do_reinsurance:
+                    (is_valid, error_msgs) = oed_validator.validate(ri_info_df, ri_scope_df)
+                    if not is_valid:
+                        print(error_msgs)
+                        exit(1)
```
sr-murthy: @mpinkerton-oasis Using `exit` here will result in the test process itself being stopped - the preferred outcome would be to skip the test.

Another problem with these tests is that in order for them to work the working directory must be `/path/to/OasisLMF/tests/exposures` - you are setting this via `os.cwd()`. If you ran the test module from `/path/to/OasisLMF` the tests all fail with that system exit, because the expected test output directory would be a subfolder of `/path/to/OasisLMF/`, e.g. `/path/to/OasisLMF/examples` etc.

The structure of this test module, including the hard coded test data and how it is used in the module, causes problems with running pytest and tox in the normal way, which is to run from the base of the repository.

I was going to merge in this branch to the FM branch, but because of these problems, which would prevent tox from passing all the other tests, I will leave it out.

sr-murthy:
@mpinkerton-oasis I ran all of the reinsurance tests again from `tests/exposures` and they all pass. It would be good to adjust the code in the `test_reinsurance.py` module so that the working directory path isn't set to `tests/exposures` but to two levels up, pointing to the repository base. That will ensure that these tests can be successfully run via pytest and tox.

sr-murthy:
Also, I would remove the use of `exit` and raising exceptions - as I mentioned yesterday, this just interrupts the whole test process, e.g. if you were using pytest or tox. To fail a test case all that is needed is to make the relevant assertion.

sr-murthy:
Note: as the `dev` branch was deleted (replaced by `develop`) a new PR from the tests branch needs to be created with `develop` as target.

PR #97, opened 2018-11-01
Fix/config json

sambles:


oasislmf/cmd/model.py
```diff
@@ -520,59 +520,68 @@ def action(self, args):
...
+        source_to_canonical_exposures_transformation_file_path = as_path(
+            inputs.get('source_to_canonical_exposures_transformation_file_path', required=True, is_path=True),
+            'Source to canonical exposures file transformation file path'
         )
-        source_exposures_file_path = as_path(inputs.get('source_exposures_file_path', required=True, is_path=True), 'Source exposures file path')
-        source_accounts_file_path = as_path(inputs.get('source_accounts_file_path', required=False, is_path=True), 'Source accounts file path')
         source_exposures_validation_file_path = as_path(
             inputs.get('source_exposures_validation_file_path', required=False, is_path=True),
             'Source exposures validation file',
             preexists=False
```
sr-murthy: The use of `preexists=False` is not intuitive here - the validation files are certainly not required (as they are optional for the transformation) but they do already exist. Does it not work if you leave it out? The point of `preexists` is to indicate that a certain file path, like for an output file, is created during the process of or is the result of running a command, e.g. the keys file path.

oasislmf/cmd/model.py
```diff
@@ -520,59 +520,68 @@ def action(self, args):
...
-            inputs.get('canonical_to_model_exposures_transformation_file_path', required=True, is_path=True),
-            'Canonical to model exposures transformation file path'
-        )
-
-        fm = inputs.get('fm', default=False)
-
         fm_agg_profile_path = as_path(
             inputs.get('fm_agg_profile_path', required=False, is_path=True),
-            'Supplier FM aggregation profile JSON file path'
+            'Supplier FM aggregation profile JSON file path',
```
sr-murthy: The aggregation profile is not tied to a supplier - this is just a command help message, but we can remove things like these in subsequent releases.

PR #105, opened 2018-11-07
Fix/gul items group ids

sr-murthy:
Set GUL items group ID to index item loc. ID


sr-murthy:
Not required - changes merged into release/1.2.1 PR

PR #117, opened 2018-11-23
API client Update

sambles:


oasislmf/api_client/session_manager.py
```diff
@@ -0,0 +1,150 @@
...
+    ## Connection Error Handler     
+    def __recoverable(self, error, url, request, counter=1):
+        if not (counter < self.retry_max):
+            self.logger.info("Max retries of '{}' reached".format(self.retry_max))
+            return False
+
+        if hasattr(error,'status_code'):
+            self.logger.debug("Connection Error Handler: {}".format(error.status_code))
+            if error.status_code in [502, 503, 504]:
+                error = "HTTP %s" % error.status_code
```
sr-murthy: I note you're using the old string formatting operator (`%`), but it may be best to use the new one `{}`. So here this would be

    error = "HTTP {}".format(error.status_code)

Same for all other places where you're using `%`.

oasislmf/cmd/api.py
```diff
@@ -0,0 +1,362 @@
...
+        '-C oasislmf.json'
+
+        3. Prompt for username / password
+    """
+    if isinstance(login_arg, (str, unicode)):
+        with io.open(login_arg, encoding='utf-8') as f:
+            return json.load(f)
+
+    elif isinstance(login_arg, (dict)):
+        if set(['password','username']) <= set(login_arg.keys()):
```
sr-murthy: On the left hand side of line 35 instead of 

    set(['password','username'])

it is more Pythonic to use

    {'password','username'}

Because the set elements are literals which you know beforehand.

oasislmf/cmd/api.py
```diff
@@ -0,0 +1,362 @@
...
+            return login_arg
+
+    else:
+        logger.info('No Login provided - Fallback to prompt')
+
+    try:
+        api_login = dict()
+        api_login['username'] = raw_input('Username: ')
```
sr-murthy: You're using `raw_input` which presents a user prompt, but I believe this only works for Python 2. I don't know this will still run on Python 3, where `input` replaces `raw_input`.

oasislmf/cmd/api.py
```diff
@@ -0,0 +1,362 @@
...
+
+    elif isinstance(login_arg, (dict)):
+        if set(['password','username']) <= set(login_arg.keys()):
+            return login_arg
+
+    else:
+        logger.info('No Login provided - Fallback to prompt')
+
+    try:
+        api_login = dict()
```
sr-murthy: `{}` is more Pythonic than `dict()`. Same for `[]` vs `list()`, although sometimes it seems simpler to use `list` when converting an iterable with unknown items to a list.

PR #125, opened 2018-12-03
Fix/api client

sambles:


oasislmf/cmd/api.py
```diff
@@ -42,14 +42,32 @@ def load_credentials(login_arg, logger=None):
 
     try:
         api_login = {}
-        api_login['username'] = six.input('Username: ')
+        api_login['username'] = six.moves.input('Username: ')
```
sr-murthy: It is probably better style to have an import like
   
    ...
    from six.moves import input
    ...

and then have

    ...
    api_login['username'] = input('Username: ')
    ...

But it is fine from a functional POV.

oasislmf/cmd/api.py
```diff
@@ -42,14 +42,32 @@ def load_credentials(login_arg, logger=None):
...
         api_login['password'] = getpass.getpass('Password: ')
         return api_login
     except KeyboardInterrupt as e:
         logger.info('\nFailed to get API login details:')
         sys.exit(1)
 
 
+def open_api_connection(input_args, logger):
+    if not input_args.get('api_server_url'):
+        logger.info('Error: argument `--api-server-url` not set')
```
sr-murthy: May be `logger.error` might be more appropriate here. Same elsewhere where you're using `logger.info` to report errors/exceptions.

PR #126, opened 2018-12-03
Fail fast - test small sample of input before handover to multiprocess

sambles:


oasislmf/exposures/csv_trans.py
```diff
@@ -18,6 +18,8 @@
     multithread,
     Task,
 )
+from oasislmf.utils.exceptions import OasisException
```
sr-murthy: An absolute isn't required here - it's simpler to use a relative import here

    from ..utils.exceptions import OasisException

Issue #204, opened 2019-03-11
get_gul_input_items uses incorrect case for columns, resulting in dtype mismatches


DanielFEvans:
Running `oasislmf model run ...` in v1.3.2, I receive the following error at line 113 in `oasislmf.model_preparation.gul_inputs.get_gul_input_items`:

`2019-03-11 16:45:40,167 - root - ERROR - You are trying to merge on int64 and object columns. If you wish to proceed you should use pd.concat`

This is because the `locnumber` column in `exposure_df` and `locid` column in `keys_df` are different, despite the calls to `get_dataframe` specifying datatypes for the columns. The issue arises because OasisLMF has converted the OED column names to all-lowercase at an earlier stage, but the calls to `get_dataframe` refer to them using cancelcase. Therefore, it is trying to convert the datatypes of the nonexistent `LocID` and `LocNumber` columns, rather than `locid` and `locnumber`.

(There's a lot of mixing of lowercase and camelcase in OasisLMF, so there may well be other, similar bugs hiding)

sr-murthy:
@DanielFEvans There is no error here - that is a Pandas warning, and it is unrelated to any column casing issues. Execution should not be interrupted, but complete normally.

Which version of Python are you running the package in?

DanielFEvans:
Python 2.7.5, with Pandas 0.24.1 (as required by OasisLMF). I think this only became an error more recently in Pandas (I think 0.23).

sr-murthy:
@DanielFEvans That is relevant, because we know that there is a Pandas bug with 0.23.x

https://github.com/pandas-dev/pandas/issues/23733

which is why we upgraded the package to use Pandas to `0.24.x`. In fact, it should work with Pandas 0.22.0 as well, but Pandas `0.24.1` gets installed with the MDK.

We did not see any runtime errors in the CI testing phase with either Python 3.5+ or Python 2.7.

sr-murthy:
@DanielFEvans It does look like the column datatypes setting step in `get_dataframe` will fail to take effect if the column case used by the caller does not match any lowercasing of the frame columns in `get_dataframe` as a result of `lowercase_cols=True`. That logic was correct prior to release 1.3.1, but  some changes were made by other developers, and as a result the column datatypes setting step is incorrect.

[...]

PR #207, opened 2019-03-12
Hotfix - Fix column data types settting in Pandas frame getter utils method

sr-murthy:
* Fix column data types settting in Pandas frame getter utils method

https://github.com/OasisLMF/OasisLMF/issues/204


sambles:
cannot publish from PR, closing until after

PR #225, opened 2019-04-03
Further test fixes for Windows

DanielFEvans:
Fixes in the updated test_data.py.

Note that Windows does not play nice with dates after the year 3000 - the `test_get_utctimestamp` would fail with an OSError when Python called through to the OS-level datetime functions.


DanielFEvans:
The issue is that Windows does not allow a file to be opened simultaneously for write and read through separate handles. A lot of the tests create a temporary file in write mode, place some data in it, and then keep the write handle open when running the code that reads this file, resulting in an OSError because Windows will not allow the file to be opened twice.

sr-murthy:
OK that's fine. This has also passed the CI checks, so will merge shortly.

PR #226, opened 2019-04-04
Complex lookup config

DanielFEvans:
Add command line option for lookup to pass in a `complex-lookup-config-file-path` (optional, default None), and pass this through to the model lookup instance.


sr-murthy:
@DanielFEvans Why is a separate argument for a complex lookup config. file required in the model subcommands?

DanielFEvans:
We require our lookup to be configurable. The current OasisLMF/API does not provide any settings file to the lookup stage - the lookup interface only passes the keys data directory and the model version information. Slack has quite a few discussions on the topic (and I think it's also been discussed separately with Mark Pinkerton).

sr-murthy:
OK, it has passed CI checks.

Issue #205, opened 2019-03-12
OasisLMF reports success even if model gulcalc or ktools calls fail


DanielFEvans:
Running a "Complex Model" via `oasislmf model run...` with our own gulcalc.py in OasisLMF v1.3.2, our gulcalc.py exited with an exception.

Despite a long stream of exceptions and errors that followed from the various ktools components, the final log output is:
`Model run completed successfully in /home/danielevans/run_model_2/runs/ProgOasis-20190312103326`

Should OasisLMF be checking that the calls to `gulcalc.py` and the ktools script completed successfully, and that any outputs were actually generated?

DanielFEvans:
Fixed by #230

PR #230, opened 2019-04-05
Improvements to error logging, debugging, test runner

DanielFEvans:
A number of small improvements that make debugging, testing, etc. a bit easier:
- When running in Verbose mode, full exception tracebacks are logged/printed to screen, rather than just the exception message.
- Now that OasisLMF is Python 3 only, use the `raise ... from ...` syntax.
- In the bash script generation, use `set -e` to ensure that failures in the bash script (either by the custom loss module or ktools) cause a CalledProcessError to be raised. This fixes the strange behaviour where error messages from those commands would appear in the output, but OasisLMF would claim the run completed successfully (#205)
- Add `{posargs}` to tox.ini, which allows arguments to be passed through to the test runner.

With the last change, one can now run a specific file with:
`tox -- tests/some_test_file.py`
Or even a specific test with:
`tox -- tests/some_test_file.py::TestClass::test_task`

### DIFF
--- a/oasislmf/cli/base.py
+++ b/oasislmf/cli/base.py
@@ -102,6 +102,8 @@ class OasisBaseCommand(BaseCommand):
     """
     def __init__(self, *args, **kwargs):
         self._logger = None
+        self.args = None
+        self.log_verbose = False
         super(OasisBaseCommand, self).__init__(*args, **kwargs)
 
     def add_args(self, parser):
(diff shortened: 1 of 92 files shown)


sr-murthy:
@DanielFEvans The changes to verbose logging and exceptions look OK, but note that in relation to testing, the real test runner is pytest, not tox, and it is already possible to run specific test modules, test classes or test cases, e.g.

    python3.6 -m pytest -sp no:flaky -v tests/utils/test_data.py -x
    python3.6 -m pytest -sp no:flaky -v tests/utils/test_data.py::TestGetDataframe -x
    python3.6 -m pytest -sp no:flaky -v tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__no_src_fp_or_buf_or_data_provided__oasis_exception_is_raised

For local testing I use pytest, not tox. The CI calls tox, but for running the entire test suite - there is no need to run specific test modules or test classes in the CI jobs.

Given that we've now only targeting one Python version tox is less useful that it was originally, when there were three separate versions of Python we were targeting.

DanielFEvans:
If pytest is now the recommended test runner, then no problem. The readme is a little confusing, as it currently suggests three different ways to run the tests, with `tox` being the first suggestion, hence why I had been using it.

DanielFEvans:
Although all that the change to tox.ini does is to pass through any extra options to `pytest`; I'd be tempted to leave the change just so it's more flexible in general?

sr-murthy:
Yes, it's OK. I was pointing out that pytest has always been the default test runner. Tox isn't really a test runner, it just uses pytest underneath.

PR #238, opened 2019-04-11
Pass output directory to complex lookup to allow extra outputs

DanielFEvans:
As discussed with @sambles. Passes the output directory for the lookup step to the complex lookup class. This is to allow us to pass additional user outputs from the lookup stage, specifically a report on how aggregated portfolios were disaggregated, which will be of interest to our users. As the disaggregation is done at the lookup stage, the report needs to be passed through to the loss calculation stage. Our gulcalc code then handles copying it to the loss output directory, where it is tarred up with the loss outputs.

Tested with OasisPlatform, and the outputs appear in the final analysis outputs as required.


sr-murthy:
@DanielFEvans Can you please check whether the `generate-keys` command works as usual? One of your previous PRs which contained changes to the lookup factory broke the keys file generation command. We don't currently have unit tests for all the model subcommands, and the normal practice, even after all CI tests pass, is to runtime testing of the `generate-keys` subcommand.

We do have a runtime test script in CI that invokes `model run`, which in turn calls the `generate-oasis-files` and `generate-losses` commands, but there's nothing for `generate-keys`.

sr-murthy:
For this runtime testing of the command, you should `pip install` your branch, and then run the command.

sr-murthy:
@DanielFEvans The base branch `develop` has changed (and will change again), so please update it on the PR itself (Update branch button).

DanielFEvans:
Merged in the latest `develop`, and made some minor changes for the generate-keys command (thanks for the pointer).

sr-murthy:
@DanielFEvans Can you please whether you did a `pip install` of your branch and were able to run the `generate-keys` command on a model, e.g. JBA or PiWind, to successfully generate a keys file? This would the PR can be approved.

sr-murthy:
@DanielFEvans Can you please confirm whether you did a `pip install` of your branch and were able to run the `generate-keys` command on a model, e.g. JBA or PiWind, to successfully generate a keys file? This would the PR can be approved.

sr-murthy:
@DanielFEvans Can you please confirm whether you did a `pip install` of your branch and were able to run the `generate-keys` command on a model, e.g. JBA UKFL or PiWind, to successfully generate a keys file? This would be required before the PR can be approved.

DanielFEvans:
I've installed OasisLMF installed from my local branch via pip with:
`pip install .`

I've run the following command from the base directory of the Master branch of OasisPiWind:
`oasislmf model generate-keys -C oasislmf.json -x tests/data/SourceLocOEDNewPiWind10.csv`

The keys keys error (empty) files are generated without any errors.

I've also run it through a local instance of the API with a modified version of our 'Complex' model and received the expected output.

PR #248, opened 2019-04-18
Fix FM processing of single level special conditions

sr-murthy:
* include cond. num. in join keys for constructing initial IL items inputs frame
* related fixes in FM level terms main loop processing


sr-murthy:
FM12 acceptance test implementation required

PR #259, opened 2019-05-01
Feature/summary sets

sambles:


oasislmf/manager.py
```diff
@@ -440,26 +449,38 @@ def generate_oasis_files(
...
         # Write the RI input files, and write the returned RI layer info. as a
         # file, which can be reused by the model runner (in the model execution
         # stage) to set the number of RI iterations
+
+
+        ## TODO: pass 'fm_summary_mapping` inplace of `xref_descriptions`
+        # Example Merge in col from exposure
+
+
+        xref_des = merge_oed_to_mapping(fm_summary_mapping, exposure_df, ['locgroup'])
```
sr-murthy: `LocGroup` (or `locgroup`) is an OED column in the locations file - I think it would be better, and consistent with existing practice, to use a profile to refer to any columns in the source files, e.g. a summary profile JSON, stored in the static data folder, which you can load via a wrapper method in `utils/defaults.py` (as with the existing locations and accounts profiles).

oasislmf/model_preparation/gul_inputs.py
```diff
@@ -38,6 +37,9 @@
 from ..utils.exceptions import OasisException
 from ..utils.log import oasis_log
 from ..utils.defaults import (
+    COVERAGE_TYPES,
```
sr-murthy: Do you need `COVERAGE_TYPES`? The `SUPPORTED_COVERAGE_TYPES` defines the coverage types currently supported - buildings (1), other (2), contents (3), bi (4), and I think this would be sufficient.

oasislmf/model_preparation/gul_inputs.py
```diff
@@ -186,6 +188,7 @@ def get_gul_input_items(
         # keys dataframes on loc. number/loc. ID; filter out any rows with
         # zeros for TIVs for all coverage types, and replace any nulls in the
         # cond.num. and TIV columns with zeros
+        exposure_df[SOURCE_IDX['loc']] = exposure_df.index.values
```
sr-murthy: There's no need to specifically assign the index values (a Numpy array) - the index will do, e.g.

    exposure_df[SOURCE_IDX['loc']] = exposure_df.index

oasislmf/model_preparation/reinsurance_layer.py
```diff
@@ -212,8 +204,42 @@ def write_ri_input_files(
...
+    ri_info_fp,
+    ri_scope_fp,
+    target_dir
+):
+    old_xref_descriptions = pd.DataFrame(generate_xref_descriptions(accounts_fp, exposure_fp))
+    xref_descriptions = direct_mapping
+   
+    ## need to update col names in code --> tmp rename here
+    xref_descriptions['xref_id'] = xref_descriptions['output_id']
+    xref_descriptions['portfolio_number'] = xref_descriptions['portnumber']
```
sr-murthy: These direct references to OED columns in the code are not good practice, but this seems to be common in the RI code. It's not my implementation, or yours, so I would leave it.

oasislmf/utils/defaults.py
```diff
@@ -59,6 +64,47 @@
     if cov_type in ['buildings', 'other', 'contents', 'bi']
 })
 
+SOURCE_FILENAMES = OrderedDict({
```
sr-murthy: Looking at the `SOURCE_FILENAMES` dict:

    SOURCE_FILENAMES = OrderedDict({
        'loc': 'source_location.csv',
        'acc': 'source_account.csv',
        'info': 'source_info.csv',
        'scope': 'source_scope.csv'
    })

Is there is a reason to prefix the filenames with `source`? It is probably better to use the naming scheme for det. loss workflow

* locations file ->`location.csv`
* accounts file -> `account.csv`
* RI info file -> `reinsinfo.csv`
* RI scope file -> `reinsscope.csv`

oasislmf/utils/defaults.py
```diff
@@ -59,6 +64,47 @@
...
+    'loc': 'source_location.csv',
+    'acc': 'source_account.csv',
+    'info': 'source_info.csv',
+    'scope': 'source_scope.csv'
+})
+
+# Store index from merged source files (for later slice & dice)
+SOURCE_IDX = OrderedDict({
```
sr-murthy: I'm not clear on this dict - are these values, e.g.  `loc_idx` etc. column names in some file?

oasislmf/utils/defaults.py
```diff
@@ -59,6 +64,47 @@
...
+})
+
+SUMMARY_OUTPUT = OrderedDict({
+        'gul': 'gulsummaryxref.csv',
+        'il': 'fmsummaryxref.csv',
+})
+
+
+## Update with load OED column names 
+SUMMARY_GROUPING = OrderedDict({
```
sr-murthy: As I mentioned in the Slack PM, it would be better to create this dict by loading it from a static JSON profile, say a summary profile, that you store in the package static data. You can create a wrapper method to load it in `utils/defaults.py`. Column names of source files, or any intermediate files, should not be hard coded - this just introduces the possibility of runtime errors.

sr-murthy:
I note that you've added the `oasis_log` decorator to all the methods in the `sum_inputs.py` module, but I would only do this for the top-level or main methods in that module.

oasislmf/model_preparation/gul_inputs.py
sambles: I think that came from 53f9a1796a44d2a010265276acfd638134e8ae8f ?

oasislmf/model_preparation/reinsurance_layer.py
sambles: same as above, it should get replaced with a call to `merge_oed_to_mapping` and like you noted some external reference to OED columns

oasislmf/model_preparation/gul_inputs.py
sr-murthy: Yes, but that has changed since then - in the latest version of `gul_inputs.py` (in `develop`) only the supported coverage types dict is imported.

oasislmf/utils/defaults.py
sambles: column names for storing index of source files,

oasislmf/model_preparation/gul_inputs.py
sambles: okey it must of got added via the merge, i'll remove it

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -142,17 +103,12 @@ def merge_oed_to_mapping(summary_map_df, exposure_df, oed_column_set):
     :rtype: list
     """
 
-    # Guard check that set(column_set) < set(exposure_col_df.columns.to_list())
-    # --> OasisExecption if not
-
-    # Add or split `merge_oed_on_accounts`
     column_set = [c.lower() for c in oed_column_set]
     exposure_col_df = exposure_df[column_set]
```
sr-murthy: As discussed, it is better here to use

    exposure_col_df = exposure_df.loc[:, column_set]

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -167,21 +123,12 @@ def group_by_oed(summary_map_df, exposure_df, oed_col_group):
...
-
-
-    # -> Check is column_set subset of summary_map_df.columns.to_list()
-
-
-    # N -> import missing from OED
     exposure_cols = [c.lower() for c in oed_col_group if c not in summary_map_df.columns.to_list()]
     mapping_cols = [SOURCE_IDX['loc']] + [c.lower() for c in oed_col_group if c in summary_map_df.columns.to_list()]
 
     summary_group_df = summary_map_df[mapping_cols]
```
sr-murthy: Same as above - better to use

    summary_group_df = summary_map_df.loc[:, mapping_cols]

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -167,21 +123,12 @@ def group_by_oed(summary_map_df, exposure_df, oed_col_group):
...
     if not exposure_cols == []:
         exposure_col_df = exposure_df[exposure_cols]
```
sr-murthy: Same as above - better to use

    exposure_col_df = exposure_df.loc[:, exposure_cols]

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -231,32 +177,24 @@ def write_mapping_file(sum_inputs_df, target_dir):
     return sum_mapping_fp
 
 
-
-@oasis_log
 def get_column_selection(summary_set):
     # Summary Set Not defined in settings file
     if "oed_fields" not in summary_set.keys():
```
sr-murthy: As with Python dicts, you can check of the existence of a key in a dict by just using the dict, and not the dict keys e.g. 

    if "oed_fields" not in summary_set:

is perfectly fine here

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -231,32 +177,24 @@ def write_mapping_file(sum_inputs_df, target_dir):
...
         return None
 
-
     # Select Default Grouping for either
     # ['lob', 'county', 'policy', 'state', 'location', 'prog']
     elif isinstance(summary_set['oed_fields'], str):
         if summary_set['oed_fields'] in SUMMARY_GROUPING.keys():
```
sr-murthy: Same as above - simpler here to use

    if summary_set['oed_fields'] in SUMMARY_GROUPING:

oasislmf/model_preparation/sum_inputs.py
```diff
@@ -297,110 +234,92 @@ def write_xref_file(summary_xref_df, target_dir):
...
+
+        if isinstance(cols_group_by, list):
+            summary_set_df['summary_id'] = group_by_oed(map_df, exposure_df, cols_group_by)
+        else:
+            # Fall back to setting all in single group
+            summary_set_df['summary_id'] =  pd.Series(1, index=range(set_size), dtype='uint32')
+
+        # Appends summary set to '__summaryxref.csv'
+        summary_set_df['summaryset_id'] =  pd.Series(summary_set['id'], index=range(set_size), dtype='uint32')
+        summaryxref_df = pd.concat([summaryxref_df, summary_set_df])
```
sr-murthy: In this `concat` statement you may want to pass the optional arguments `sort=True` (sorting the non-concatenation axis) and `ignore_index=True` (telling to create a new index starting from 0 on the concatenated dataframe, and not to just concatenate the two axes). Not setting `sort=True` will generate an annoying warning for users.

PR #262, opened 2019-05-02
Feature/exposure summary report (WIP)

hchagani-oasislmf:
Produces exposure summary report: summary of TIVs and location_ids grouped by peril and validation. Writes out this exposure summary report as same directory as keys csv file. Added Status column to keys_errors csv file to make this work.


sr-murthy:
@hchagani-oasislmf I will look at this a bit later today.

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
 
+        # Fill summary of exposure data dictionary
+        def fill_exposure_summary(
+            exposure_summary_df,
+            exposure_summary,
+            peril,
+            status,
+        ):
+
+            for coverage_type in SUPPORTED_COVERAGE_TYPES:
```
sr-murthy: @hchagani-oasislmf I will review the PR fully later, but this Pandas code will have to be refactored to perform operations on columns and sections, rather than doing loops over rows. Various other changes will be required, especially in how columns and subsets of columns are accessed.

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
             exposure_fp, _keys_fp, exposure_profile=exposure_profile
         )
 
+        # Fill summary of exposure data dictionary
+        def fill_exposure_summary(
```
sr-murthy: This method is clearly not a top level method - it needs to be moved out of `manager.py` into a separate module, probably a module named `summaries.py` in `model_preparation`. @sambles I think your `sum_inputs.py` module is also better renamed as `summaries.py`, and should contain all methods and attributes related to the summary stuff.

Also, the method needs renaming - `get_exposure_summary` is a better method name. In general, for simplicity, any method that is producing and returning data (using `return`, not `yield`) should have a `get` name prefix; any method that produces data but generates it using `yield` should have a `generate` name prefix.

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+                    exposure_summary[peril]['all']['tiv_by_coverage'][coverage_type] += tiv_sum
+                else:
+                    exposure_summary[peril]['all']['tiv_by_coverage'][coverage_type] = tiv_sum
+                exposure_summary[peril][status]['tiv'] += tiv_sum
+                exposure_summary[peril]['all']['tiv'] += tiv_sum
+            exposure_summary[peril][status]['location_ids'] =\
+                exposure_summary_df.loc[
+                    exposure_summary_df['peril_id']==peril,
+                    'locnumber'].drop_duplicates().tolist()
+
```
sr-murthy: @hchagani-oasislmf This method isn't returning or generating anything.

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+        try:
+            gul_inputs_errors_df, _ = get_gul_input_items(
+                exposure_fp, _keys_errors_fp, exposure_profile=exposure_profile
+            )
+        except OasisException:
+            gul_inputs_errors_df = pd.DataFrame(columns=gul_inputs_df.columns)
+        keys_errors_df = pd.read_csv(_keys_errors_fp)
+        gul_inputs_errors_df = gul_inputs_errors_df.merge(
+            keys_errors_df,
+            left_on=['locnumber', 'peril_id', 'coverage_type_id'],
```
sr-murthy: `locnumber` is an OED source file column name, and as such should not be hard-coded. You can use the OED hierarchy terms profile/dict to refer to it indirectly. You can see this import at the top of the `manager.py`

    from oasislmf.utils.profiles import get_oed_hierarchy_terms

and how it is used around line 348. There is a variable `loc_num` which holds the value of `locnumber`, which is sourced from that profile. So `loc_num` should be used whereever you are referring to `locnumber`.

sr-murthy: BTW it is OK to refer to the keys file columns directly because the keys file is specific to Oasis and we control the format. The OED source files (locations, accounts, etc.) are in OED format, which is external to Oasis, which is why we have to use profiles to process the files.

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+            right_on=['LocID', 'PerilID', 'CoverageTypeID'],
```
sr-murthy: Columns in the keys dataframe (when it is loaded from the keys file) are lowercased by default - the processing has to be case-insensitive, to here `right_on` should be modified

    right_on=['locid', 'perilid', 'coveragetypeid']

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+            how='left'
+        )
+        gul_inputs_errors_df = gul_inputs_errors_df.drop(
```
sr-murthy: Better when dropping columns (or rows) to use the `inplace=True` argument - that means the drop operations is performed on the object directly, and does not produce another object in memory which has to be reassigned.

sr-murthy: I think you also need `axis=1` here. So this line should be

    gul_inputs_errors_df.drop(columns=['locid', 'perilid', 'coveragetypeid', 'message'], axis=1, inplace=True)

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+            columns=['LocID', 'PerilID', 'CoverageTypeID', 'Message']
+        )
+
+        # Compile summary of exposure data
+        exposure_summary = {}
+        for peril in gul_inputs_df['peril_id'].drop_duplicates():
```
sr-murthy: Are you looping over the set of (unique) peril IDs in the `peril_id` column, or the ordered list of values in the column?

oasislmf/manager.py
```diff
@@ -417,6 +418,67 @@ def generate_oasis_files(
...
+            exposure_summary[peril] = {}
+            for status in ['all'] + list(OASIS_KEYS_STATUS.keys()):
+                exposure_summary[peril][status] = {}
+                exposure_summary[peril][status]['tiv'] = 0.0
+                exposure_summary[peril][status]['tiv_by_coverage'] = {}
+                if status == 'success':
+                    fill_exposure_summary(gul_inputs_df, exposure_summary, peril, status)
+                elif status != 'all':
+                    fill_exposure_summary(gul_inputs_errors_df[gul_inputs_errors_df['Status']==status], exposure_summary, peril, status)
```
sr-murthy: All dataframe columns should be in lowercase - the code cannot be sensitive to case.

oasislmf/manager.py
hchagani-oasislmf: Line 441 includes the assignment, split over multiple lines to ease readability:
```
exposure_summary[peril][status]['location_ids'] =\
    exposure_summary_df.loc[
        exposure_summary_df['peril_id']==peril,
        'locnumber'].drop_duplicates().tolist()
```

sr-murthy: OK, so you're modifying the summary dataframe that the method receives from the main body of `generate_oasis_files` method in the manager. But as I said, I would move this method and the code on lines 465-476 to a separate module in `model_preparation` - probably `summaries.py` which I'm adding in my next update of `develop`. The methods for generating that summary report and/or writing it to file, should be imported in the manager, and then used wherever it is needed.

hchagani-oasislmf:
@sr-murthy I understand. When `summaries.py` is added to `develop`, I'll move the methods and implement the changes you stated above before reissuing the pull request.

hchagani-oasislmf:
Updated with requested changes. Re-requesting review.

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,153 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+    )
+    gul_inputs_errors_df.drop(
+        columns=['locid', 'perilid', 'coveragetypeid', 'message'],
+        axis=1,
+        inplace=True
+    )
+
+    # Compile summary of exposure data
+    exposure_summary = {}
+    for peril_code in gul_inputs_df['peril_id'].drop_duplicates():
```
sr-murthy: Are you trying to loop over the set of peril IDs in the `peril_id` column or over the list of peril IDs in the column? If you want a set then you can use the `unique()` method on the column. It is best to avoid `for` loops in Pandas rows, or, if you have to use one, at least minimise the number of iterations - I think perhaps you may want `groupby` here, i.e. to group the rows in the GUL inputs dataframe by peril ID values, and then deal with each group separately.

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,153 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+        # Use descriptive names of perils as keys
+        for k, vals in PERILS.items():
+            if vals['id'] == peril_code:
+                peril = vals['desc']
+                break
+        exposure_summary[peril] = {}
+        # Create dictionary structure for all and each validity status
+        for status in ['all'] + list(OASIS_KEYS_STATUS.keys()):
+            exposure_summary[peril][status] = {}
```
sr-murthy: As mentioned above, replace chained indexing, i.e. `df[col1][col2]...` with `.loc`. So please replace lines 541-544 with
 
    exposure_summary.loc[:, (peril, status)] = {}
    exposure_summary.loc[:, (peril, status, 'tiv')] = 0.0
    exposure_summary.loc[:, (peril, status, 'tiv_by_coverage')] = {}
    exposure_summary.loc[:, (peril, status, 'number_of_locations')] = 0

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,153 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+                    gul_inputs_df,
+                    exposure_summary,
+                    peril,
+                    peril_code,
+                    status,
+                    loc_num
+                )
+            elif status != 'all':
+                exposure_summary = get_exposure_summary(
+                    gul_inputs_errors_df[gul_inputs_errors_df['status']==status],
```
sr-murthy: Please leave spaces around comparison operators, e.g.

    gul_inputs_errors_df[gul_inputs_errors_df['status'] == status]

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,153 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+                    exposure_summary,
+                    peril,
+                    peril_code,
+                    status,
+                    loc_num
+                )
+
+    # Write exposure summary as json file
+    with io.open(
```
sr-murthy: Pandas provides a `to_json` method for writing a dataframe to a JSON file. That might be more convenient. For example, we use the `to_csv` method all the time to write CSV files from dataframes, which is much more convenient than explicitly doing the file IO bit yourself.

sr-murthy:
@hchagani-oasislmf I've rebased your PR against the latest changes in `develop`. I reviewed this last on Friday, with some comments. Please also run the `flake8` code linting tool against the core `oasislmf` subpackage and make any fixes and commit (you can `pip install flake8` if you don't have it). The command is

    flake8 oasislmf/ --ignore=E501,E402

This will display a list of linting errors/warnings which can fix.

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,153 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+    """
+
+    # Separate TIVs by coverage type and acquire sum
+    for coverage_type in SUPPORTED_COVERAGE_TYPES:
+        tiv_sum = df.loc[
+            (df['peril_id']==peril_code) &\
+            (df['coverage_type_id']==SUPPORTED_COVERAGE_TYPES[coverage_type]['id']),
+            'tiv'].sum()
+        tiv_sum = float(tiv_sum)
+        exposure_summary[peril][status]['tiv_by_coverage'][coverage_type] = tiv_sum
```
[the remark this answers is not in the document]
hchagani-oasislmf: @sr-murthy exposure_summary is a dict. Would you prefer it to be converted to a pandas dataframe?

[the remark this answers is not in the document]
sr-murthy: Is it a dict? I thought it was a dataframe. In that case, you can leave this as it is.

If a dataframe contains a multi-index then you can make those dict-like nested references - I thought it was a dataframe.

oasislmf/model_preparation/summaries.py
hchagani-oasislmf: @sr-murthy I want to look over the set of unique peril IDs. I was under the impression that there was no discernible difference between the `.drop_duplicates()` and `.unique()` methods. The `for` loop should be looping over the result of `gul_inputs_df['peril_id'].drop_duplicates()` which returns a numpy array. The `for` loop shouldn't be looping over any rows in the dataframe.

sr-murthy: For the case of a single column, there's probably no difference, but I have always used `unique()` when looking at the underlying set of values in a column, and I think that would be preferable.

hchagani-oasislmf: OK, I understand. Actually, I was wrong: `.drop_duplicates()` returns a pandas Series, `.unique()` returns a numpy array. I've changed it to `gul_inputs_df['peril_id'].unique()`.

sr-murthy: OK. A Pandas series is basically a wrapper around a Numpy 1d-array, and a Pandas dataframe a wrapper around a Numpy nd-array. You can see this by using the `.values` attribute on any Pandas series or dataframe.

sr-murthy: This means that you can do Numpy operations on the underlying values of any Pandas series or dataframe, and often this is quicker than using Pandas operations.

hchagani-oasislmf:
Had to add status to keys errors dict in unit test. Implemented requested changes with exception of those treating `exposure_summary` as a dataframe. Re-requesting review.

oasislmf/manager.py
```diff
@@ -420,6 +421,16 @@ def generate_oasis_files(
             exposure_fp, _keys_fp, exposure_profile=exposure_profile
         )
 
+        # Write exposure summary file
```
sr-murthy: A change is required on line 395 in the manager to support the deterministic loss generation workflow - in that scenario there is no model and the keys are auto-generated, and there are no keys errors by definition. So an empty keys errors file must be written out anyway.

I will test this, and if it works, update the PR directly.

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,155 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+
+    # Compile summary of exposure data
+    exposure_summary = {}
+    for peril_code in gul_inputs_df['peril_id'].unique():
+        # Use descriptive names of perils as keys
+        for k, vals in PERILS.items():
+            if vals['id'] == peril_code:
+                peril = vals['desc']
+                break
+        exposure_summary[peril] = {}
```
sr-murthy: @hchagani-oasislmf This line 540

        exposure_summary[peril] = {}

breaks - the `peril` variable is defined inside the `for` loop over `PERILS`, but not outside. This causes the deterministic loss generation command to break (`oasislmf exposure run`).

Can you please test your branch version with FM3?

    oasislmf exposure run -s validation/examples/fm3

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,155 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+        columns=['locid', 'perilid', 'coveragetypeid', 'message'],
+        axis=1,
+        inplace=True
+    )
+
+    # Compile summary of exposure data
+    exposure_summary = {}
+    for peril_code in gul_inputs_df['peril_id'].unique():
+        # Use descriptive names of perils as keys
+        for k, vals in PERILS.items():
```
sr-murthy: @hchagani-oasislmf In this `for` loop you're trying to get a matching entry from the `PERILS` dict on the `peril_code` from the GUL inputs dataframe. A more concise way of doing this is

    peril = [v['id'] for v in PERILS.values() if v['id'] == peril_code][0]

It is possible that if the `peril_code` from the dataframe is invalid then this list will be empty and you will get an `IndexError`. So you have to deal with this case by either throwing an exception, which will interrupt the calling method, or skip it in some way. I would recommend the following (replacing the existing `for` loop on line 536)

    try:
        peril = [v['id'] for v in PERILS.values() if v['id'] == peril_code][0]
    except IndexError:
        # no match above, so either raise an OasisException or skip

oasislmf/model_preparation/summaries.py
```diff
@@ -413,3 +419,155 @@ def generate_summaryxref_files(model_run_fp, analysis_settings):
...
+
+    # Get keys in GUL input items dataframe and merge with key errors to obtain
+    # status information
+    try:
+        gul_inputs_errors_df, _ = get_gul_input_items(
+            exposure_fp, keys_errors_fp, exposure_profile=exposure_profile
+        )
+    except OasisException:   # Empty dataframe
+        gul_inputs_errors_df = pd.DataFrame(columns=gul_inputs_df.columns)
+    keys_errors_df = pd.read_csv(keys_errors_fp)
```
sr-murthy: There is a wrapper method called `get_dataframe` in `oasislmf/utils/data.py`, which is recommended for loading dataframes from file. I would recommend setting data types explicitly for every column, which you can do via passing a data types dict.

        dtypes = {
            'locid': 'str',
            'perilid': 'str',
            'coveragetypeid': 'uint32',
            'status': 'str',
            'message': 'str'
    }

So line 517 can be replaced by

    keys_errors_df = get_dataframe(src_fp=keys_errors_fp, col_dtypes=dtypes)

sr-murthy:
@hchagani-oasislmf Just added some comments. The exposure summary stuff might be more relevant to the model loss generation workflow (model, lookup assets provided, direct call to lookup to generate keys), than for the deterministic loss generation (no model, no lookup exists, auto-generation of keys). If you look at `generate_oasis_files` in the manager, you will see that it supports both types - there is a `deterministic` flag that is used to distinguish the two types of workflow.

This branch isn't working for the det. loss generation (`oasislmf exposure run`).

You will need to add this line in the manager (inside the `if deterministic:` block)

    _, _ = olf.write_oasis_keys_errors_file([], _keys_errors_fp)

oasislmf/model_preparation/summaries.py
hchagani-oasislmf: Inserted `if not deterministic` condition before `write_exposure_summary` function call.

hchagani-oasislmf:
All FM test cases pass.

sr-murthy:
OK, I've just updated a block comment in the manager - we can wait for Jenkins to finish running the tests. If it's a green I will merge it in.

PR #277, opened 2019-05-15
Add complex data passthrough from API

DanielFEvans:
Changes that correspond to the OasisPlatform changes in https://github.com/OasisLMF/OasisPlatform/pull/195

For lookup, the API may now pass through a directory containing additional data files, uploaded by the user. They're treated similarly to the input portfolio (lookup), or the input keys data tarfile (loss calculation).


DanielFEvans:
@pauldunning-jbarisk, @carlfischerjba - pull request for OasisLMF corresponding to those API changes.

oasislmf/cli/model.py
```diff
@@ -280,6 +281,10 @@ def action(self, args):
             inputs.get('complex_lookup_config_file_path', required=False, is_path=True),
             'Complex lookup config JSON file path', preexists=False
         )
+        user_data_dir = as_path(
+            inputs.get('user_data_path', required=False, is_path=True),
+            'Directory containing additional user-supplied model data files', preexists=False
+        )
```
carlfischerjba: Any idea why the duplication between the CLI argument parser and here? The help string and `required` appear in both places.

oasislmf/cli/model.py
```diff
@@ -343,7 +348,8 @@ def action(self, args):
             accounts_profile_fp=accounts_profile_fp,
             fm_aggregation_profile_fp=aggregation_profile_fp,
             ri_info_fp=ri_info_fp,
-            ri_scope_fp=ri_scope_fp
+            ri_scope_fp=ri_scope_fp,
+            user_data_dir=user_data_dir
```
carlfischerjba: You could follow PEP8 and add a trailing comma so that future additions of parameters only create a single line diff.

oasislmf/model_execution/bin.py
```diff
@@ -96,11 +97,11 @@ def prepare_run_directory(
     Darwin or Linux, otherwise the source folder tree is recursively
     copied into the ``static`` subfolder.
 
-    :param run_directory: the model run directory
-    :type run_directory: str
+    :param run_dir: the model run directory
```
carlfischerjba: Interesting that the Sphinx style docstrings seem to be the norm is this repository, in contrast to the Google style docstrings used in the platform repository.

oasislmf/model_execution/bin.py
```diff
@@ -151,6 +155,18 @@ def prepare_run_directory(
                     os.symlink(path, os.path.join(model_data_dst_fp, fn))
             except Exception:
                 shutil.copytree(model_data_fp, os.path.join(model_data_dst_fp, fn))
+
+        if user_data_dir and os.path.exists(user_data_dir):
```
carlfischerjba: Log a warning or fail if the specified directory doesn't exist? Otherwise it might take us a while to figure out that there's a typo in the directory path.

oasislmf/model_execution/bin.py
```diff
@@ -151,6 +155,18 @@ def prepare_run_directory(
...
+            for path in glob.glob(os.path.join(user_data_dir, '*')):
+                fn = os.path.basename(path)
```
carlfischerjba: Took me a while to realise `fn` was `filename`.

oasislmf/model_execution/bin.py
```diff
@@ -151,6 +155,18 @@ def prepare_run_directory(
...
+                try:
+                    if os.name == 'nt':
+                        shutil.copy(path, os.path.join(oasis_dst_fp, fn))
+                    else:
+                        os.symlink(path, os.path.join(oasis_dst_fp, fn))
+                except Exception:
```
carlfischerjba: That's quite a generic `except`. If you really don't want to be more specific consider a comment explaining why `copytree` is expected to work when `copy` and `symlink` didn't.

oasislmf/model_preparation/lookup.py
```diff
@@ -322,15 +323,25 @@ def get_custom_lookup(
...
+                output_directory=output_directory
+            )
+        else:
+            return klc(
+                keys_data_directory=keys_data_path,
+                supplier=model_info['supplier_id'],
+                model_name=model_info['model_id'],
+                model_version=model_info['model_version'],
+                complex_lookup_config_fp=complex_lookup_config_fp,
+                user_data_dir=user_data_dir,
```
carlfischerjba: Can't you just let this be `None` and avoid the extra `elif` above (and the other `elif` above that)? I would have thought that leaving out the argument would cause it to take a default value which I would expect to be `None` inside the lookup function.

oasislmf/model_execution/bin.py
DanielFEvans: Yes, hence why I think that one you spotted elsewhere got through - it's quite easy to get mixed up when you're swapping back and forth between two or three projects.

oasislmf/cli/model.py
DanielFEvans: Not particularly. Another one to keep in mind is that there's a slightly confusing swap from`names-like-this` and `names_like_this` - even though the CLI defines arguments as the former, the `inputs` contains them as the latter.

sr-murthy: What is the duplication here? The `inputs` object (created from the `cli.base.InputValues` class) is a more user-friendly object wrapper around the arguments `args` object parsed by the command parser (when `action` is called), and `as_path` is a "path cleaner" that is applied to the path-like arguments in `inputs`. There is some overlap between some of the optional arguments used by the parser, the `inputs` object and the `as_path` cleaner, because this is required in the way that `InputValues` and the path cleaner are meant to work.

sr-murthy: Also, with respect to the format of the argument option names for command arguments, as you may know, from using various command line tools, e.g. `curl`, the convention is hyphen-separated, e.g.

    $ curl --help
    Usage: curl [options...] <url>
    Options: (H) means HTTP/HTTPS only, (F) means FTP only
         --anyauth       Pick "any" authentication method (H)
         -a, --append        Append to target file when uploading (F/SFTP)
         ...

but the variables which are used internally by the command line tool to refer to those option names don't need to be hyphen separated. There is no rule saying that they have to use the same hyphen-separated convention, which is probably better for readability. Internally, the variable name syntax for CLI argument option names in the MDK follows Python convention, which is underscoring. Argparse automatically converts hyphen-separated option names to values where the hyphen is replaced with an underscore, e.g. `keys-file-path` -> `keys_file_path`, etc.

oasislmf/cli/model.py
sr-murthy: We use Flake8 to lint the code, and I didn't see any PEP8 warnings/errors about the lack of a trailing comma.

oasislmf/model_execution/bin.py
sr-murthy: MDK uses Sphinx for its docs, hence the reST syntax for docstrings. This will continue to be the case. The Google style docstring format used in Oasis Platform may be more appropriate there because of the requirements for documenting the API. There's no reason why the two have to be the same.

oasislmf/cli/model.py
DanielFEvans: I believe Carl was referring to the fact that the help string, whether an option is required, etc., are defined both in `add_args` and `action`. Am I right to think that the ones in `action` are used in error messages (e.g. "XXX was indicated as pre-existing...")?

Interesting point about argparse doing the conversion - I didn't realise that was actually its default behaviour.

sr-murthy: The help text in `add_args` is used when you run `--help` on the command. In `action` we keep it because it may be useful when raising an exception for example.

sr-murthy: In Python you can't have hyphenated variable names anyway, and so if you the convention of hyphenated command line option names then you must convert those names to a format that can be processed by Python, e.g. underscoring.

DanielFEvans: Oh, yes, that's an excellent point - that does make it obvious that argparse has to do a conversion.

carlfischerjba: I'm just pointing out that when you add an argument you have to do it in at least two places and that could be considered a code smell. If there is no better alternative to the way it's currently done, that's fine.

oasislmf/cli/model.py
carlfischerjba: From https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas:
> When trailing commas are redundant, they are often helpful when a version control system is used, when a list of values, arguments or imported items is expected to be extended over time. The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line. 

Apparently Flake8 doesn't report it but you can add the `flake8-commas` plugin which will report "C813 - missing trailing comma in Python 3" in cases like this (if I've understood correctly). It's a minor detail.

oasislmf/cli/model.py
sr-murthy: There are probably better alternatives, but for the moment we will keep it as it is. Help text doesn't figure at all in the logic of processing the commands, so the duplication isn't going to break anything, therefore it can't be considered code smell.

oasislmf/model_execution/bin.py
carlfischerjba: Consistency between projects would be nice but if there are technical reasons for doing things differently in different projects we'll respect that.

sr-murthy:
This has to pass the tests, which don't appear to have been triggered, but in any case it will have to wait until `develop` is updated, which will be later today or tomorrow.

oasislmf/cli/model.py
sr-murthy: You're free to add trailing commas in the appropriate places in your contributions, but we can't guarantee that this will be standard practice, because different people in the team have different approaches to writing the code - for example, I always use trailing commas when writing multi-line import statements within parentheses, but not in multi-line method calls, but other members write and indent imports and method calls differently etc. We try to be consistent but 100% consistency isn't always possible.

Also, as indicated in the PEP8 guide, the comment on trailing commas is really a suggestion in special cases, not a general recommendation, which is why it does not appear in the canonical list of PEP8 or Flake8 errors or warnings.

sr-murthy:
@carlfischerjba @DanielFEvans I will merge this in now. For your other comments/suggestions in relation to some of the core MDK modules, you can submit a separate PR for this.

PR #286, opened 2019-05-27
Fix/corelogic

sambles:


sambles:
> @sambles Calc. rule IDs and policy TC IDs, like all ID columns, are sequences of positive integers, therefore unsigned integer data types are best suited to represent them. Using signed integers here is not correct.

the columns 'calcrule_id' and 'policytc_id' were being set to signed values with an unsigned datatype: 
https://github.com/OasisLMF/OasisLMF/blob/develop/oasislmf/model_preparation/il_inputs.py#L388-L394

this leads to the following situation: 
```
   acc_idx                  accnumber  agg_id  areaperil_id  attachment  calcrule_id  ...  policytc_id                  polnumber     portnumber  share          tiv  vulnerability_id
0        0  OakOneSite_COM-STL_COMBUS       1            -1         0.0   4294967295  ...   4294967295  OakOneSite_COM-STL_COMBUS  OakOneSiteLim    0.0  210000000.0                -1
1        0  OakOneSite_COM-STL_COMBUS       2            -1         0.0   4294967295  ...   4294967295  OakOneSite_COM-STL_COMBUS  OakOneSiteLim    0.0  175000000.0                -1
2        0  OakOneSite_COM-STL_COMBUS       3            -1         0.0   4294967295  ...   4294967295  OakOneSite_COM-STL_COMBUS  OakOneSiteLim    0.0  132000000.0                -1
```

So we either need to keep the dtype to signed int, or not use `-1` as a placeholder

sambles:
> @sambles This may cause problems if you end up with 0s in the calc. rule ID columns - there is no valid calc. rule ID of 0. The merge here should work in terms of not producing any NaNs - if there are NaNs it means some items have combinations of financial terms which don't correspond to any existing calc. rule.

ah yeah, I just found this out. The issue was on the df merge it was throwing "Cannot convert non-finite values (NA or inf) to integer"  because some of the columns had NaN's. would adding back in the following be sufficient?
https://github.com/OasisLMF/OasisLMF/commit/60e44ca798b0a9cb92062b630c79cbe5fff302a4

sr-murthy:
@sambles RE the unsigned int. data type for the ID columns, I think using 0 as the placeholder value will be fine. 0 is the smallest value in the 32 bit unsigned data type range, and as IDs are always positive integers, 0s can represent temporary values.

sr-murthy:
RE the NaNs in the calc. rules df merge, I think we could keep the `fillna` method call, post-merge, but use a default calc. rules ID when there is non-standard combination of financial terms. You can look at this table

https://github.com/OasisLMF/OasisLMF/blob/develop/oasislmf/_data/calc_rules.csv

It's a static file that describes all currently existing calc. rules - I think you have to find out what FM level  those NaNs appear in (`orig_level_id` in the IL inputs table), and check the financial terms (deductible, deductible min., dedutible max., limit, attachment, share) applicable for that level. Perhaps you can ask @johcarter.

sambles:
@sr-murthy 
> I will look at this on Monday, but I don't think this is necessary - we want to avoid any checks here. The calc. rules table should assign a unique calc. rule ID for every valid combination of financial terms, and as long as this is true, no matter what combinations of financial terms the items have in the IL inputs table, the merge between `il_inputs_calc_rules_df` and `calc_rules` will not produce nulls. All of the calc. rules logic should be encoded in the calc. rules table (sourced from the `calc_rules.csv` file in `_data`).
> 
> So I will be removing this anyway.

I think that, for the moment, its useful having a check which we can remove once there is some kind of input validation in place. 

The problem isn't the valid combinations its the invalid ones, which currently didn't pickup on

sr-murthy:
But an invalid combination of financial terms implies an error in the way the terms are extracted from the exposure, and we can't code the calc. rules logic to process an invalid combination of financial terms. It makes no sense to do it that way.

For all items the financial terms (deductible, deductible min., deductible max., limit, ded. type, ded. code, lim. code, lim. type etc.) must be processed correctly, and based on the indicator columns (`deductible_gt_0` whcih says whether the ded. is > 0, etc.) the merge with the calc. rules dataframe should produce a column of non-null calc. rule IDs. 

The problem here isn't "invalid combinations", it's the fact that the calc. rules table is missing additional rules dealing with some combinations of financial terms which aren't covered. If you add those missing rules in that CSV files you will not get any nulls in that merge.

sambles:
So we're talking up to an additional (2^10 - current_rows)  extra 1000 rows to add in to `calc_rules` ?

less than that if we add x or *don't care* values

sr-murthy:
@sambles No. In the calc. rule ID logic the calc. rule ID is viewed as a map on combinations of ranges of financial terms, not on combinations of specific values. Below is a screenshot of the calc. rules table as a spreadsheet - note the calc. rule ID 7 (limit and max. deductible)

Screen Shot 2019-05-28 at 14 21 53

In this rule, the (blanket) deductible term is 0, which means `deductible_gt_0` is 0 (it's 1 if deductible > 0). As it's written this rule only applies when the ded. type is 0 (deductible expressed as a dollar amount), but does not apply when the ded. type is 1 (deductible as a % of loss) or 2 (deductible as a % of TIV). I think what @johcarter is saying is that there should be another rule in the table, with almost identical values in all columns for calc. rule ID 7, but where the ded. type is 1 or 2. So we have to extend the ID 7 calc. rule to cover the cases when ded. type is 1 or 2. So here, you would create two rows in the existing table which are copies of the existing rule for ID 7, and in copy you would only set ded. type is 1 and leave all columns as they are, and in the other copy you would only set ded. type is 2 and leave all other columns as they are.

Similarly and independently, we have to extend all existing calc. rules in the table where limit is 0 and limit type is 0 (limit as a dollar amount) to cases where all other financial terms are the same but the limit type is 1 or 2.

sr-murthy:
@sambles What you could do is to revert your changes here, and instead add those missing calc. rules in the calc. rules CSV using the following procedure.

1. Open the calc. rules CSV (the actual copy in `oasislmf/_data/`) in an editor.

2. In any existing where deductible is 0 and ded. type is 0 (deductible as a monetary amount), i.e. `deductible_gt_0` is `0` and `ded_type` is `0`, look for whether there is a corresponding entry, with the same calc. rule ID, where `ded._type` is `1` (deductible as % of loss) or `2` (deductible as % of TIV), and if there isn't, then add those entries.

3. Do the same for rules where limit is 0 (`limit_gt_0` is `0`) and limit type is 0 (`lim_type` is `0`).

4. Save the amended file, and then in iPython import the `update_calc_rules` method from `oasislmf.utils.calc_rules`, and then call that method. It will automatically update the `id_key` column (a composite column made of the indicator columns) and update the file. Then save it, and then commit.

5. Try running your model again.

sambles:
Reverted the check and merging this in to release -> `1.3.8`

PR #292, opened 2019-05-29
Fix/exposure summary peril groups

hchagani-oasislmf:
Add ability to deal with peril groups when creating exposure summary report.


hchagani-oasislmf:
Found some issues in `write_exposure_summary` method while developing unit test. Updated `model_preparation/summaries.py`.

PR #306, opened 2019-06-07
Testing write_exposure_summary method

hchagani-oasislmf:
Initial unit test of `write_exposure_summary` method. Checks arithmetic in output file.


tests/model_preparation/test_summaries.py
```diff
@@ -0,0 +1,240 @@
...
+    source_exposure,
+    write_source_files,
+    write_keys_files,
+)
+
+
+class TestSummaries(TestCase):
+
+    @staticmethod
+    def get_keys_dict(peril_ls, nlocations, successful=True, start_loc=1):
```
sr-murthy: @hchagani-oasislmf This isn't my full review (which I will do on Monday), but I note that this method (`get_keys_dict`) is unnecessary. So we already have exposure and keys data generators which use the Hypothesis library in the `tests/data.py` module

https://github.com/OasisLMF/OasisLMF/blob/develop/tests/data.py

This is the keys data generator (generates a list of keys dicts)

https://github.com/OasisLMF/OasisLMF/blob/develop/tests/data.py#L354

and you can see it being used in the FM acceptance tests module

https://github.com/OasisLMF/OasisLMF/blob/develop/tests/acceptance/test_fm_acceptance.py#L108

This is a locations data generator

https://github.com/OasisLMF/OasisLMF/blob/develop/tests/data.py#L197

and you can see it also being used in the FM tests module. 

There are also methods that will write a list of keys records to a keys file).

hchagani-oasislmf: I wanted to create keys data where multiple perils were associated with a single location. I couldn't see a way to do this using the keys data generator in `tests/data.py`.

Potentially, I could add a method in `tests/data.py` to do this instead of the `get_keys_dict` method in `tests/model_preparation/test_summaries.py`.

sr-murthy: You can probably re-use the `keys` method in `data.py` - just generate a list of items with temporary values for peril ID and then in the test case modify the dicts as you need at the top of the test case, before processing. We do something similar in the FM acceptance tests.

You can do the same thing for the locations.

If you can't make this work then of course you can use your custom method. But I think the general preference should to re-use existing data generators, which were designed to be customised.

hchagani-oasislmf:
I've updated the test to reuse the `keys` method in `data.py`. I've also introduced the random generation of keys and keys errors in an attempt to enhance the test.

sr-murthy:
@hchagani-oasislmf The new tests don't pass though - if you click on the Details link in the CI status checks box above, you will see the details

https://ci.oasislmfdev.org/blue/organizations/jenkins/oasis_oasislmf/detail/PR-306/2/pipeline

sr-murthy:
I will review it, but you would need to get the tests passing.

hchagani-oasislmf:
I just noticed. I'll take a look.

sr-murthy:
@hchagani-oasislmf Test cases are best debugged using `ipdb` - set a breakpoint in your test case using `import ipdb; ipdb.set_trace()` where you need it, and run the specific failing test case using

    python3.6 -m pytest -sp no:flaky -v tests/model_preparation/test_summaries.py::TestSummaries::test_write_exposure_summary

sr-murthy:
You will need to install `ipdb` (`pip install ipdb`) if you don't already have it.

hchagani-oasislmf:
Build failed on data generation to slow HealthCheck. Suppressed this HealthCheck.

PR #323, opened 2019-06-27
Add flag to make exposure summary generation optional

hchagani-oasislmf:
Added flags `-S` and `--summarise-exposure` to generate exposure summary report. Default value is set to `None`, so flag must be specified for report to be generated. Flag can also be set to `true` in MDK config file.

### DIFF
--- a/oasislmf/cli/config.py
+++ b/oasislmf/cli/config.py
@@ -16,6 +16,7 @@ class ConfigCmd(OasisBaseCommand):
     with appropriate values, but can be overridden by providing a runtime flag.
 
     :analysis_settings_file_path: Analysis settings (JSON) file path
+    :exposure_summary_report: Whether to generate exposure summary report
     :keys_data_path: Model lookup/keys data path (optional)
     :lookup_config_file_path: Model built-in lookup config. (JSON) file path (optional)
     :lookup_package_path: Model custom lookup package path (optional)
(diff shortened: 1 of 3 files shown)


oasislmf/cli/config.py
```diff
@@ -29,6 +29,7 @@ class ConfigCmd(OasisBaseCommand):
     :source_accounts_profile_path: Source OED accouns (JSON) profile describing the financial terms contained in the source accounts file (optional)
     :source_exposure_file_path: Source OED exposure (CSV) file path
     :source_exposure_profile_path: Source OED exposure (JSON) profile describing the financial terms contained in the source exposure file (optional)
+    :summarise_exposure: Whether to generate exposure summary report
```
sr-murthy: It might be better to get rid of "whether etc.", and to add the file type to the help text, e.g.

   Generates an exposure summary report in JSON 

I think this reads better. I think there are probably other commands where we've got "whether" in the help text for boolean arguments, which we should also rewrite to make them simpler, but for the moment, we can just update this one.

oasislmf/manager.py
```diff
@@ -422,7 +423,7 @@ def generate_oasis_files(
         )
 
         # If not in det. loss gen. scenario, write exposure summary file
-        if not deterministic:
+        if not deterministic and summarise_exposure:
```
sr-murthy: @hchagani-oasislmf Looks fine, but can you please make that change to the help text (https://github.com/OasisLMF/OasisLMF/pull/323/files/dd5f3555aae65f3da120b9cbb1deb3f9da1402ce#r298207663), and also rewrite this conditional as follows:

    if summarise_exposure and not deterministic:

I think the conditional is a bit ambiguous if it started with checking for `deterministic`.

hchagani-oasislmf: @sr-murthy Done on both counts. Pushed new files to repo.

PR #324, opened 2019-06-28
Changed default values of ktools boolean flags from False to None

hchagani-oasislmf:
Encountered this issue when adding a new flag to generate the exposure summary report. Issuing a command line argument to set the value for boolean arguments to `True` works. However, using the MDK config file to do this results in being ignored in favour of the default `False` value. Default values for the bools `ktools_mem_limit` and `ktools_fifo_relative` have been set to `None` to correct for this.


oasislmf/cli/model.py
```diff
@@ -406,8 +406,8 @@ def add_args(self, parser):
         parser.add_argument('-r', '--model-run-dir', default=None, help='Model run directory path')
         parser.add_argument('-p', '--model-package-path', default=None, help='Path containing model specific package')
         parser.add_argument('-n', '--ktools-num-processes', default=None, help='Number of ktools calculation processes to use', type=int)
-        parser.add_argument('-m', '--ktools-mem-limit', default=False, help='Force exec failure if Ktools hits memory the system  memory limit', action='store_true')
-        parser.add_argument('-f', '--ktools-fifo-relative', default=False, help='Create ktools fifo queues under the ./fifo dir', action='store_true')
+        parser.add_argument('-m', '--ktools-mem-limit', default=None, help='Force exec failure if Ktools hits memory the system  memory limit', action='store_true')
```
sr-murthy: I'm not sure what the problem was here originally, but default values are useful to have. We use a custom input values object to process the command arguments in the CLI (whether provided directly or via a config. file)

https://github.com/OasisLMF/OasisLMF/blob/develop/oasislmf/cli/base.py#L36

After adding arguments to the parser you will note this line in most of the commands

    inputs = InputValues(args)

The `inputs` object has a `get` method that provides a dict-like access to the arguments. In that method there are separate branches for processing directly provided arguments and arguments provided via a config. file. Default values override the values coming in from either of those branches only if those values are null.

This seems to make sense.

hchagani-oasislmf: @sr-murthy I ran the PiWind model putting a breakpoint in `generate_model_losses`. If I add the following line to the end of `oasislmf.json`:

hchagani-oasislmf: @sr-murthy I ran the PiWind model putting a breakpoint in generate_model_losses. If I add the following line to the end of oasislmf.json:
```
    "ktools_mem_limit": true
```
Then run the following from the command line:
```
oasislmf model run -C oasislmf.json
```
Then ask for the value of `ktools_mem_limit` at the breakpoint, it returns `False`.

Everything works fine if I give the `--ktools-mem-limit` argument on the command line. However, the setting in `oasislmf.json` appears to be ignored.

sr-murthy: OK, I'll take a look, but that behaviour isn't correct.

sr-murthy: @hchagani-oasislmf It is `model run` you're using? I inserted a breakpoint in the model run command just after the point where it creates the input values object

    inputs = InputValues(args)

and when running PiWind with the ktools mem. limit flag set to true, it is picking up the correct value.
```
(oasislmf-py3-venv) [2019-07-05 12:38:02 srm@~/Documents/sandeep/cst/dev/oasis/oasislmf]$ oasislmf model run -C ../OasisPiWind/oasislmf.json 

Processing arguments for model run
> /Users/srm/Documents/sandeep/cst/dev/oasis/OasisLMF/oasislmf-py3-venv/src/oasislmf/oasislmf/cli/model.py(556)action()
    555 
--> 556         utcnow = get_utctimestamp(fmt='%Y%m%d%H%M%S')
    557         default_model_run_fp = os.path.join(os.getcwd(), 'runs', 'ProgOasis-{}'.format(utcnow))

ipdb> inputs                                                                                                                      
<oasislmf.cli.base.InputValues object at 0x1109a43c8>
ipdb> inputs.get('ktools_mem_limit')                                                                                              
True
```

sr-murthy:
https://github.com/OasisLMF/OasisLMF/issues/332

sr-murthy:
@hchagani-oasislmf I'm closing this because I updated `develop` last Friday with some fixes which make this PR unnecessary.

The default value for any argument in the model run and loss generation commands is only used if there is no value provided for the argument on the command line or the config. JSON. In particular, for booleans this means default values set in the command parser can never override the command line or config. derived values if they are present.

PR #351, opened 2019-07-17
Fixed GUL calcs,and added option to output results to file.

mpinkerton-oasis:
Fix to output GULs plus addition of output to file option.

### DIFF
--- a/oasislmf/cli/exposure.py
+++ b/oasislmf/cli/exposure.py
@@ -77,6 +77,10 @@ class RunCmd(OasisBaseCommand):
             '-o', '--output-level', default='item', 
             help='Level to output losses. Options are: item, loc, pol, acc or port.', type=str
         )
+        parser.add_argument(
+            '-f', '--output-file', default=None, 
+            help='Write the output to file.', type=str
+        )
 
     def action(self, args):
         """
@@ -114,6 +118,8 @@ class RunCmd(OasisBaseCommand):
             raise OasisException(
                 'Invalid output level. Must be one of port, acc, loc, pol or item.'
             )
+
+        output_file = inputs.get('output_file', default=None, required=False)
 
         src_contents = [fn.lower() for fn in os.listdir(src_dir)]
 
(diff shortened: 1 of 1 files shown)


oasislmf/cli/exposure.py
```diff
@@ -115,6 +119,8 @@ def action(self, args):
                 'Invalid output level. Must be one of port, acc, loc, pol or item.'
             )
 
+        output_file = inputs.get('output_file', default=None, required=False)
```
sr-murthy: @mpinkerton-oasis Since `output_file` is a path we need to apply a little method we use everywhere else in the CLI when dealing with file paths, which is `as_path` (a method in `utils.path` which is already imported in `exposure.py`). This method is used to do a couple of things: convert relative paths to absolute paths if the original argument was provided as a relative path, and also check whether the paths pre-exist or are created in the course of the backend execution. Here `output_file` appears to be optional and doesn't necessarily exist prior to running the command. So I would replace line 122 with this one

    as_path(inputs.get('output_file', required=False, is_path=True), 'Output file path', preexists=False)

oasislmf/cli/exposure.py
```diff
@@ -217,6 +223,8 @@ def action(self, args):
                 loss_factor, total_gul, total_il, total_ri_ceded)
 
         print_dataframe(all_losses_df, frame_header=header, string_cols=all_losses_df.columns)
+        if output_file:
+            all_losses_df.to_csv(output_file, index=False)
```
sr-murthy: May be not necessarily here if the table only contains numbers and no text, but there is no harm in adding argument to set encoding type

    all_losses_df.to_csv(output_file, index=False, encoding='utf-8')

oasislmf/cli/exposure.py
mpinkerton-oasis: Fixed

oasislmf/cli/exposure.py
mpinkerton-oasis: Fixed. Note that utf-8 is the default encoding in Python 3 anyway.

Issue #383, opened 2019-08-21
Set dtypes for exposure data in lookup


mpinkerton-oasis:
Amend the code in lookup.py to explicitly set the OED field types. This is specifically to address an issue where some geographic codes have leading 0s and the default Pandas behavior is to drop the 0s and cast them as numbers, causing the lookup to fail.

This is the list of all loc. file columns (OED) which in the OED schema are defined to be “string” columns
```[
    "AccNumber (Loc)",
    "AreaCode (Loc)",
    "AreaName (Loc)",
    "BuildingValuation (Loc)",
    "City (Loc)",
    "CompletionDate (Loc)",
    "CountryCode (Loc)",
    "CustomFloodZone (Loc)",
    "Datum (Loc)",
    "FlexiLocZZZ (Loc)",
    "FloodZone (Loc)",
    "FloorsOccupied (Loc)",
    "Geocoder (Loc)",
    "GeogName1 (Loc)",
    "GeogName2 (Loc)",
    "GeogName3 (Loc)",
    "GeogName4 (Loc)",
    "GeogName5 (Loc)",
    "GeogScheme1 (Loc)",
    "GeogScheme2 (Loc)",
    "GeogScheme3 (Loc)",
    "GeogScheme4 (Loc)",
    "GeogScheme5 (Loc)",
    "LocCurrency (Loc)",
    "LocExpiryDate (Loc)",
    "LocGroup (Loc)",
    "LocInceptionDate (Loc)",
    "LocName (Loc)",
    "LocNumber (Loc)",
    "LocPeril (Loc)",
    "LocPerilsCovered (Loc)",
    "LocUserDef1 (Loc)",
    "LocUserDef2 (Loc)",
    "LocUserDef3 (Loc)",
    "LocUserDef4 (Loc)",
    "LocUserDef5 (Loc)",
    "OrgConstructionCode (Loc)",
    "OrgConstructionScheme (Loc)",
    "OrgOccupancyCode (Loc)",
    "OrgOccupancyScheme (Loc)",
    "PortNumber (Loc)",
    "PostalCode (Loc)",
    "ReinsTag (Loc)",
    "StreetAddress (Loc)"
]```
Of these the mandatory (required) loc. file columns which are string columns are
```[
    "AccNumber (Loc)",
    "CountryCode (Loc)",
    "LocCurrency (Loc)",
    "LocNumber (Loc)",
    "LocPerilsCovered (Loc)",
    "PortNumber (Loc)"
[...]

PR #387, opened 2019-08-23
Setting dtypes for exposure data in lookup

hchagani-oasislmf:
Regarding https://github.com/OasisLMF/OasisLMF/issues/383 , the following changes have been made:

1. New JSON `_data/loc_dtypes.json` with data types for exposure file derived from `OED_LOCATION_COLS` in `utils/summary_levels.py`.
2. Add method in `utils/defaults.py` to get dict from aforementioned JSON.
3. Add method in `utils/data.py` to get datatype dict and list of required columns from aforementioned dict.
4. Data types set explicitly in exposure dataframe for string columns.
5. Required columns of data type `string` are now a requirement in a valid exposure dataframe.
6. Minor changes to unit tests to account for required columns in exposure dataframe.

Currently, only columns of data type `string` are affected. Required columns of other data types are not checked. Additionally, data types for non-string columns are not altered. This can be changed easily now or in the future.


oasislmf/utils/data.py
```diff
@@ -293,7 +295,13 @@ def get_dataframe(
     na_values = list(PANDAS_DEFAULT_NULL_VALUES.difference(['NA']))
 
     if src_fp and src_type == 'csv':
-        df = pd.read_csv(src_fp, float_precision=float_precision, memory_map=memory_map, keep_default_na=False, na_values=na_values)
+        # Find flexible fields in loc file and set their data types to that of
+        # FlexiLocZZZ
+        if 'FlexiLocZZZ' in col_dtypes.keys():
```
sr-murthy: Hard-coding column names isn't recommended - it would be better to add this column to the loc. JSON profile (in static data), as it is a loc. file column, describe it as part of the OED hierarchy (see how acc. num., loc. num., etc. are defined in the loc. profile), and use the OED hierarchy terms wrapper (in `utils/profiles.py` I think) to access the column name internally. This probably sounds very complicated, but it is about the principle of making the code independent of column names of files where the schema is subject to change externally.

For the moment, you can keep it - perhaps you can do this in the next release.

oasislmf/utils/data.py
```diff
@@ -360,6 +368,28 @@ def get_dataframe(
...
 
+def get_dtypes_and_required_cols(get_dtypes):
+    """
+    Get OED column data types and required column names from JSON.
+
+    :param get_dtypes: method to get dict from JSON
+    :type get_dtypes: function
+    """
+    dtypes = get_dtypes()
+    col_dtypes = {
```
sr-murthy: This only deals with string columns - there are of course OED columns with SQL data types which map to other Python data types such as ints and floats, so the name `col_dtypes` is not quite accurate, because you're only interested in columns whose (Python) values are expected to be strings.

Also, as you're only selecting string columns and set the values for the keys to `v['py_dtype']`, which by construction must be `str`, every value in this dictionary will be `str`. I don't think is useful or necessary.

PR #421, opened 2019-10-16
method merge_check to raise exeption on empty col intersection

sambles:


oasislmf/utils/data.py
```diff
@@ -471,6 +472,37 @@ def get_utctimestamp(thedate=datetime.utcnow(), fmt='%Y-%b-%d %H:%M:%S'):
     return thedate.astimezone(pytz.utc).strftime(fmt)
 
 
+def merge_check(left, right, on=[], raise_error=True):
+    """
+
```
hchagani-oasislmf: Unnecessary line break.

PR #431, opened 2019-11-06
Feature/429 il allocrule

sambles:
* Set default allocation rules  GUL=1, IL=2, RI=3
* Add RI allocation rule flag


oasislmf/manager.py
```diff
@@ -730,7 +786,11 @@ def run_deterministic(
...
             net_ri=net_ri,
-            alloc_rule=alloc_rule
+            il_alloc_rule=il_alloc_rule,
+            ri_alloc_rule=ri_alloc_rule
         )
 
         return losses['gul'], losses['il'], losses['ri']
+
+
+
```
hchagani-oasislmf: Unnecessary line breaks.

PR #495, opened 2020-02-05
Fix/various issues jan

### LINKED ISSUE #475: OasisLMF may copy the entire model data directory to networked output
We set up an OasisLMF run on a Linux machine, with the run outputs pointed to a networked (non-Linux-based) drive, which is one of our centralised data storage devices, and also provides much more space than is available on a local machine (OasisLMF outputs and intermediate data can be large, especially if doing a few dozen runs).

However, we've found that OasisLMF is copying the entire model data directory across to the run outputs directory - i.e. somewhere around 1TB of data over the network.

This has occurred following the `STARTED: oasislmf.model_execution.bin.prepare_run_directory
` log entry. I believe it's in the linking of the `static` data directory within the run directory. This is usually symlinked to the model data directory, which is failing (and falling back on copying) when referencing local data from a networked (non-Linux-based) drive.

OasisLMF v1.4.5 for reference, although I don't think this particular functionality has changed in a while.

### LINKED ISSUE #477: setup.py fails when behind a proxy
Proxy does not seem to be inherited from pip into setup.py, so when you try to install from the package folder like 
```
pip install . --proxy $HTTP_PROXY
```
... it will not install the package because it won't be able to download the ktools part.

https://github.com/OasisLMF/OasisLMF/blob/29d4b8be03eced79c8ac05d0e80e18161cefb684/setup.py#L70

Furthermore, timeout is set to 5000 seconds, which means it just hangs.

Preferred behavior would be...
1 - shorter timeout
2 - continue and finish the installation with a warning if the download fails

### LINKED ISSUE #482: symlink error when rerunning analysis using existing analysis_folder
os.symlink will raise an exception if the destination already exists. So the following will cause generate_model_losses to crash if you try to reuse an existing analysis folder...

https://github.com/OasisLMF/OasisLMF/blob/29d4b8be03eced79c8ac05d0e80e18161cefb684/oasislmf/model_execution/bin.py#L155

What do you think about adding the following except?
```
    try:
        # Use symbolic link if we can
        os.symlink(sourcefile, destfile)

    except OSError as why:
        if why.errno == errno.EEXIST and os.path.islink(destfile):
            # If the link already exists, check files are different replace it
            if os.readlink(destfile) != os.path.abspath(sourcefile):
                os.symlink(sourcefile, destfile + ".tmp")
                os.replace(destfile + ".tmp", destfile)


```

### LINKED ISSUE #492: Complex Model error guard
Call to a complex model needs to be wrapped in a sub-shell 

**Currently is**
```
(  MyBin -m 1 -n 2 -a /home/admin/exposures/runs/ProgOasis-20200205110504/analysis_settings.json -p . -S10 -c /tmp/4Nua7Uoqq2/fifo/gul_P1 -i - | tee /tmp/4Nua7Uoqq2/fifo/gul_P1 2> log/gul_stderror.err | fmcalc -a2 > /tmp/4Nua7Uoqq2/fifo/il_P1  ) 2>> log/stderror.err &
```

**Change too**
```
( ( MyBin -m 1 -n 2 -a /home/admin/exposures/runs/ProgOasis-20200205110504/analysis_settings.json -p . -S10 -c /tmp/4Nua7Uoqq2/fifo/gul_P1 -i - | tee /tmp/4Nua7Uoqq2/fifo/gul_P1 ) 2>> log/gul_stderror.err | fmcalc -a2 > /tmp/4Nua7Uoqq2/fifo/il_P1  ) 2>> log/stderror.err &
```

### LINKED ISSUE #493: generate_model_losses crashes with spaces in filepath
This code will cause an error if there is a space in the input or output file path...

https://github.com/OasisLMF/OasisLMF/blob/0bc1cb26aafc723848690b1487716ba96f9e90c3/oasislmf/model_execution/bin.py#L332

It is a simple fix...
```
cmd_str = "{} < \"{}\" > \"{}\"".format(conversion_tool, input_file_path, output_file_path) 
```

sambles:
* Fix #475 - Remove fallback copy of model data                                                                                                                                                                                                                                                                                                                                                
* Fix #482 - check for existing symlinks and replace
* Fix #477 - remove excessive timeout + add back proxy handling if set 
* Fix #493 - spaces in filepath
* Fix #492 - custom gul error subshell


tests/model_execution/test_bin.py
```diff
@@ -528,16 +528,16 @@ def test_model_data_src_is_supplied___symlink_to_output_dir_static_is_created(se
...
+#    def test_model_data_src_is_supplied_sym_link_raises___input_is_copied_from_static(self):
+#        with TemporaryDirectory() as run_dir, TemporaryDirectory() as oasis_src_fp, TemporaryDirectory() as model_data_fp:
+#            analysis_settings_fp = os.path.join(oasis_src_fp, "settings.json")
+#            Path(analysis_settings_fp).touch()
+#            Path(os.path.join(model_data_fp, 'linked_file')).touch()
+#
+#            with patch('os.symlink', Mock(side_effect=OSError())):
+#                prepare_run_directory(run_dir, oasis_src_fp, model_data_fp, analysis_settings_fp)
+#
+#            self.assertTrue(os.path.exists(os.path.join(run_dir, 'static', 'linked_file')))
```
hchagani-oasislmf: Delete commented-out code?

PR #501, opened 2020-02-09
Simplify step policy flag detection in det. loss generation

sr-murthy:
* Use more Pythonic approach to step policy flag detection in det. loss generation

### DIFF
--- a/oasislmf/utils/deterministic_loss.py
+++ b/oasislmf/utils/deterministic_loss.py
@@ -60,12 +60,12 @@ def generate_deterministic_losses(
     ri = any(re.match(r'RI_\d+$', fn) for fn in os.listdir(input_dir))
 
     step_flag = ''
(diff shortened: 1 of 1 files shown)


oasislmf/utils/deterministic_loss.py
```diff
@@ -60,12 +60,12 @@ def generate_deterministic_losses(
     ri = any(re.match(r'RI_\d+$', fn) for fn in os.listdir(input_dir))
 
     step_flag = ''
-    if 'fm_profile.csv' in os.listdir(input_dir):
-        with open(os.path.join(input_dir, 'fm_profile.csv')) as f:
-            reader = csv.reader(f)
```
sr-murthy: It is probably best to avoid using the `csv` lib. - Pandas is the best for processing CSVs, even for simple reading and writing.

PR #506, opened 2020-02-11
Improve performance

JBADougBurns:
We ran a 3M lat-lon exposure portfolio and ran out of memory when write_exposure_summary() was called in summaries.py. A look into it shows this mainly comes from the full oed portfolio which has been read into memory (200+ columns and about 5GB) merging with GUL inputs and perils. I have imposed column restrictions based on the columns required, reducing the exposure down to 50MB  (in the same vein as https://github.com/OasisLMF/OasisLMF/pull/440/files).

In addition, albeit a very minor implementation... I have added a function which drops a group peril in the exposure if its constituents are already present. i.e. If OO1;OSF;ORF is present, then modify to OSF;ORF

### DIFF
--- a/oasislmf/model_preparation/summaries.py
+++ b/oasislmf/model_preparation/summaries.py
@@ -23,6 +23,7 @@ from ..utils.data import (
     merge_dataframes,
     set_dataframe_column_dtypes,
     get_dtypes_and_required_cols,
+    reduce_df
 )
 from ..utils.defaults import (
     find_exposure_fp,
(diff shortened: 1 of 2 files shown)


oasislmf/model_preparation/summaries.py
```diff
@@ -823,3 +841,32 @@ def write_exposure_summary(
...
+    :return: dataframe with peril group code removed if all component perils are provided
+    :rtype: pandas.DataFrame
+    """
+    unique_loc_per_cov = exposure_df[loc_per_cov].unique()
+    to_replace = {}
+    for exposure_perils_set in unique_loc_per_cov:
+        exposure_perils = exposure_perils_set.split(';')
+        for peril_group in peril_groups:
+            if peril_group in exposure_perils and \
+                    all(peril in exposure_perils for peril in peril_groups[peril_group]):
```
hchagani-oasislmf: Apparently, it would be more "pythonic" to break line as follows:
```
if peril_group in exposure_perils and all(
    peril in exposure_perils for peril in peril_groups[peril_group]
):
```
Very minor issue; I can change this before merging.

oasislmf/model_preparation/summaries.py
```diff
@@ -717,23 +718,41 @@ def write_exposure_summary(
...
     """
+    loc_per_cov = oed_hierarchy['locperilid']['ProfileElementName'].lower()
+    exposure_df = reduce_df(
+        exposure_df,
+        cols=[loc_per_cov, 'loc_id']
+    )
+    gul_inputs_df = reduce_df(
+        gul_inputs_df,
+        cols=['peril_id', 'coverage_type_id', 'loc_id', 'tiv']
+    )
```
hchagani-oasislmf: Could also be achieved with:
```
gul_inputs_df = gul_inputs_df[['peril_id', 'coverage_type_id', 'loc_id', 'tiv']]
```
or:
```
gul_inputs_df = gul_inputs_df.loc[:, ['peril_id', 'coverage_type_id', 'loc_id', 'tiv']]
```
However, separate function does look tidier, so no objection.

PR #512, opened 2020-02-17
removing commas at the end of lines bring back lmf to work

nastasi-oq:
Trying to port a model previously developed with LFM version 1.4.0, using 1.5.1 we found some trouble and investigating our and your code we found what could be the source of our problems.


sambles:
Looks good, thanks for the fix!

PR #527, opened 2020-03-04
Group OED fields from model settings

hchagani-oasislmf:
~~Groups can be assigned with OED fields specified in `analysis_settings.json`. The `analysis_settings.json` file is generated when using the UI from `model_settings.json`. The schema for `model_settings.json` has been updated to incorporate this.~~

Command line arguments for group fields take priority, followed by those defined by the model. If neither exist, `loc_id` is used as the fall-through option.

See comment below for changes.

### DIFF
--- a/oasislmf/cli/model.py
+++ b/oasislmf/cli/model.py
@@ -179,13 +179,13 @@ class GenerateOasisFilesCmd(OasisBaseCommand):
         super(self.__class__, self).add_args(parser)
 
         parser.add_argument('-o', '--oasis-files-dir', default=None, help='Path to the directory in which to generate the Oasis files')
+        parser.add_argument('-a', '--analysis-settings-json', default=None, help='Analysis settings JSON file path')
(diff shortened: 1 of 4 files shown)


hchagani-oasislmf:
Groups can be assigned with OED fields specified in analysis_settings.json. The analysis_settings.json file is generated when using the UI from model_settings.json. The schema for model_settings.json has been updated to incorporate this.

Groups can now be defined as a property of the model. This is done through the `group_fields` key in `model_settings.json` under `data_settings`. When generating oasis files or running the model end to end, the argument `-M` or `--model-settings-json` should point to the `model_settings.json` file to use this. For example:

```
$ oasislmf model generate-oasis-files --model-settings-json meta-data/model_settings.json -C oasislmf.json
```

Command line arguments for group fields take priority, followed by those defined by the model. If neither exist, `loc_id` is used as the fall-through option.

Issue #550, opened 2020-04-23
full_correlation option generates bug in run_ktools.sh


johanxnilsson:
The attached analysis_setting makes oasis create an incorrect run_ktool.sh file with a missing space character in "fmcalc-a2". This makes the run fail due to incorrect command.

[stderror.err.txt](https://github.com/OasisLMF/OasisLMF/files/4522451/stderror.err.txt)
[run_ktools.sh.txt](https://github.com/OasisLMF/OasisLMF/files/4522444/run_ktools.sh.txt)
[analysis_settings.json.txt](https://github.com/OasisLMF/OasisLMF/files/4522445/analysis_settings.json.txt)

Issue #548, opened 2020-04-20
GUL Alloc Rule 0 doesn't work in oasislmf package


benhayes21:
line 672 in model_execution/bash.py uses:

`gul_item_stream = (gul_alloc_rule and isinstance(gul_alloc_rule, int))`

which interprets gul_alloc_rule as False when gul_alloc_rule = 0, so the gul_item_stream is not used.

line 273 has a similar condition:

`if gul_alloc_rule:`

which is also interpreted as False for gul_alloc_rule = 0

sambles:
Issue Notes: 
Originally `-a0` was used to selected the now deprecated  (item/coverage) stream type.  

Now that  `--ktools-alloc-rule-gul 0`  is an valid allocation rule:
*  The option needs to be changed to add `gulcalc -a0 -i - ` using the new stream type
* A separate option added to change to the old stream types (only if needed)

sambles:
Note: 
* default == 0
* option should be defined in model_settings

sambles:
Added a new flag **--ktools-legacy-gul-stream** to revert to deprecated item/coverage, this overrides any options set for **--ktools-alloc-rule-gul**

Issue #545, opened 2020-04-17
Logging in the CLI


mpinkerton-oasis:
The default logging in the CLI quite verbose, writing out all the function entry/exits. Would it be an issue to make the default logging debug?

Issue #437, opened 2019-11-15
Invalid coverage file generated


benhayes21:
If an OED loc file has many rows with the same portnumber, accnumber and locnumber, it seems that duplicate coverage ids are generated in the coverage file. These are potentially valid OED files (there can be different peril terms for a single location across lines) but break the MDK

sambles:
Added handing so that when duplicated `loc_id`s are found, the MDK sends a warning the to user.
```
WARNING: Duplicate keys ['portnumber', 'accnumber', 'locnumber'] detected in location file
	 oasislmf doesn't currently support multiple terms for a single location
	 dropping the following row(s): [6, 7]
```
This is a placeholder fix until we decide how to handle multiple terms per location.

sambles:
Closing this and moving it to #558

PR #554, opened 2020-04-30
Fix/collected issues may

sambles:
* Update to the API client, try the reference user/password before prompting the user for input

* Fixed issue #550 - full_correlation option generates bug in run_ktools.sh

* Fixed issue #548 - GUL Alloc Rule 0 doesn't work in oasislmf package
    Added a new flag **--ktools-legacy-gul-stream** to revert to deprecated item/coverage, this overrides any options set for `--ktools-alloc-rule-gul`

* Partial fix for #437 - Invalid coverage file generated

1) Currently if duplicated loc_id are found, send a warning the to user and strip out these location rows.  This is a placeholder fix until we decide how to handle multiple terms per location.
```
WARNING: Duplicate keys ['portnumber', 'accnumber', 'locnumber'] detected in location file
	 oasislmf doesn't currently support multiple terms for a single location
	 dropping the following row(s): [6, 7]
```

2) `loc_id` is now set based on the sorted values of `['portnumber', 'accnumber', 'locnumber']` rather than the location file's row order 

* Fix for #545 -  Reduce the standard log verbosity, removed `__init__` functions and exit timestamps (Still visible when using `--verbose`)  

@johcarter  if you have time, could you double check the changed expected FM outputs in? 82246d2da2d979b4da731752033cae51bc5aeeac


oasislmf/model_execution/runner.py
```diff
@@ -68,7 +69,7 @@ def custom_get_getmodel_cmd(
                 max_process_id,
                 os.path.abspath("analysis_settings.json"),
                 "input")
-            if coverage_output != '' and not gul_alloc_rule:
+            if coverage_output != '' and gul_legacy_stream:
```
hchagani-oasislmf, line 72: Would it be more readable to switch the order?:

```suggestion
            if gul_legacy_stream and coverage_output != '':
```

oasislmf/utils/data.py
```diff
@@ -429,13 +428,38 @@ def get_ids(df, usecols, group_by=[]):
     :param group_by: A subset of the column subset to use a subgroup key
     :param group_by: list
 
+    :param allow_duplicates: allow non-unique ids in return
+    :param allow_duplicates: Boolean
+
```
hchagani-oasislmf, line 433: Parameter `allow_duplicates` not an argument in `get_ids` function. Suggest removing it to avoid confusion:
```suggestion
```

sambles:
@hchagani-oasislmf  Requested changes added

PR #600, opened 2020-06-23
Test Data Generation

hchagani-oasislmf:
For performance testing, generates the following files with dummy data:

```
input/coverages.bin
input/events.bin
input/fm_policytc.bin
input/fm_profile.bin
input/fm_programme.bin
input/fmsummaryxref.bin
input/fm_xref.bin
input/gulsummaryxref.bin
input/items.bin
input/occurrence.bin
static/damage_bin_dict.bin
static/footprint.bin
static/footprint.idx
static/random.bin
static/vulnerability.bin
```

Currently uses standalone script `generate_test_data.py` with arguments to generate aforementioned files. Generating the files and running them through `ktools` routines can be done by running `gen_data_and_run_ktools.sh`.

Intention is to integrate performance testing with MDK at a later date.


oasislmf/gen_test_data/generate_test_data.py
```diff
@@ -0,0 +1,725 @@
...
+import os
+import struct
+
+
+class ModelFile:
+    def __init__(self):
+        pass
+
+    def seed_rng(self):
+        if self.random_seed == 0:
```
sstruzik, line 13: This condition could be included in the else

hchagani-oasislmf: Sorry, that is a mistake. That value should use current system time as random seed. See suggestion below for change.

oasislmf/gen_test_data/generate_test_data.py
```diff
@@ -0,0 +1,725 @@
...
+            np.random.seed(0)
```
hchagani-oasislmf, line 14:
```suggestion
            np.random.seed()
```

sstruzik: yes make sens

oasislmf/gen_test_data/generate_test_data.py
```diff
@@ -0,0 +1,725 @@
...
+        elif self.random_seed == -1:
+            np.random.seed(1234)
+        else:
+            np.random.seed(self.random_seed)
+
+    def write_file(self):
+        with open(self.file_name, 'wb') as f:
+            if self.start_stats:
+                for k, v in self.start_stats.items():
```
sstruzik, line 23: is the key in start_stats ever used, I think you can simply use a list for it

oasislmf/gen_test_data/generate_test_data.py
```diff
@@ -0,0 +1,725 @@
...
+                    f.write(struct.pack(v['dtype'], v['value']))
+            for col in self.data:
+                for idx, dtype in enumerate(self.dtypes.values()):
+                    try:
+                        f.write(struct.pack(dtype, col[idx]))
+                    except TypeError:   # If line is not tuple
+                        f.write(struct.pack(dtype, col))
```
sstruzik, line 30: you can use struct like this
>>> col = (1,2,3)
>>> struct.pack('idi', *col)
b'\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x03\x00\x00\x00'
you don't need to iterate
I would also enforce self.data to yield tuples to remove the usage of the try except

oasislmf/gen_test_data/generate_test_data.py
```diff
@@ -0,0 +1,725 @@
...
+
+    # Method for debugging file output
+    def debug_write_file(self):
```
sstruzik, line 33: I think in this debug part, it would be interesting to put a header,
Also I would use format, something like
line_format = "{}"+", {}"*(len(self.dtypes)-1)
for line in self.data:
  print(line)

sstruzik:
For the write_file you always surcharge the method:
with something like
    def write_file(self, input_dir):
        self.file_name = os.path.join(input_dir, 'fm_profile.bin')
        self.data = self.generate_data()
        super().write_file()

I think you can generalise the function by putting those step in the parent class method.
def write_file(self, directory):
    self.file_path = os.path.join(directory, self.filename)
    ....
    for col in self.generate_data():
        ...

self.filename would be for example vulnerability.bin and moved into the init
and for the few file class like RandomFile, you put the specific code in generate_data()

you'll have to change the debug_write_file too

alternatively, I think you can change the name of generate_data to data and set it as a property

sstruzik:
For the seeds, (this is actually the most important comment I have to make)
If I understand correctly, if you set it to a value other than 0, you will then have all the random generator starting with the same seed. For me this is not ideal as I believe you will have totally correlated data.
>>> np.random.seed(1)
>>> np.random.randint(10000000)
4686059
>>> np.random.seed(1)
>>> np.random.randint(10000000)
4686059

to have reproducibility without this tight correlation, one simple idea for our case is to use "salt" meaning you can add a value to the seed that will depend on the file class.
Or you can prepare a list of files that you want to create and generate a seed from your initial seed for each of them

hchagani-oasislmf:
Thanks @sstruzik for the review. I've implemented the changes you suggested above.

Issue #578, opened 2020-06-02
missing combination of terms in calcrules to add


johcarter:
-	Amount deductible + % of loss limit
-	% of loss (or % of TIV) ded + min and/or max ded + amount limit
-	Amount deductible + % of TIV limit

johcarter:
check for other missing combos.

johcarter:
Amount deductible + % Loss limit  - caclrule 15
Unlimited layer - calcrule 34

johcarter:
% of loss (or % of TIV) ded + min and max ded + amount limit - calcrule26

johcarter:
calcrule 35 to add to calc_rules.csv and test

johcarter:
identified outstanding missing combos and raised https://github.com/OasisLMF/OasisLMF/issues/606

Issue #566, opened 2020-05-21
Handle unlimited LayerLimit without large default value


johcarter:
In OED, a zero LayerLimit means unlimited, whereas a zero LayerParticipation means 100%. We use calcrule 2 for layers. 
Rather than insert a large default for limit, it would be better to use a calcrule without a limit parameter to avoid having to put in a large default value in the limit field.

johcarter:
done - testing to complete

PR #605, opened 2020-06-30
Feature/extend calcrules

johcarter:
https://github.com/OasisLMF/OasisLMF/issues/566
https://github.com/OasisLMF/OasisLMF/issues/578


johcarter:
yes, that is an expected change. it used to be set to be the same as sidx 1, but now that we support multiple samples i think it has been changed to 0.  The reporting (loc_summary.csv) is done on sidx > 1 and these are unchanged / floating point differences.

PR #619, opened 2020-07-24
Include coverages unsupported by model in type 2 financial terms calculation

hchagani-oasislmf:
This change only applies when the default FM aggregation profile is used. Otherwise, previous behaviour remains unchanged as coverages that are not supported for the model are not included in the `keys.csv` file and therefore it is not possible to implement a generalised algorithm for aggregation. Currently, no users implement their own FM aggregation profile to our knowledge.


oasislmf/model_preparation/il_inputs.py
```diff
@@ -773,6 +788,60 @@ def assign_calcrule_flag(row):
...
+                subset=['loc_id', 'peril_id', 'orig_level_id', 'agg_id'],
+                keep='first'
+            ).groupby([
+                'peril_id', 'orig_level_id', 'is_bi_coverage', 'agg_id'
+            ])[['tiv_sum', 'bitiv']].sum()
+        )
+        intermediate_fm_agg_tivs.reset_index(inplace=True)
+        intermediate_fm_agg_tivs['agg_tiv'] = intermediate_fm_agg_tivs['tiv_sum']
+        # Adjust agg_tiv for any levels that use BI coverage aggregation key
+        bi_coverage_levels = [
```
sstruzik, line 815: you should use a generator here

PR #622, opened 2020-07-30
Feature/add tests

johcarter:
Added more reinsurance tests


johcarter:
the reinsurance input files need to be ri_info and ri_scope - they get relabelled when copied to /run as reinsinfo and reinsscope which is why they didn't work when you put them in input. I have relabelled them - would you like to try again please @sambles ? thanks

PR #635, opened 2020-08-21
Extract default values for OED mapped FM terms and apply to missing columns

hchagani-oasislmf:
Default values for OED mapped FM terms are currently hard coded in `preparation/il_inputs.py`. These values can be extracted from `_data/acc_dtypes.json` and `_data/loc_dtypes.json` instead. Not only does this make the code a bit more robust, but solves the issue when `LayerParticipation` is missing or 0 in the acc file.


sambles (approved):
Looks good

PR #637, opened 2020-08-26
Feature/636 improve error guard

sambles:
* Always run process script cleanup (even on success)
* Add ktools process complete check 
* Fix ktools script logging with `--verbose`  flag


sstruzik (approved):
It's a good idea to pack all the multiple print_commands into one block but I don't think you really need the extra function do_script_block, you can actually use the existing print_command that is the same beside '\n'.
Otherwise looks good

Issue #640, opened 2020-08-27
Reduce the memory usage of oasis files generation.


sambles:
An idea from **Daniel Evans** is to use the pandas categorical datatype
https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html

Example: 
https://github.com/ianozsvald/dtype_diet

PR #641, opened 2020-08-27
Use pandas' categorical dtype for strings

DanielFEvans:
#640 

Pandas' categorical datatypes are good for reducing memory where a large proportion of values are non-unique. Rather than storing every single value, each unique value is stored and a pointer to it placed in the data array. It seems the functionality is almost completely transparent.

I was motivated to try this after finding out that some string-typed columns with very few unique values (e.g. AccNumber, PortNumber, CountryCode) were responsible for a surprisingly large fraction of my portfolio's memory footprint, with the [DType Diet](https://github.com/ianozsvald/dtype_diet) script suggesting the memory use could be greatly reduced by categoricals.

The change here is simply to tell Pandas to use dtype 'category' for any OED field that is defined as 'str' - the rest just falls into line. It appears that Pandas always forces categorical value to be strings, even for all-integer inputs, which avoids some potential issues.

I've given this a test with our model, and there's nothing obvious that goes wrong - our tests involving Oasis still pass, and it's quite happily chugging through a large portfolio at the moment.


DanielFEvans:
Right, I think that's now vaguely in working order - I had fun fighting with a bug within Pandas 1.1.x, so I've set the requirements files to exclude 1.1.0 and 1.1.1. v1.1.2 will have a fix for the bug I encountered.

The one concern I have is that with categorical dtypes, it's awkward to add a new, not-before-seen value to a column - you have to add it to the list of allowed values in the category. I've handled the few cases where I've found that being needed in the OasisLMF code, but as the Loc dataframe is being passed on to external software, someone downstream might get a nasty surprise.

For our uses, I can't think of anywhere where our model code ever alters OED values - we instead generate new ones with our model data. I've run through some reasonably sized analyses without encountering any issues.

oasislmf/utils/data.py
```diff
@@ -547,6 +548,12 @@ def get_dataframe(
         if missing:
             raise OasisException('Missing required columns: {}'.format(missing))
 
+    # Force categorical column values to string
```
sstruzik, line 551: I'm not sure if this is useful, I don't see change in the dtype of the category
>>> import pandas as pd
>>> import io
>>> a = io.StringIO("a,b,c\n1,a,1\n2,a,1")
>>> df = pd.read_csv(a, dtype={"b": 'category', 'c':'category'})
>>> df["c"].cat.categories
Index(['1'], dtype='object')
>>> df["c"].cat.categories = df["c"].cat.categories.astype(str)
>>> df["c"].cat.categories
Index(['1'], dtype='object')
I didn't see it forbid int values for example
>>> df['c'].cat.add_categories([2], inplace=True)
>>> df['c'][0]=2
>>> df
   a  b  c
0  1  a  2
1  2  a  1

oasislmf/utils/data.py
```diff
@@ -1042,3 +1048,41 @@ def set_dataframe_column_dtypes(df, dtypes):
...
+            continue
+
+        col = df[col_name]
+        if pd.api.types.is_categorical_dtype(col):
+            # Force to be a string - using categorical for string columns
+            value = str(value)
+            fill_value[col_name] = value
+
+            if value not in col.cat.categories:
+                fill_value[col_name] = value
```
sstruzik, line 1081: you can remove this line.

oasislmf/utils/data.py
```diff
@@ -1042,3 +1048,41 @@ def set_dataframe_column_dtypes(df, dtypes):
...
+                col.cat.add_categories([value], inplace=True)
+
+    # Note bug in Pandas 1.1.0/1.1.1 for fillna(inplace=True) when categorical dtypes are involved
+    # It doesn't hit here, but leads to very confusing errors internal to Pandas later on
+    # https://github.com/pandas-dev/pandas/issues/35731
+    df.replace('nan', np.nan, inplace=True)
```
sstruzik, line 1087: Sorry, not sure to understand this specific case. 'nan' should be evaluated as NaN=np.nan during the loading and the issue mentioned shouldn't be present as you exclude the bugged version.

oasislmf/utils/data.py
```diff
@@ -1042,3 +1048,41 @@ def set_dataframe_column_dtypes(df, dtypes):
     df = df.astype(_dtypes)
 
     return df
+
+
+def fill_na_with_categoricals(df, fill_value):
```
sstruzik, line 1053: As this function is "public"/ imported in summaries, can you add it the the list in __all__ at the top of the file

sstruzik (changes requested):
Hi Daniel,
Thanks for this PR. I think this is an excellent idea to use categories in those cases.
Implementation is good overall, I just have few places I'm not sure to understand what the code is bringing.
regards,
Stephane

oasislmf/utils/data.py
DanielFEvans: I'll try to remind myself of where this turned out to be a problem. The reasoning is that we want the category values to be string - there's been plenty of fun where one column is of type "int" and the other "string", and Pandas won't merge `1` with `"1"`. However, I recall this being a rather awkward case that didn't always occur - as in your example, `type(df['c'][0].iat[0])` returns `<class 'str'>`.

(It may have been a complication of Oasis replacing special/null values with OED defaults like `1`)

oasislmf/utils/data.py
DanielFEvans: This one was brought up by the Oasis tests. When using 'categorical' as the data type, it seems that Pandas _doesn't_ automatically do that mapping. The test was therefore inputting 'nan' as a string, expecting NaN back, but was actually getting 'nan' again.

The comment about the bug is separate to the actual code (but boy, did it make debugging the problem fun) - I'll make it all a little clearer.

DanielFEvans: (As with my other comment, getting Pandas not to map 'nan' again seems to have been an awkward edge case, as a simple CSV test actually shows it does convert as expected. I'll work out the exact cause and note it down!)

oasislmf/utils/data.py
DanielFEvans: Amusingly, I've not been able to reproduce any problem by removing these lines. I suspect that this was working around another issue.

oasislmf/utils/data.py
DanielFEvans: As with the other section, I can't reproduce the original issue I was trying to fix - it may have been part of the confusion caused by that Pandas bug.

sstruzik (approved):

PR #642, opened 2020-08-28
Calc. rules overhaul

hchagani-oasislmf:
Replace `_data/calc_rules.csv` with `_data/calc_rules_direct.csv` for all coverage levels except `policy layer` and `_data/calc_rules_direct_layer.csv` for `policy layer` level. When assigning calc. rules, deductible and limit types 2 are converted to type 0 as the calc. rules are the same, reducing duplication.


johcarter:
descriptions have been added to calcrules_direct and all tests pass @hchagani-oasislmf

sstruzik (approved):

PR #643, opened 2020-08-28
fix for supplier_model_runner

sstruzik:


hchagani-oasislmf (approved):
Looks good. Thanks for changing that so quickly.

PR #644, opened 2020-09-09
Fix issue with calc. rules merger

hchagani-oasislmf:
Actual source of problem lies with way aggregated TIVs are merged into `il_inputs_df` in `preparation.il_inputs.get_il_input_items()`. Merger needs to occur on `is_bi_coverage` in addition to remove erroneous duplicate rows.

### DIFF
--- a/oasislmf/preparation/il_inputs.py
+++ b/oasislmf/preparation/il_inputs.py
@@ -127,7 +127,8 @@ def set_calc_rule_ids(
         ].transpose().values)
     ]
     il_inputs_calc_rules_df = merge_dataframes(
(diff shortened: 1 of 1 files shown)


sambles (approved):
Looks good, the fix matches the output from `1.9.1`

sstruzik (approved):
looks good if we are sure this column is present in the df

hchagani-oasislmf:
> looks good if we are sure this column is present in the df

I think so: I've been looking through `gul_inputs.py`, which is where `is_bi_coverage` is introduced. It looks like the column must always exist if `gul_inputs_df.groupby(by=['coverage_type_id'], sort=True)` is not an empty dataframe (which is shouldn't be). I think it is alright.

Issue #568, opened 2020-05-22
Min/max deductible not working when prior level locations are partially limited in OED input


johcarter:
We have an issue  where there is a min and/or max deductible (say at policy level) which is not being applied when the prior level locations are _**partially**_   limited.

- If there are no prior  level limits on any coverage or sites min and max deductible is applied correctly
- If there are prior  level limits on all coverages or site, min and max deductible is applied correctly
- If there are prior  level limits on some coverages or sites but not all the min or max deductible is **_not_** applied  correctly

Until this is fixed in ktools, the workaround is to fill in gaps in limits in the OED location file using a suitably large number.
https://github.com/OasisLMF/ktools/issues/123

--

PR #645, opened 2020-09-11
Fix/568 unlimited loss

johcarter:
Fixes Min/max deductible not working when prior level locations are partially limited in OED input #568


hchagani-oasislmf (approved):
Looks good.

PR #655, opened 2020-09-28
Reduce number of layers being applied by fac contracts

hchagani-oasislmf:
If `ReinsType = 'FAC'`, the layer number is not incremented unless entry in reinsurance info file is duplicate, determined according to risk level. This reduces the length of `fm_policytc.csv` when multiple FAC contracts are present, removing unnecessary pass-through policy TC IDs.


benhayes21 (approved):

sambles (approved):
Looks good

PR #668, opened 2020-10-28
Feature/fm rework

sstruzik:
Add an optional new implementation of fmcalc in python (fmpy) for insurance loss computation.
to activate this option add the flag --fmpy in the mdk
to run fmpy in low memory mode add the flag --fmpy-low-memory
fmpy manage alloc rule 0, 1, 2
fmpy manage all basic insurance policies (1 to 26, 33 to 36, 100)


sambles (approved):
Changes to the shared common files look good

Issue #662, opened 2020-10-07
model settings to define relationships between event and occurrence options


benhayes21:
occurrence files are often tied to event set options, and it would be good if this relationship/dependency was represented in the model settings file in some way.

Example, Model A has two event sets (one probabilistic and one historical) and three occurrence files (two options for the probabilistic events and one for the historic event set). The model settings should make available this information so that a user can see which files should go together. A UI might interpret this and only allow certain combinations.

sambles:
**Note:** one way link from event set to occurrence

PR #673, opened 2020-11-04
Feature/update model settings schema

sambles:
### Issue #662 - define relationships between event and occurrence sets
Each event_set section has an optional list of strings `valid_occurrence_ids` that represents the relationship/dependency between the two sets of files.

 **Example:**
```
   "model_settings":{
      "event_set":{
         "name":"Event Set",
         "default":"A",
         "desc": "Lists of event sets",
         "options":[
            {   
               "id":"A",
               "desc":" ... ",
               "valid_occurrence_ids":["1", "2"]
            },  
            {   
               "id":"B",
               "desc":" ... ",
               "valid_occurrence_ids":["2", "3"]
            }   
         ]   
      },  
      "event_occurrence_id":{
         "name":"Occurrence Set",
         "desc":" ... ",
         "default":"1",
         "options":[
            { "id":"1", "desc": "..",  "max_periods": 10000},
            { "id":"2", "desc": "..",  "max_periods": 20000},
            { "id":"3", "desc": "..",  "max_periods": 30000},
            { "id":"4", "desc": "..",  "max_periods": 40000},
         ]
      }, 
```

### Issue #400 - Add extra optional parameters to model_settings
Expanded the model settings file with addition metadata for hosting models. 

#### 1. Meta-data for noting which model a settings file applies to 
```
{                                                                                                                                                              
   "name": "ModelName",
   "description": "Longer model description",
   "supplier_id": "OasisLMF",
   "model_id": "piwind",
   "model_version_id": "1",
   "model_settings":{
       ...
```

#### 2. Extra information on model data and assets
```
{
    "model_settings":{
       ...
    },
    "data_settings":{
        "model_data_version": "1.0.0",
        "keys_data_version": "1.0.0",
        "worker_image": "coreoasis/model_worker",
        "worker_version":"1.12.0",
        "uses_model_files": true,
        "countries": ["UK", "US"],
        "mandatory_fields": ["AreaName", "PostalCode", "OccupancyCode"],
        "additional_assets": [
            {    
               "name": "catresponsedata",
               "version": "1.0",
               "path": "model_data/catresponse/"
            }
        ],
   }
}
```

### Support complex model configuration in the OasisUI 

A new workflow has been added to the OasisUI, this lets user select and edit model_settings values before the file generation step. To enable a model add `"model_configurable": true,` to the `model_settings.json` file
```
{
     "model_configurable": true,
     "model_settings":{
           ...
```

Each parameter has a new **used_for** option with two valid values. Either `generation` or `losses`, which marks that parameter as only valid for that stage of the workflow. The assumption is that if an option applies to both then no option needs to be set. 

**Example:**
```
 "boolean_parameters":[
         {                                                                                                                                                     
            "name":"A",
            "desc":"Boolean option",
            "default":false,
            "used_for": "generation"
         },  
         {   
            "name":"B",
            "desc":"Boolean option",
            "default":true,
            "used_for": "losses"
         },
         {   
            "name":"C",
            "desc":"Boolean option",
            "default":true
         }   
      ], 
```

### DIFF
--- a/oasislmf/schema/example_model_settings.json
+++ b/oasislmf/schema/example_model_settings.json
@@ -7,7 +7,8 @@
          "options":[
             {
                "id":"P",
-               "desc":"Probabilistic"
+               "desc":"Probabilistic",
+               "valid_occurrence_ids":["1"]
             },
             {
                "id":"H",
--- a/oasislmf/schema/example_model_settings.json
+++ b/oasislmf/schema/example_model_settings.json
@@ -1,4 +1,11 @@
 {
+   "name": "MyModelName",
+   "description": "Longer model description",
+   "supplier_id": "OasisLMF",
+   "model_id": "piwind",
+   "model_version_id": "1",
+   "model_configurable": true,
+
    "model_settings":{
       "event_set":{
          "name":"Event Set",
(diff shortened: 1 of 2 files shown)


ndaqjn:
Looks good to me

PR #674, opened 2020-11-09
Introduce check for step policies in genbash

hchagani-oasislmf:
@benhayes21 : does this do the trick?

@sambles : if the code looks good to you, and Ben's happy with the fix, shall I start a patch release?


sambles (approved):
Looks good

PR #677, opened 2020-11-11
fmpy: manage multi stream in

sstruzik:
Enable Fmpy to handle multiple stream_in
Use python high level Selector class instead of select

### DIFF
--- a/oasislmf/pytools/fm/manager.py
+++ b/oasislmf/pytools/fm/manager.py
@@ -1,5 +1,5 @@
 from .financial_structure import create_financial_structure, load_financial_structure
-from .stream import read_stream_header, read_event, EventWriter, EXTRA_VALUES
+from .stream import read_stream_header, read_streams, EventWriter, EXTRA_VALUES
 from .compute import compute_event, init_variable, reset_variabe
 from .common import allowed_allocation_rule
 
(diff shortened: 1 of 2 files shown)


sambles (approved):
Tested the new option in my environment and it looks good. 

**Notes for reference:** 
Feature allows fmpy to accept multiple gul streams without the load balancer (lb). 
Currently the lb is slightly faster due to buffering so the change has not been added to bash generation. If added it would look like
```
eve 1 4 | getmodel | gulcalc -S10 -L0 -a0 -i - | tee fifo/gul_P1 > fifo/gul_lb_P1 &
eve 2 4 | getmodel | gulcalc -S10 -L0 -a0 -i - | tee fifo/gul_P2 > fifo/gul_lb_P2 &
eve 3 4 | getmodel | gulcalc -S10 -L0 -a0 -i - | tee fifo/gul_P3 > fifo/gul_lb_P3 &
eve 4 4 | getmodel | gulcalc -S10 -L0 -a0 -i - | tee fifo/gul_P4 > fifo/gul_lb_P4 &
fmpy -i fifo/gul_lb_P1 fifo/gul_lb_P2 fifo/gul_lb_P3 fifo/gul_lb_P4 -o fifo/lb_il_P1 -a2 | tee fifo/il_P1 | fmpy -a3 -n -p RI_1 > fifo/ri_P1 & 
```

PR #682, opened 2020-11-17
fmpy handle stepped policies

sstruzik:
- Add step policy functionality to fmpy
- fix issue with a1 allocation rule


johcarter:
it produces results for fm54 without changing results for anything else so i'd say its good to merge if the integration tests pass. i'll do more detailed testing next week

johcarter (approved):

Issue #678, opened 2020-11-12
Make env vars take precendence over oasislmf.json parameters for containers


ndaqjn:
When the ktools chain is running inside a container, it's important to be able to override settings from the oasislmf.json using environment variables and to make sure that the env vars takes precedence over the config.ini settings.

PR #683, opened 2020-11-17
Fix ENV var loading and add override option for worker

sambles:
Fix for issue #678 
* Fixed problem with environment variables not loading
* Added option for worker `OASIS_ENV_OVERRIDE` which changes the load order so environment variables take priority 

**Example**
```
$ export OASIS_ENV_OVERRIDE=true
$ export OASIS_ANALYSIS_SETTINGS_JSON=analysis_settings.json
$ export OASIS_KTOOLS_NUM_PROCESSES=5
$ oasislmf model run --ktools-num-processes 66

Processing arguments - RunModel
Warning - environment variable override: OASIS_ANALYSIS_SETTINGS_JSON=/home/sam/repos/models/piwind/analysis_settings.json
Warning - environment variable override: OASIS_KTOOLS_NUM_PROCESSES=5
  .... 

2020-11-17 16:52:04,929 - root - DEBUG -     number_of_processes == 5
2020-11-17 16:52:04,929 - root - DEBUG -     filename == /home/sam/repos/models/piwind/runs/losses-20201117165159/run_ktools.sh
2020-11-17 16:52:04,929 - root - DEBUG -     num_reinsurance_iterations == 1
2020-11-17 16:52:04,929 - root - DEBUG -     set_alloc_rule_gul == 0
2020-11-17 16:52:04,929 - root - DEBUG -     set_alloc_rule_il == 2
2020-11-17 16:52:04,929 - root - DEBUG -     set_alloc_rule_ri == 3
2020-11-17 16:52:04,929 - root - DEBUG -     num_gul_per_lb == 0
2020-11-17 16:52:04,929 - root - DEBUG -     num_fm_per_lb == 0
2020-11-17 16:52:04,929 - root - DEBUG -     run_debug == False
2020-11-17 16:52:04,929 - root - DEBUG -     stderr_guard == True
2020-11-17 16:52:04,929 - root - DEBUG -     gul_legacy_stream == False
2020-11-17 16:52:04,929 - root - DEBUG -     fifo_tmp_dir == True
2020-11-17 16:52:04,929 - root - DEBUG -     custom_gulcalc_cmd == None
2020-11-17 16:52:04,929 - root - DEBUG -     fmpy == False
2020-11-17 16:52:04,929 - root - DEBUG -     fmpy_low_memory == False
```

### DIFF
--- a/oasislmf/utils/inputs.py
+++ b/oasislmf/utils/inputs.py
@@ -87,12 +87,14 @@ class InputValues(object):
             self.logger.error('\nexiting.')
 
     def get_env(self, name, dtype=None, default=None):
-        prefix = 'OASIS'
-        env_var = os.getenv(f'{prefix}_{name.upper()}', default=default)
+        env_var = os.getenv(f'OASIS_{name.upper()}', default=default)
         if dtype and env_var:
             return dtype(env_var)
         else:
             return env_var
+
+    def has_env(self, name):
+        return f'OASIS_{name.upper()}' in os.environ
 
     def get(self, name, default=None, required=False, is_path=False, dtype=None):
         """
@@ -132,7 +134,7 @@ class InputValues(object):
         value = getattr(self.args, name, None)
 
         # Load order 1: ENV override (intended for worker images)
-        if str2bool(os.getenv('OASIS_ENV_OVERRIDE', default=False)) and name != 'config':
+        if str2bool(os.getenv('OASIS_ENV_OVERRIDE', default=False)) and self.has_env(name):
             source = 'env_override'
             value = self.get_env(name, dtype)
 
--- a/oasislmf/utils/inputs.py
+++ b/oasislmf/utils/inputs.py
@@ -162,9 +162,9 @@ class InputValues(object):
             else:
                 value = os.path.abspath(value)
 
-        # Warn user of Enviroment variable load
+        # Warn user of environment variable load
         if source == 'env_override':
-            self.logger.warn(f'Warning - enviroment variable override: OASIS_{name.upper()}={value}')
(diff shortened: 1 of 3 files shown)


oasislmf/utils/inputs.py
```diff
@@ -110,31 +119,53 @@ def get(self, name, default=None, required=False, is_path=False):
...
 
         :return: The found value or the default
         """
-        value = getattr(self.args, name, None)
+        # Load order 0:  Get from CLI flag
         source = 'arg'
+        value = getattr(self.args, name, None)
+
+        # Load order 1: ENV override (intended for worker images)
+        if str2bool(os.getenv('OASIS_ENV_OVERRIDE', default=False)) and name != 'config':
```
sstruzik, line 135: Not sure here, should we check if value is None?
If user specify a value in the MDK, do we want to override it with env variable even if OASIS_ENV_OVERRIDE is True.

oasislmf/utils/inputs.py
```diff
@@ -85,7 +86,15 @@ def confirm_action(self, question_str, no_confirm=False):
         except KeyboardInterrupt:
             self.logger.error('\nexiting.')
 
-    def get(self, name, default=None, required=False, is_path=False):
+    def get_env(self, name, dtype=None, default=None):
+        prefix = 'OASIS'
+        env_var = os.getenv(f'{prefix}_{name.upper()}', default=default)
```
sstruzik, line 91: As you don't reuse prefix in this function, I would hardcode it in the string or make it a global variable if we want to indicate that all env variables we use need to be prefixed with OASIS

sstruzik (approved):
Looks good.
I've added few comments to think about but I don't have a strong opinion on the solution.

PR #684, opened 2020-11-19
Add flags for lookup multiproc options

sambles:
Added the following flags to the CLI commands: 
* run
* generate-oasis-files
* generate-keys 

`--lookup-num-processes`: Set the number of workers in the multiprocess pool, default is the number of logical cores available 
`--lookup-num-chunks` Set the number of chunks to break the location file into for assignment in the pool, default is 2 chunks per 'num-processes'


benhayes21:
Looks good to me

benhayes21 (approved):
Looks good to me

Issue #680, opened 2019-04-15
User defined return periods


mpinkerton-oasis:
Upload a set of return periods, for explicit generation in an analysis.

sambles:
Notes: 

Add list of periods option  (integers)  to the `analysis_settings.json` 

* Create a `return_periods.csv` from given list (stored)

sambles:
**analysis_settings.json**
```
{                                                                                                     
    "model_version_id": "PiWind",
    "module_supplier_id": "OasisLMF",
    "number_of_samples": 10, 
    "gul_threshold": 0,
    "return_periods": [10, 1000, 100],     <-- not sensitive to ordering 
       ...
}
```

**returnperiods.bin**
```
return_period
1000
100
10
```

PR #685, opened 2020-11-19
Feature/680 user set return periods

sambles:
#680 - Create a `returnperiods.bin`  from values given in `analysis_settings.json`

### DIFF
--- a/oasislmf/execution/bin.py
+++ b/oasislmf/execution/bin.py
@@ -26,6 +26,7 @@ import shutil
 import shutilwhich
 import subprocess
 import tarfile
+import pandas as pd 
 
 from itertools import chain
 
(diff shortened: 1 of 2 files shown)


oasislmf/execution/bin.py
```diff
@@ -247,7 +263,13 @@ def prepare_run_inputs(analysis_settings, run_dir, ri=False):
 
         # Prepare occurrence / returnperiod depending on output calcs selected
         if _calc_selected(analysis_settings, 'lec_output'):
-            _prepare_input_bin(run_dir, 'returnperiods', model_settings, ri=ri)
+            if 'return_periods' in analysis_settings:
```
sstruzik, line 266: In order to avoid failure if there is something like this in the settings
"return_periods": {} 
I would use:
if analysis_settings.get('return_periods'):
It feels safer

sstruzik (approved):
looks good.

oasislmf/execution/bin.py
sambles: ah yes fair point, I'll switch it over

PR #697, opened 2020-12-01
Set default value of optional OED step policy fields to 0

hchagani-oasislmf:
Set value to 0 of optional step policy fields that are missing from the acc file.


sstruzik (approved):
looks good

Issue #700, opened 2020-12-03
Update error guard to cover all ktools binaries 


sambles:
currently doesn't monitor `eltcalc`, `aalcalc`, and `leccalc` for output on stderr

PR #707, opened 2020-12-14
Feature/700 error guard update

sambles:
Update bash generation for issue #700 

Linked to: https://github.com/OasisLMF/ktools/issues/172  eltcalc can silently fail, returning partial events


benhayes21 (approved):
looks good

PR #718, opened 2021-01-12
Feature/fm special conditions sstruzik

sstruzik:
[DONE]
keep layer in gul df
manage layer from the condition level
manage all condition use case (multi layer and multi level)
fix write fm xref 
check coverage and limit/deductible special rule
integrate cleanly the step policy feature


johcarter (approved):
all tests pass for existing test cases. still need to test new functionality with additional test cases for special conditions but is ok to merge in. thanks

PR #726, opened 2021-01-18
Fix/712 gul errors map file dropped

benhayes21:
two changes made to:

 - add in location count by coverage type measure to exposure summary report
 - added back the gul error report to map rejected keys records


sambles (approved):
Looks good

PR #738, opened 2021-02-03
Fix setup.py fallback for ktools install in development mode

DanielFEvans:
When installing in development/editable mode, OasisLMF tries to install ktools from source, and if that fails, should fall back on installing pre-compiled binaries. However, the wrong function was called in the fallback branch.


DanielFEvans:
Per discussion with Sam, it seems even Oasis developers would prefer a binary install by default, so that is now the behaviour for both standard and developer installs.

sambles (approved):
Looks good

PR #742, opened 2021-02-03
Keys errors stream

DanielFEvans:
We ran into issues where the collection of tens millions of keys results into a list before writing the keys file exhausted memory - the overhead of so many dictionaries was too much.

The keys lookup interface is already such that each key is streamed back to Oasis in turn (or, at least, that's what we implemented!). This PR changes the CSV writing to stream these out, too. For JSON, they are still collected into a list, due to the awkwardness of JSON formatting (e.g. the last entry in each JSON output cannot have a final comma), but Ben stated that the JSON format is not used widely and is likely to be removed eventually.


sambles:
Changes in this PR will be rolled into a general Keys server update (WIP) by @sstruzik, closing this pull request for the moment.

PR #745, opened 2021-02-09
raise issue if locations have missing accounts

sstruzik:
check and raise an error if locations have an account that is missing in the account file


benhayes21 (changes requested):
Functionally, it's working fine...

The issue with returning the list of loc_ids is that when running through the UI, the generated gul_summary_map.csv file is not available for a failed generation, so the user will not be able to map the list of loc_ids to their original location file. It might be better to present a list of portfolio/account number combinations from the input location file which are not present so that they can do something about the problem

benhayes21 (approved):

benhayes21:
looks good

PR #746, opened 2021-02-10
Keys_Data handling

KamalCharles:
_Keys data handling_ -
Check the length of results & if there is no data present within the `json`  raises Oasis Exception


sambles (approved):
Looks good!

Note: not for this PR but might be good to add better handling of exceptions from supplier code

PR #749, opened 2021-02-11
Fix/group id internal oasis fields

sambles:
* set the cli flag `--group-id-cols` to take priority over the model_settings.json
* Added the following internal oasislmf columns as valid options for `group_id_col`
```
valid_oasis_group_cols = [
       'item_id',
       'peril_id',
       'coverage_id',
       'coverage_type_id',
]
```


sambles:
Fix needed for model_settings.json value

benhayes21 (changes requested):
looks like it's working form the command line but not from the model_settings file

benhayes21 (approved):

PR #756, opened 2021-02-16
Remove dependence on ReinsNumber order when assigning layer ID

hchagani-oasislmf:
Remove dependence on `ReinsNumber` order when assigning `layer_id`. `ReinsNumber`s must be unique in the `ri_info.csv` file but can be in any order.


johcarter (approved):
all good

PR #758, opened 2021-02-23
add validation to special conditions

sstruzik:
Add logic to make sure special condition structure from the location and account files are supported by Oasis


johcarter (approved):
behaves as expected

PR #760, opened 2021-02-23
Feature/consolidate fm tests update

johcarter:
validation/ substantial changes
tests/fm/ updated test_fm.py and added test_fmpy.py test
everything else should be in line with develop


sambles (approved):
Looks good

PR #764, opened 2021-02-25
set fm default to fmpy

sstruzik:


sambles (approved):
Tested disable/enable on CLI and in oasislmf.json, All works as expected 
` oasislmf model run --fmpy false` 

```
{                                                                                                                                            
    "fmpy": false,
    ...
}
```

Issue #771, opened 2021-02-26
genbash, fmpy calls --create-financial-structure-files when running GUL only 


sambles:
```
...
RUNNING: oasislmf.preparation.summaries.write_df_to_file
RUNNING: oasislmf.execution.bin.csv_to_bin
RUNNING: oasislmf.execution.bin.prepare_run_inputs
RUNNING: oasislmf.execution.runner.run
Traceback (most recent call last):
  File "/home/sam/repos/venv/py3/bin/fmpy", line 8, in <module>
    sys.exit(main())
  File "/home/sam/repos/venv/py3/lib/python3.8/site-packages/oasislmf/pytools/fmpy.py", line 33, in main
    manager.run(**kwargs)
  File "/home/sam/repos/venv/py3/lib/python3.8/site-packages/oasislmf/pytools/fm/manager.py", line 15, in run
    create_financial_structure(kwargs['allocation_rule'], kwargs['static_path'])
  File "/home/sam/repos/venv/py3/lib/python3.8/site-packages/oasislmf/pytools/fm/financial_structure.py", line 598, in create_financial_structure
    fm_programme, fm_policytc, fm_profile, stepped, fm_xref, items, coverages = load_static(static_path)
  File "/home/sam/repos/venv/py3/lib/python3.8/site-packages/oasislmf/pytools/fm/financial_structure.py", line 84, in load_static
    programme = load_file('fm_programme', fm_programme_dtype)
  File "/home/sam/repos/venv/py3/lib/python3.8/site-packages/oasislmf/pytools/fm/financial_structure.py", line 81, in load_file
    with open(os.path.join(static_path, name + '.csv')) as file_in:
FileNotFoundError: [Errno 2] No such file or directory: 'input/fm_programme.csv'

KTOOLS_STDERR:


STDOUT:
Ktools Run Error - exitcode=1
```

https://github.com/OasisLMF/OasisLMF/blob/master/oasislmf/execution/bash.py#L1048

PR #772, opened 2021-02-26
Fix/771 fix genbash

sambles:
Hot fix needed for GUL only runs using fmpy #771

### DIFF
--- a/oasislmf/execution/bash.py
+++ b/oasislmf/execution/bash.py
@@ -308,7 +308,7 @@ def do_kats(
 
             if summary.get('eltcalc'):
                 anykats = True
-                
+
                 cmd = 'kat -s' if sort_by_event else 'kat'
                 for process_id in range(1, max_process_id + 1):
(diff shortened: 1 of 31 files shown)


oasislmf/execution/bash.py
```diff
@@ -1046,11 +1046,13 @@ def genbash(
     print_command(filename, '')
 
     if fmpy:
-        print_command(
-            filename, f'{get_fmcmd(fmpy)} -a{il_alloc_rule} --create-financial-structure-files'
-        )
-        for i in range(1, num_reinsurance_iterations + 1):
-            print_command(filename, f'{get_fmcmd(fmpy)} -a{ri_alloc_rule} --create-financial-structure-files -p RI_{i}')
+        if il_output:
```
sstruzik, line 1049: you will need this if:
il_output or ri_output is true because if you have ri and not il you still need to create the financial file to compute il.

sstruzik (approved):

Issue #775, opened 2021-03-01
Review and update requirements.txt files


sambles:
* ideally automate on publish

PR #778, opened 2021-03-02
Update pinned requitments file

sambles:
Linked to issue #775 - updare `requirments.txt`


requirements.txt
```diff
@@ -4,102 +4,248 @@
...
+    #   sphinx
+    #   tox
+pandas==1.1.0
+    # via -r ./requirements-package.in
+parameterized==0.7.1
+    # via -r requirements.in
+parso==0.5.2
+    # via jedi
+pathlib2==2.3.5
+    # via -r ./requirements-package.in
```
DanielFEvans, line 128: If you're in the modd to remove some more Python 2 stuff, you should be able to just remove this, and replace `from pathlib2 import...` to `from pathlib import...`

sambles: Thats a good shout, I've give it a sweep, thanks!

PR #799, opened 2021-03-23
fix multilayer stream writer

sstruzik:
issue arising from the stream writer when there is multiple layer.
the proper value for i_layer was not properly tracked which cause problem with item written twice or omitted.


sambles (approved):
Looks good
> Note to self, might need a patch release to 1.15.5

PR #810, opened 2021-04-19
Fix/special condition multi policy issue [WIP]

sstruzik:


sstruzik:
fix to that is much more complex that anticipated, need more though about how we express it.
https://github.com/OasisLMF/OpenDataStandards/issues/30

johcarter:
i suggest we hold off until OasisLMF/OpenDataStandards#30 is answered. i am preparing another post to this issue which will provide more examples

PR #811, opened 2021-04-19
reset loss to 0 after event

sstruzik:
In non step use-cases we overwrite the loss so there is no issue but in the steeped policy we add loss to the previous step so we need to make sure loss are null when the step policies start to be applied.


johcarter (approved):
example test case 807 passes, with losses now identical to fmcalc
all validation tests pass

PR #821, opened 2021-05-04
Add missing items to data settings

fl-ndaq:
* docs_version and test_files_version: as a way to track metadata for model specific documentation and test files versions
* keys_data_version: some model have their keys_data included in model_data, in which case item could be empty
* additional_assets.deploy: needed to target specific deployments (default value for all is "always")


sambles (approved):
Looks good

sambles:
Note: PR needed to point to `develop` branch fixing this manually

Issue #824, opened 2021-05-06
Inputs directory preparation issue when ORD is enabled


johanxnilsson:
I tried to enable ORD outputs but it made piwind fail with the following error:

```
KTOOLS_STDERR:
FATAL: Error reading file input/occurrence.bin
```

It seems to happen when any psept_* or ept_* setting is enabled. My analysis settings:

```
{
  "analysis_tag": "144",
  "full_correlation": false,
  "gul_output": true,
"gul_summaries": [
    {
        "id": 1,
        "ord_output": {
            "ept_full_uncertainty_aep": true,
            "ept_full_uncertainty_oep": true,
            "ept_mean_sample_aep": true,
            "ept_mean_sample_oep": true,
            "ept_per_sample_mean_aep": true,
            "ept_per_sample_mean_oep": true,
            "psept_aep": true,
            "psept_oep": true,
            "return_period_file": true
        }
    }
],
  "gul_threshold": 0,
  "il_output": false,
  "model_settings": {
    "event_occurrence_id": "lt",
    "event_set": "p"
  },
  "model_version_id": "PiWind",
  "module_supplier_id": "OasisLMF",
  "number_of_samples": 1,
  "ri_output": false,
  "source_tag": "ord test 2"
}
```

Accordig to Ben it looks like a bug in the input directory preparation.

PR #825, opened 2021-05-10
Fix for #824 - check ordleccalc & leccalc when creating occurrence.bin

sambles:
### Fixed issue creating occurrence.bin when using ord options in analysis settings. 
Using only ord output options skipped occurrence file creation because the binary preparation code was keyed to check for 
`"lec_output": true,` which is not needed for ord settings. The following JSON now correctly prepares an `occurrence.bin` file. See issue #824 for details   
```
"gul_summaries": [
    {
        "id": 1,
        "ord_output": {
            "ept_full_uncertainty_aep": true,
            "return_period_file": true
        }
    }
],
```


hchagani-oasislmf (approved):
Looks good.

PR #836, opened 2021-05-21
flexible key lookup

sstruzik:
    Built-in Lookup class that implement the OasisLookupInterface
    provide a data driven lookup capability that will be both flexible (sub-classing capability) and efficient.
    lookup function provided include:
        - Location preparation (set default min and max value)
        - 2 geospacial option (direct computation or using rtree)
        - merge to map column keys to an id
        - split_loc_perils_covered to properly use data from LocPerilsCovered
        - simple_pivot allow to pivot columns of the locations dataframe into multiple rows
    example of config provided in Piwind lookup_new_key_server.json

    it provide several generic function factory that can be define in the config under the "step_definition" key (ex:)
    "step_definition": {
        "split_loc_perils_covered":{
            "type": "split_loc_perils_covered" ,
            "columns": ["locperilscovered"],
            "parameters": {
                "model_perils_covered": ["WTC", "WSS"]
            }
        },
        "vulnerability": {
            "type": "merge",
            "columns": ["peril_id", "coverage_type", "occupancycode"],
            "parameters": {"file_path": "%%KEYS_DATA_PATH%%/vulnerability_dict.csv",
                           "id_columns": ["vulnerability_id"]
                          }
        }
    }
    mapper key: is called the step_name,
        it will be added the the lookup object method once the function has been built
        it can take any value but make sure it doesn't collide with already existing method
    type: define the function factory to call.
        in the class for type <fct_type> the function factory called will be build_<fct_type>
        ex: "type": "merge" => build_merge
    columns: are the column required to be able to apply the step.
        those are quite important as any column (except 'loc_id')
        from the original Locations Dataframe that is not in any step will be drop to reduce memory consumption
    parameters: the parameter passed the the function factory.

    Once all the functions have been defined, the order in which they must be applied is defined in the config
    under the "strategy" key (ex:)
        "strategy": ["split_loc_perils_covered", "vulnerability"]

    It is totally possible to subclass Lookup in order to create your custom step or function factory
    for custom step:
        add your function definition to the "mapper"with no parameters
    "my_custom_step": {
            "type": "custom_type" ,
            "columns": [...],
    }
    simply add it to your "strategy": ["split_loc_perils_covered", "vulnerability", "my_custom_step"]
    and code the function in your subclass
    class MyLookup(Lookup):
        @staticmethod
        def my_custom_step(locations):
            <do something on locations>
            return modified_locations

    for function factory:
    add your function definition to the "step_definition" with the required parameters
    "my_custom_step": {
            "type": "custom_type" ,
            "columns": [...],
            "parameters": {
                "param1": "value1"
            }
    }
    add your step to "strategy": ["split_loc_perils_covered", "vulnerability", "my_custom_step"]
    and code the function factory in your subclass
    class MyLookup(Lookup):
        def build_custom_type(self, param1):
            def fct(locations):
                <do something on locations that depend on param1>
                return modified_locations

            return fct


sambles (commented):
New builtin key server code looks good, one comment on the package dependencies in setup.py. 

I think we should removed the hard requirements for `rtree`, `shapely` in **requirements-package.in**
https://github.com/OasisLMF/OasisLMF/blob/feature/key_server_revamp/requirements-package.in#L13-L15

And add a new file to list optional dependencies, maybe by added a file called `optional-package.in` which list the extra deps needed like 
```
shapely
geopandas
sklearn
```
and read that into a section for setup.py called `extras_require={ .. }`
https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies

sambles:
Going to open the above comment a separate issue

sambles (approved):
Looks good, adding to `1.17.0rc1`

PR #845, opened 2021-05-27
Set summarycalc option to output summary index files in genbash

hchagani-oasislmf:
### Set summarycalc option to output summary index files in genbash
The `ktools` binary `summarycalc` can now produce index files, which leads to a reduction in memory use by the downstream output components `leccalc` and `ordleccalc`. To take advantage of this, the option to output these summary index files is set as the default in genbash.


sambles (approved):
Looks good

PR #868, opened 2021-07-05
Introduce Vectorisation in generation of Reinsurance FM Files

hchagani-oasislmf:
### Introduce Vectorisation in generation of Reinsurance FM Files
The introduction of vectorisation, and resultant dropping of the tree structure, should lead to an improvement in performance in the generation of reinsurance FM files.


benhayes21 (approved):
Tested with 100 Fac contracts and time goes down from 3m30s to 7s for input generation with 100 Fac Contracts

PR #879, opened 2021-07-25
Refactor genbash & loss generation for Arch2020

sambles:
### Refactor genbash and LossGeneration to chunk events separately 

CLI Example: 
```
NUM_CHUNKS=10
oasislmf model generate-oasis-files --oasis-files-dir  ~/ram/run/input

# run each chunk of events 
for (( i=1; i<=$NUM_CHUNKS; i++ )); do
oasislmf model generate-losses-chunk --max-process-id $NUM_CHUNKS --process-number $i --model-run-dir ~/ram/run --oasis-files-dir  ~/ram/run/input
    echo $i
done

# run output reports 
oasislmf model generate-losses-output --model-run-dir ~/ram/run/ --oasis-files-dir ~/ram/run/input/ 
```
This creates individual scripts to run each event chunk plus a single output report script
```
├── 1.run_analysis.sh
├── 2.run_analysis.sh
├── 3.run_analysis.sh
├── 4.run_analysis.sh
├── 5.run_analysis.sh
├── 6.run_analysis.sh
├── 7.run_analysis.sh
├── 8.run_analysis.sh
├── 9.run_analysis.sh
├── 10.run_analysis.sh
├── run_outputs.sh
  ..
```


hchagani-oasislmf (approved):
Looks good. Just one, minor comment.

oasislmf/execution/bash.py
```diff
@@ -1075,94 +1150,90 @@ def genbash(
...
+    bash_params['fifo_full_correlation_dir'] = os.path.join(bash_params['fifo_queue_dir'], 'full_correlation/')
 
     # Set default alloc/shuffle rules if missing
-    gul_alloc_rule = gul_alloc_rule if isinstance(gul_alloc_rule, int) else KTOOLS_ALLOC_GUL_DEFAULT
-    il_alloc_rule = il_alloc_rule if isinstance(il_alloc_rule, int) else KTOOLS_ALLOC_IL_DEFAULT
-    ri_alloc_rule = ri_alloc_rule if isinstance(ri_alloc_rule, int) else KTOOLS_ALLOC_RI_DEFAULT
-    num_gul_per_lb = num_gul_per_lb if isinstance(num_gul_per_lb, int) else KTOOL_N_GUL_PER_LB
-    num_fm_per_lb = num_fm_per_lb if isinstance(num_fm_per_lb, int) else KTOOL_N_FM_PER_LB
-    event_shuffle = event_shuffle if isinstance(event_shuffle, int) else EVE_DEFAULT_SHUFFLE
+    bash_params['gul_alloc_rule'] = gul_alloc_rule if isinstance(gul_alloc_rule, int) else KTOOLS_ALLOC_GUL_DEFAULT
```
maxwellflitton, line 1210: wouldn't this be better just having a default dictionary that we override if there is an instance for instance:

```python 
class BashParams(dict):

    DEFAULT = {
        "one": 1,
        . . . 
    }

    def __init__(self):
        super().__init__(self.DEFAULT)

   def add_int(self, input_value, key_value):
       if isinstance(input_value):
           self[key_value] = input_value
```
This will give us a reference point where we can see what are the defaults in one point, and adding the parameter is less code as there is a lot of checks here

maxwellflitton (approved):

oasislmf/execution/bash.py
sambles: I agree, thats a better approach.  

I'll merge this PR to get the release candidate rolling and open a new issue to update the `bash_params` using your comment

PR #887, opened 2021-08-11
squash merged conflicts from feature/873-automated-issues-tests

sambles:
#873 
#885 

### Past fm issues added to regression test
A group of tests called 'issues' has been added to the fm regression tests so that issues related to the financial module calculations will be regression tested if a set of oed files has been added to the set of test cases in validation/issues.

> This PR is squashed changes by @johcarter, @KamalCharles  from https://github.com/OasisLMF/OasisLMF/pull/886

Issue #885, opened 2021-08-11
set fmpy sort option flag to true for oasislmf exposure run and test fm


johcarter:
## Issue Description

fmpy-sort-output flag sorts the output_ids for losses in ascending order for fmpy to be compatible with fmcalc. The validation tests sets this flag to true by default, but it is not default true on the command line. For oasislmf exposure run, and test fm, which are commands used principally for fm testing and where performance is not critical, it would be convenient to have the flag set to true to keep in line with the validation test.

## Version / Environment information

 
`feature/873-automated-issues-tests`

johcarter:
thank you @sambles

Issue #873, opened 2021-07-07
Add 'issue' regression cases to fm testing 


sambles:
## Issue Description
Test stored in [validation/issues](https://github.com/OasisLMF/OasisLMF/tree/master/validation/issues) aren't currently check when running the package testing. 

- [x] The file formats are inconsistent and need aligning 
- [x] Combine into single test case 
- [x] Joh to review

sambles:
https://github.com/OasisLMF/OasisLMF/tree/master/validation/issues/828

johcarter:
hey @sambles please can you test this for me? tests/fm/ pytest now has an issues test
My branch is
``feature/873-automated-issues-tests``
pytest is not recognising the expected folder ``issues/expected`` hence the issues test is failing. Is this the same for you?
thanks!

sambles:
@johcarter 
I think the keys return was missing the 2nd peril_id, its tripping over at:  
```
FAIL: generated /home/sam/repos/core/oasisLMF/validation/runs/test-fm-p2-issues-20210713163233/keys.csv vs expected /ho
.. etc
```

I've added what the tests are expecting to this commit: https://github.com/OasisLMF/OasisLMF/commit/55d02161fc209551f091df6c3801fdf2167e2c34 

Do the changed files look consistent?

sambles:
With that change pytest does run ok for me
```
tests/fm$ pytest 
===========================================================================
platform linux -- Python 3.8.7, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/sam/repos/core/oasisLMF, configfile: setup.cfg
plugins: cov-2.12.1, hypothesis-6.14.2, ignore-flaky-2.0.0
collected 18 items                                                                                                                                                                                                                                                                         

test_fm.py .........                                                                                                                                                                                                                                                                
test_fmpy.py .........                  
```

johcarter:
all tests are failing for me now. its difficult to tell why, but something to do with the location of the validation folder.  
E                   oasislmf.utils.exceptions.OasisException: The path /home/joh/dev/OasisLMF_branch/tests/fm/validation (Test directory - should contain test directories containing OED files and expected results) must exist for Computation Step RunFmTest

johcarter:
all tests pass on a clean branch. @sambles please could you review the PR?

johcarter:
This is complete on branch `feature/873-automated-issues-tests`. one enhancement to make on the branch before merging #885

PR #888, opened 2021-08-11
Feature/873 issues testing update2

johcarter:
### Past fm issues added to regression test
A group of tests called 'issues' has been added to the fm regression tests so that issues related to the financial module calculations will be regression tested if a set of oed files has been added to the set of test cases in validation/issues.


johcarter:
#873 #887 #885

sambles (approved):
Looks good

PR #898, opened 2021-09-02
max ded ba based on underlimit

sstruzik:
**IMPORTANT: Please attach or create an issue after submitting a Pull Request.

### max deductible back allocation based on under limit
change the back allocation method when max deductible is activated.

In the former implementation the amount back allocated due to the activation of max deductible was proportional to the sub node computed loss if it was not null and to the sub node input loss otherwise.
this had multiple issue regarding the consistency of the result.
depending on the financial structure and loss input, back allocated loss could:
- be bigger than the limit
- increase loss of item that had no deductible
- be not back allocated if the sub-nodes input loss were all null

in the new implementation, all the back allocated loss above the computed aggregated loss is reallocated based on the under limit of the sub nodes.
As the under limit is the minimum between the deductible and the loss available under the limit, this ensure that:
- the node loss will never be over the limit.
- the loss is back allocated proportionally to what is relevant to max deductible (the more deductible a node have the more is back allocated)
- no loss is lost (if the max deductible is activated then under limit is not null)


johcarter (approved):
All tests pass, thanks @sstruzik

PR #933, opened 2021-10-29
Feature/830 calcrule 28 and limit

sstruzik:
### Step policy 28 improvment
adding new calcrule 37 equivalent to 28 + limit
create specific calcrule for conditional coverage loss (was 28 with payout_start= 0)


johcarter:
in rule 38, we need to scale up the step loss input by scale 1 separately. this is the reason for the 3rd and 4th line in cpp below, which will return a loss even if condloss = 0.
This part of the loss is the scale1 uplift on the main coverage, which cannot be applied prior to calculating the extra expenses loss (condloss) so it must be applied after. 

in cpp 
condloss = x.step_loss * scale2; //calculate conditional payout based on prior step payout (extra expenses)
if (condloss > limit2) condloss = limit2; //limit conditional payout
debrisloss = (x.step_loss + condloss) * scale1; // debris removal uplift
loss = condloss + debrisloss; // payout is conditional loss for extra expenses plus full debris removal uplift

johcarter:
@sstruzik I've updated ctrlpy/fm59_a0_2.csv with the correct results in ktest/ftest for testing rule 38

johcarter (approved):
tests pass

PR #945, opened 2021-11-11
fmpy: init step policy loss

sstruzik:
init loss_out for step policy


johcarter (approved):
looks good, results match fmcalc on use case

PR #964, opened 2021-12-14
event id list added

benhayes21:
### Event ID list in Analysis Settings
Allows a list of event ids to be passed in the analysis settings file, which will be used as a priority over the default or delected event set file.


sambles (approved):

sambles:
Note: need to update the platform schema once merged

PR #966, opened 2021-12-15
Footprint server

maxwellflitton:
### TCP server that serves footprint data between getmodel processes [Experimental]
This feature is a Python TCP server that serves footprint data between getmodel processes. Once the specific data is served to the getmodel process it is deleted in the server process in order to reduce the amount of memory required by stopping duplication

To enable this on command line add the flags `--modelpy` and ` --model-py-server` 

Or via the configuration file **oasislmf.json**
```
{
   "modelpy": True, 
   "model_py_server": True, 
   ...
}
```


maxwellflitton:
This feature is a Python TCP server that serves footprint data between getmodel processes. Once the specific data is served to the getmodel process it is deleted in the server process in order to reduce the amount of memory required by stopping duplication

sambles:
Note on iterating on this feature: 

https://github.com/OasisLMF/OasisLMF/blob/footprint-server/oasislmf/pytools/data_layer/footprint_layer.py#L25-L26
```
TCP_IP = '127.0.0.1'
TCP_PORT = 8080
```

Binding the server on a fixed port number might hit problems in the long run, generally it should run ok inside worker because i don't think that port is used in the docker container.

PR #973, opened 2022-01-11
manage -4 and pass through -5 sidx

sstruzik:
### Handle Max loss -4 and Chance of Loss -5
- create a new category of sidx that will be pass through and added to the output stream
- handle sidx -4 (Max Loss) like other computed sidx

linked issue in Ktool:
https://github.com/OasisLMF/ktools/issues/255


johcarter (approved):
Tested with PiWind. chance of loss (-4) values are consistent across input and output for the same items . MaxLoss (-5) out < MaxLoss in and consistent with ImpactedExposure (sidx-3) out and in.

PR #985, opened 2022-02-07
Add cdftocsvpy tool

### LINKED ISSUE #987: Add a Python implementation of the `cdftocsv` tool
## Issue Description
It would be useful to have inside the oasislmf package a Python implementation of the `cdftocsv` tool.

mtazzari:
This PR closes #987.

### Added `cdftocsvpy` tool
A Python implementation of the `cdftocsv` tool is now available: it is called `cdftocsvpy`.

Like `cdftocsv`, `cdftocsvpy` prints the cdf produced by `getmodel` to `stdout` in `csv` format.
It has the same input/output interface of `cdftocsv`. It can be used with a stream as an input, e.g.:
```bash
eve 1 1 | getmodel | cdftocsv | head
```
produces:
```bash
event_id,areaperil_id,vulnerability_id,bin_index,prob_to,bin_mean
1,1,9,1,0.268378,0.000000
1,1,9,2,0.318824,0.125000
1,1,9,3,0.665534,0.375000
1,1,9,4,0.818048,0.625000
1,1,9,5,0.933643,0.875000
1,1,9,6,1.000000,1.000000
1,1,10,1,0.048263,0.000000
1,1,10,2,0.548144,0.125000
1,1,10,3,0.597388,0.375000
```
By passing the `-s` optional argument, the header is not printed, e.g.:
```bash
eve 1 1 | getmodel | cdftocsv -s | head
```
produces:
```bash
1,1,9,1,0.268378,0.000000
1,1,9,2,0.318824,0.125000
1,1,9,3,0.665534,0.375000
1,1,9,4,0.818048,0.625000
1,1,9,5,0.933643,0.875000
1,1,9,6,1.000000,1.000000
1,1,10,1,0.048263,0.000000
1,1,10,2,0.548144,0.125000
1,1,10,3,0.597388,0.375000
```

### DIFF
--- a/oasislmf/pytools/cdftocsv.py
+++ b/oasislmf/pytools/cdftocsv.py
@@ -4,119 +4,93 @@ import sys
 import os
 from contextlib import ExitStack
 import argparse
-import logging
 import numpy as np
 
 from oasislmf.pytools.getmodel.manager import get_mean_damage_bins
 from oasislmf.pytools.gul.common import ProbMean, damagecdfrec
-
-logger = logging.getLogger(__name__)
 
 parser = argparse.ArgumentParser(
     usage='use "%(prog)s --help" for more information',
     formatter_class=argparse.RawTextHelpFormatter  # for multi-line help text
 )
 
-parser.add_argument('-i', '--file-in', help='input filename', action='store', type=str)
 parser.add_argument('-s', help='skip header (default: False).', default=False, action='store_true', dest='skip_header')
 parser.add_argument('--run-dir', help='path to the run directory', default='.')
-parser.add_argument('--logging-level',
-                    help='logging level (debug:10, info:20, warning:30, error:40, critical:50)', default=30, type=int)
 
 
-def run(run_dir, skip_header, file_in=None, file_out=None):
+def run(run_dir, skip_header):
     """Run cdftocsv command: convert the cdf output from getmodel into csv format.
-    Presently, the csv file is streamed to stdout.
+    The binary data is read from an input stream, and the csv file is streamed to stdout.
 
     Args:
         run_dir ([str]): Path to the run directory.
+
         skip_header ([bool]): If True, does not print the csv header.
-        file_in ([str], optional): If provided, it reads the cdf from that file. Defaults to None.
-        file_out ([str], optional): Placeholder for a future output file writer. Defaults to None.
 
     Raises:
-        NotImplementedError: If file_out is not None.
         ValueError: If the stream type is not 1.
 
     """
-    with ExitStack() as stack:
-        if file_in is None:
-            streams_in = sys.stdin.buffer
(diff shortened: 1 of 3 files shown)


oasislmf/pytools/cdftocsv.py
```diff
@@ -0,0 +1,127 @@
...
+        # maximum number of damage bins (individual items can have up to `total_bins` bins)
+        if damage_bins.shape[0] == 0:
+            total_bins = 1000
+        else:
+            total_bins = damage_bins.shape[0]
+
+        # determine stream type
+        stream_type_mv = memoryview(bytearray(4))
+        stream_type = np.ndarray(1, buffer=stream_type_mv, dtype='i4')
+        len_read = streams_in.readinto(stream_type_mv)
```
sstruzik, line 68: I think you should do dirrectly
len_read = np.frombuffer(stream_in.read(4), dtype='i4')

oasislmf/pytools/cdftocsv.py
```diff
@@ -0,0 +1,127 @@
...
+        rec_mv = memoryview(bytearray(total_bins * ProbMean.size))
+        rec = np.ndarray(total_bins, buffer=rec_mv, dtype=ProbMean.dtype)
+
+        # start reading the stream
+        # each record from getmodel is expected to contain:
+        # 1 damagecdfrec obj, 1 int (Nbins), a number `Nbins` of ProbMean objects
+        while True:
+            len_read = streams_in.readinto1(damagecdf_mv)
+            len_read = streams_in.readinto1(Nbins_mv)
+            len_read = streams_in.readinto1(rec_mv[:Nbins[0] * ProbMean.size])
```
sstruzik, line 90: You can't really use readinto1 like this. 
With readinto1 you cannot expect to have values read that is more than 1 byte.
so if for some reason you read less than the size of your mv you will get false results

oasislmf/pytools/cdftocsv.py
mtazzari: Good idea, I've implemented the suggestion. Actually if I understand correctly `np.frombuffer` returns the array of the read values, not `len_read` so I've updated it with:
```py
stream_type = np.frombuffer(streams_in.read(4), dtype='i4')
```

oasislmf/pytools/cdftocsv.py
mtazzari: Thanks! I agree: I got mixed up and initially I was using `readinto` which ensure to fill the mv. Then I changed it to `readinto1` following our discussion but `readinto1` was more appropriate here.

I'll now update it with `readinto` so that this is safe, and hopefully can be merged into develop.
I'll then rebase the `feature/gulpy` branch on top of this  so I can work at a version that can be re-used for gulpy.

sstruzik (approved):

PR #988, opened 2022-02-08
Fix/negative sidx issue

sstruzik:
### fix size of intermediary dense array 
with the introduction of sidx -5, intermediary dense array  was too small and overlap with sidx (num_sample - 1).
increase the size by 2 to fix the issue


sambles (approved):
PiWind results look good, thanks for the fix!

PR #990, opened 2022-02-16
Feature/986 ods tools dyptes

sambles:
### Replace Dtype files with ods-tools package 
* Added ods_tools==2.1.2 and replaced dtype files with calls to fetch dtype from OED spec
* Added a workaround for https://github.com/pandas-dev/pandas/issues/30552


sstruzik (approved):
well done.

PR #991, opened 2022-02-21
Refactor group id seed

maxwellflitton:
### hashing group IDs
adding a hashing function that will replace the ```group_id``` with a hashed ID. This can be done by turning the ```hashed_group_id``` flag to ```true```


oasislmf/pytools/data_layer/conversions/items.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    Generates a hashed version of the group_id.
+
+    Args:
+        group_id: (int) the group ID to be hashed
+
+    Returns: (int) the hashed group ID
+    """
+    id_map = IdMap()
+    np.random.seed(group_id)
```
mtazzari, line 91: Note that `np.random.seed` is now a legacy method and numpy docs advise against re-seeding a generator. 
The new way of setting the seed is by creating random number generators objects, e.g. `np.random.default_rng`.
https://towardsdatascience.com/stop-using-numpy-random-seed-581a9972805f

PR #1024, opened 2022-05-03
Feature/back allocation

sstruzik:
### support OED policy coverage terms

- allow fmpy to handle multi-tree structure via using negative agg_id to indicate a reference to item level
- back allocate loss at each level during fm (including deductible overlimit and underlimit)
- add new level PolCoverage


johcarter:
@sstruzik although the output looks correct, the latest change has introduced some duplicate entries in fm_programme for the insurance test case (fmcalc)(https://github.com/OasisLMF/OasisLMF/pull/1024/commits/3041d4d2663b2923b2bc43af85d66f90d922fdff)

PR #1027, opened 2022-05-06
Fix/gulpy reduce memory

mtazzari:


mtazzari:
Not a good solution performance-wise.
We solved this in PR #1033

PR #1030, opened 2022-05-11
Drop kat sort flag in genbash

hchagani-oasislmf:
### Drop kat sort flag
From `ktools v3.8.1`, `kat` will attempt to detect table type and sort by event ID in the case of Exceedance Loss Tables (ELTs) and by period ID, and then by event ID, in the case of Period Loss Tables (PLTs), prior to concatenation. As this is now the default option, the sort flag `-s` has been dropped. Sorting correctly requires `eve` to employ the deterministic method to shuffle event IDs. Therefore, if other shuffling methods have been used by `eve`, the flag `-u`, which produces unsorted output, is employed.


johcarter (approved):
All works as expected

sambles (approved):

PR #1033, opened 2022-05-12
Improve gulpy memory usage and performance

### LINKED ISSUE #1041: Improve memory usage of gulpy
## Issue Description
gulpy uses too much memory when used on a real model. Memory keeps piling up as the stream from getmodel is read.
gulpy reaches above 256 GB RAM usage for 1/8 of the model computation and gets killed even on an m4.16xlarge AWS instance. 

Run `oasis model run --gulpy` on a real model.

OS: Linux
Oasis version: 1.26.0rc1, 1.26.1rc1

mtazzari:
This PR is to Fix #1041 

### Releasing production-ready `gulpy`
This PR significantly improves memory usage and performance of gulpy, which is now ready for production.


oasislmf/pytools/gul/io.py
```diff
@@ -2,283 +2,351 @@
...
-    # maximum number of damage bins (individual items can have up to `total_bins` bins)
-    if damage_bins.shape[0] == 0:
-        max_Nbins = 1000
-    else:
-        max_Nbins = damage_bins.shape[0]
-
     # maximum number of entries is buff_size divided by the minimum entry size
     # (corresponding to a 1-bin only cdf)
-    min_size_cdf_entry = damagecdfrec_stream.size + ProbMean.size + 4
+    min_size_cdf_entry = damagecdfrec_stream.size + 4 + ProbMean.size
```
sstruzik, line 71: I think this 4 is not correct if area_peril_id is an int64

oasislmf/pytools/gul/io.py
```diff
@@ -2,283 +2,351 @@
...
 
-        else:
-            # this is not the last cycle
-            # move the un-read data to the beginning of the memoryview
-            mv[:valid_buf - cursor_buf] = mv[cursor_buf:valid_buf]
 
-            # update the length of the valid data
-            valid_buf -= cursor_buf
+@njit(cache=True, fastmath=True)
+def insert_item_data_val(items_data, item_id, damagecdf_i, rng_index):
```
sstruzik, line 153: The ordering here is not useful, we can insert directly in the cur_item position then increment cur_item

oasislmf/pytools/gul/io.py
mtazzari: The 4 is for the int32 storing `Nbins`.
The areaperil_int is already accounted for in `damagecdfrec_stream`
```py
damagecdfrec_stream = nb.from_dtype(np.dtype([('event_id', np.int32),
                                              ('areaperil_id', areaperil_int),
                                              ('vulnerability_id', np.int32)
                                              ]))
```

So, I think this is correct. Please resolve if you agree.

oasislmf/pytools/gul/manager.py
```diff
@@ -235,190 +205,126 @@ def run(run_dir, ignore_file_type, sample_size, loss_threshold, alloc_rule, debu
...
+            return cursor, cursor * int32_mv.itemsize, last_processed_coverage_ids_idx
 
-        # sort item_ids (need to convert to np.array beforehand)
-        item_ids = np.empty(Nitem_ids, dtype=ITEM_ID_TYPE)
-        for j in range(Nitem_ids):
-            item_ids[j] = item_ids_by_coverage_id[coverage_id][j]
+        items = items_data[coverage['start_items']: coverage['start_items'] + coverage['cur_items']]
 
-        item_ids_argsorted = np.argsort(item_ids)
-        item_ids_sorted = item_ids[item_ids_argsorted]
```
mtazzari, line 378: I've just found that here actually I was sorting the items within each coverage by `item_id`.

oasislmf/pytools/gul/io.py
mtazzari: Ok, I've removed the insert-sort.

sstruzik (approved):

PR #1034, opened 2022-05-16
Distributed Platform Fixes

sambles:
#### Fixed Parallel analysis chunks for new platform 
The Updated platform in https://github.com/OasisLMF/OasisPlatform/tree/platform-2.0 handles ktools parallel workloads using concurrent bash scripts. These were causing interference with each other, where error checking and temporary file cleanup commands would kill scripts from other analysis chunks. 

1. Disabled the script `check_complete` calls when running in chunked mode - this is used to detect dropped processes (OOM errors) and would incorrectly trigger when analysis chunk run in parallel. 

2. When running as an analysis chunks, ktools logs are  now placed in `<run_dir>/log/<chunk_id>`  or `<run_dir>/log/out` for report generation, this separates the logging on a per chunk basis and stops output in **log/stderror.err** killing all currently running chunks. 

3. Replaced cleanup calls for  **fifo** dirs, switched from `rm -R -f fifo/*` to a more selective `find fifo/ \( -name '*P<chunk_id>[^0-9]*' -o -name '*P<chunk_id>' \) -exec rm -R -f {} +`

4. Fixed summaryxref generation, this was incorrectly called before processing each chunk causing failures when reading the GUL xref file. 

5. Fixed Chunk storage, the work dir for each batch of events has been moved to `{chunk_id}.work` to prevent interference, so the outputs for script `1.run_analysis.sh` are accumulated and stored in `1.work`  
```
-rw-r--r--  1 root root 3436 May 20 07:33 1.run_analysis.sh
drwxr-xr-x  5 root root 4096 May 20 07:33 1.work
-rw-r--r--  1 root root 3432 May 20 07:33 2.run_analysis.sh
drwxr-xr-x  5 root root 4096 May 20 07:33 2.work
-rw-r--r--  1 root root 3432 May 20 07:33 3.run_analysis.sh
drwxr-xr-x  5 root root 4096 May 20 07:33 3.work
```
6. Moved the fmpy financial structures generation from bash to to the `prepare-losses-generation-directory` sub-task. This prevents IL and RI loss generation from crashing with: 
```
KTOOLS_STDERR:
  File "/root/.local/bin/fmpy", line 8, in <module>
    sys.exit(main())
  File "/root/.local/lib/python3.8/site-packages/oasislmf/pytools/fmpy.py", line 34, in main
    manager.run(**kwargs)
  File "/root/.local/lib/python3.8/site-packages/oasislmf/pytools/fm/manager.py", line 24, in run
    return run_synchronous(**kwargs)
  File "/root/.local/lib/python3.8/site-packages/oasislmf/pytools/fm/manager.py", line 46, in run_synchronous
    run_synchronous_sparse(max_sidx_val, allocation_rule, streams_in=streams_in, files_out = files_out, net_loss=net_loss, **kwargs)
  File "/root/.local/lib/python3.8/site-packages/oasislmf/pytools/fm/manager.py", line 95, in run_synchronous_sparse
    compute_info, nodes_array, node_parents_array, node_profiles_array, output_array, fm_profile = load_financial_structure(
  File "/root/.local/lib/python3.8/site-packages/oasislmf/pytools/fm/financial_structure.py", line 680, in load_financial_structure
    nodes_array = np.load(os.path.join(static_path, f'nodes_array_{allocation_rule}.npy'), mmap_mode='r')
  File "/root/.local/lib/python3.8/site-packages/numpy/lib/npyio.py", line 445, in load
    raise ValueError("Cannot load file containing pickled data "
ValueError: Cannot load file containing pickled data when allow_pickle=False
FATAL: summarycalc: Read error on stream
``` 

#### Fixed loading the fm_aggregation_profile
The FM file generation in the new platform fails when using the fm_aggregation_profile. This was due to the fm_aggregation_profile loading the level keys as type `str`, added a fix to check and switch these to `int`
```
  File "/root/.local/lib/python3.8/site-packages/oasislmf/computation/generate/files.py", line 274, in run
    il_inputs_df = get_il_input_items(
  File "/root/.local/lib/python3.8/site-packages/oasislmf/utils/log.py", line 111, in wrapper
    result = func(*args, **kwargs)
  File "/root/.local/lib/python3.8/site-packages/oasislmf/preparation/il_inputs.py", line 1178, in get_il_input_items
    prev_agg_key = [v['field'].lower() for v in fm_aggregation_profile[level_id]['FMAggKey'].values()]
KeyError: 1
```


sambles:
TODO: The oasislmf file generation won't create IL / RI files because the file paths are not set on command line / config 
https://github.com/OasisLMF/OasisLMF/blob/develop/oasislmf/computation/generate/losses.py#L210-L213
However they may exist in the files dir. 

Fix should go here: 
https://github.com/OasisLMF/OasisPlatform/blob/platform-2.0/src/model_execution_worker/distributed_tasks.py#L516-L524

sambles:
### To do: 
-  [ ] concurrent calls to `fmpy -a2 --create-financial-structure-files` causes fmpy.py to crash, try moving these from the script to the directory prep step
- [ ] Analysis run chunks can exit leaving blank work bin files, these then crashes the output_report task with `FATAL:: Not a summarycalc stream invalid stream type: 0`, add a check to the script exit

PR #1039, opened 2022-05-20
Fix/282 events not in occurrence file (Backport)

sambles:
###  Ignore events that do not exist in occurrence file when creating summary indexes in aalcalc

From: https://github.com/OasisLMF/ktools/pull/284

It is no longer a requirement for all event IDs to be present in the occurrence file in order to create the summary index files in aalcalc. Should an event ID be encountered that does not exist in the occurrence file, only its offset is recorded.


hchagani-oasislmf (approved):
Looks good.

PR #1044, opened 2022-05-31
Correlation values

maxwellflitton:


oasislmf/pytools/data_layer/oasis_files/model_settings.py
```diff
@@ -0,0 +1,112 @@
...
+        if not os.path.exists(path=self.file_directory_stash):
+            raise FileNotFoundError(
+                "stash file not found please implement the defile_path function before trying to read "
+                "model settings data"
+            )
+        with open(self.file_directory_stash) as file:
+            return file.read()
+
+    @property
+    def file_directory_stash(self) -> str:
```
sstruzik, line 93: use os.path.join to support Windows os

PR #1062, opened 2022-06-28
add unsupported peril in the keys-errors file

sstruzik:
### add unsupported peril in the keys-errors file
In the keys server, Keys with Peril Id not present in the model peril covered were simply discarded. To make it clearer as to why the key was removed, they will now appear in the keys-errors.csv file with the status "noreturn" and message "unsuported peril_id"


sstruzik:
[keys-errors.csv](https://github.com/OasisLMF/OasisLMF/files/9002507/keys-errors.csv)
output example from the original issue

sambles (approved):
Looks good

PR #1074, opened 2022-07-08
Added lower-case-cols and raise-error flags

sambles:
### Added more flags to compare-parquet script
* `--lower-case-cols` lower case column names of both dataFrames before running compare 
* `--raise-error` default is now to catch and print the exception message, add flag to also raise the exception (like before PR) 

Found these extra options handy, merge if useful otherwise flag branch for deletion


validation/utils/compare_parquet_csv.py
```diff
@@ -13,6 +13,8 @@
 parser.add_argument('-c', '--csv-path', help='path to csv file', type=str, required=True)
 parser.add_argument('-s', '--sort', help='sort by string, int and categorical columns before check', action='store_true')
 parser.add_argument('--check-exact', help='Whether to compare number exactly.', action='store_true')
+parser.add_argument('--lower-case-cols', help='Lower cased column names before compare', action='store_true')
```
sstruzik, line 16: I suggest calling this param --case-sensitive (default false)

validation/utils/compare_parquet_csv.py
```diff
@@ -13,6 +13,8 @@
...
+parser.add_argument('--rasie-error', help='Lower cased column names before compare', action='store_true')
```
sstruzik, line 17: there is a typo in rasie.
For help something like:
raise error if data are different

validation/utils/compare_parquet_csv.py
```diff
@@ -23,22 +25,31 @@ def is_sorting_type(dtype):
...
     if sort:
         sort_column = [column for column, dtype in dtypes.items() if is_sorting_type(dtype)]
         parquet_df = parquet_df.sort_values(by=sort_column)
         csv_df = csv_df.sort_values(by=sort_column)
 
-    pd.testing.assert_frame_equal(parquet_df, csv_df, **kwargs)
+    try:
+        pd.testing.assert_frame_equal(parquet_df, csv_df, **kwargs)
+    except AssertionError as e:
+        print(e)
```
sstruzik, line 45: Maybe
if raise_error:
  raise
else:
   print(e)
to not have the error printed twice

sambles:
Thanks for catching, updated PR

sstruzik (approved):

PR #1077, opened 2022-07-11
Feature/peril filter

sstruzik:
### Peril specific filter for modelpy and gulpy

add a filter option for modelpy and gulpy
- modelpy --peril-filter WTC < eve.bin > /dev/null 
- gulpy -S10 -L0 -a1 --peril-filter WTC > /dev/null < cdf.bin

allow peril filter to be specified via the MDK 
- in the mdk => --peril-filter WSS WTC
- in oasislmf.json =>  "peril_filter": ["WSS", "WTC"],

if filter is specified in the MDK only modelpy is filtering as is is sufficient. Option in gulpy can be use for custom model.


oasislmf/pytools/gul/common.py
```diff
@@ -68,3 +68,10 @@
                                              ('loss', oasis_float),
                                              ]))
 gulSampleslevelRec_size = gulSampleslevelRec.size
+
+
+Keys = {'LocID': np.int32,
```
mtazzari, line 73: `Keys` is already defined in  `pytools/getmodel/common.py` and it should not be redefined here.
We should import `Keys` from that module, or viceversa, but not defining it twice.

oasislmf/pytools/gul/io.py
```diff
@@ -31,7 +31,16 @@ def gen_structs():
     return group_id_rng_index, rec_idx_ptr
 
 
-def read_getmodel_stream(stream_in, item_map, coverages, compute, seeds, buff_size=GETMODEL_STREAM_BUFF_SIZE):
+@njit(cache=True)
+def gen_valid_area_peril(valid_area_peril_id):
```
mtazzari, line 35: Could you add docstring to this function?

oasislmf/pytools/gul/common.py
sambles: Good point about duplication, it seems like we should add a file like `oasislmf/pytools/common.py` for dictionaries/data common to all of the pytool components (rather than sharing them across sub-dirs)

mtazzari:
Also, I don't understand why we need two nested ifs:
```
        if valid_area_peril_dict is not None:
            if areaperil_id not in valid_area_peril_dict:
                cursor += 2 * oasis_float_to_int32_size * Nbins_to_read
                continue
```
It would have been more efficient to have `valid_area_peril_dict={}` by default instead of None, so that we could have done:
```
        if areaperil_id not in valid_area_peril_dict:
            cursor += 2 * oasis_float_to_int32_size * Nbins_to_read
            continue
```
Since these nested `if`s are inside `stream_to_data`, they will lead to a definite performance loss if valid_area_peril_dict is not None. Any tests were done on the performance of this case?

oasislmf/pytools/gul/common.py
mtazzari: Indeed, Sam, this is indeed what I've started doing here: https://github.com/OasisLMF/OasisLMF/pull/1118

PR #1000, opened 2022-02-25
Feature/gulpy

mtazzari:
This PR introduces the `gulpy` tool, Python implementation of the `gulcalc` tool.
This PR fixes #999 .

Features to be implemented/action items:
- [x] `get_gul`: test point-like and linear interpolation
- [x] `get_gul`: test quadratic interpolation
- [x] implement cli argument to print version (`-v` or `-V`)
- [x] implement back-allocation `-a1`
- [x] implement back-allocation `-a2`
- [x] test back-allocation `-a1` with fixed random values
- [x] test back-allocation `-a2` with fixed random values
- [x] implement debug (`-d` arg) output random numbers instead of gul
- [x] implement new random number generators
- [x] test new random number generators 
- [x] [WON'T DO] implement output pipe (`-i` arg)
- [x] [WON'T DO IN THIS PR] implement correlated stream (`-j` arg); may not be needed depending on how we implement the correlation factor in #910
- [x] [NOT NEEDED] incorporate new hashing algorithm from #991: not needed because the new hashing algo produces a new `group_id` and `gulpy` takes the `group_id` agnostically, so they are decoupled implementation.
- [x] optimize/speed up the code with numba wherever possible
  - [x] accelerate `gulpy` input (`read_getmodel_stream()`) with numba
  - [x] accelerate `gulpy` for alloc_rule=0
  - [x] accelerate `gulpy` for alloc_rule=1
  - [x] accelerate `gulpy` for alloc_rule=2
- [x] [DEFERRED TO FUTURE PR] test on large models
- [x] [WON'T FIX IN THIS PR] add option to MDK (see fmpy) -> created issue #1010  to address this after this PR is merged.
- [x] add/update/finalise docstrings of all functions

### Add `gulpy`, the Python version of gulcalc
This PR introduces `gulpy`, a Python version of gulcalc. At the time of merging this PR, gulpy is functionally equivalent to gulcalc, i.e. the following commands are equivalent:
```bash
# with gulcalc                ...with gulpy
gulcalc -a0 -S10 -i -         gulpy -a0 -S10
gulcalc -a1 -S20 -i -         gulpy -a1 -S20
gulcalc -a2 -S30 -i -         gulpy -a2 -S30
````
By default, `gulpy` uses the Latin Hypercube Sampler algorithm to draw random numbers for the positive sidx samples, which is shown to require less samples than the Mersenne Twister used by `gulcalc` when probing a given probability distribution function. 

[...]

PR #1078, opened 2022-07-12
Add gulpy_random_generator flag to `oasislmf` model runner

### LINKED ISSUE #1049: random number generator can be set to 0 at oasislmf command line
This is so models can be run using the Mersenne Twister random number generator --gulpy-random-generator 0, as the default gulpy generator is Latin Hypercube.

Low priority, as people can still run gulcalc (which uses Mersenne Twister) using gulpy=false flag

mtazzari:
Fix #1049 

### New `gulpy_random_generator` flag in the `oasislmf` model runner
This PR introduces the `gulpy_random_generator` flag in the `oasislmf` model runner. This will allow users to set the random number generator to be used in `gulpy`. By default `gulpy` uses the Latin Hypercube Sampling algorithm (see #1000). However, it also implements the Mersenne Twister random generator (namely, the generator that was used in `gulcalc`).

With the introduction of `gulpy`, the user can already set the random number generator with its `--random-generator` flag:
```bash
  --random-generator RANDOM_GENERATOR
                        random number generator
                        0: numpy default (MT19937), 1: Latin Hypercube. Default: 1.
```
Regarding the `oasislmf` model runner CLI, so far the following command
```bash
oasislmf model run --gulpy [...]
```
implicitly used the Latin Hypercube generator. 

This PR introduces the possibility for the user to specify which random number generator to be used in `gulpy` through the ` --gulpy-random-generator` flag. The following commands are equivalent and use the Latin Hypercube Sampling:
```
oasislmf model run --gulpy [...]
oasislmf model run --gulpy --gulpy-random-generator=1 [...]  
```
To run a model with the Mersenne Twister it is now possible with:
```
oasislmf model run --gulpy --gulpy-random-generator=0 [...]  
```


sambles (changes requested):
Its worth adding your command to the `_check_ktool_rules` function:
```
        rule_ranges = { 
            'ktools_alloc_rule_gul': KTOOLS_ALLOC_GUL_MAX,
            'ktools_alloc_rule_il': KTOOLS_ALLOC_FM_MAX,
            'ktools_alloc_rule_ri': KTOOLS_ALLOC_FM_MAX,
            'ktools_event_shuffle': EVE_STD_SHUFFLE,                                                                           
            'gulpy_random_generator': 1
        } 
```

That way it will flag to the user if the value isn't valid: 
```
$ oasislmf model run --gulpy-random-generator 42

Stating oasislmf command - RunModel
   .... 
Error: gulpy_random_generator=42 - Not within valid ranges [0..1]
```   
   
```

https://github.com/OasisLMF/OasisLMF/blob/6832b6a116f5b29707a5afb46056b352b473c973/oasislmf/computation/generate/losses.py#L111-L120

PR #1079, opened 2022-07-12
fix gulpy error for empty cdf

sstruzik:


sambles (approved):
Retested on azure, looks good 

[FIXED_chunk-3-output.log](https://github.com/OasisLMF/OasisLMF/files/9091618/FIXED_chunk-3-output.log)

PR #1082, opened 2022-07-13
Run unittests with github actions

sambles:
<!--- IMPORTANT: Please attach or create an issue submitting a Pull Request. -->

<!--start_release_notes-->
### Move package unit testing to GitHub actions 
oasislmf is testing vs multiple python versions with an option to pin a single dependent package: 

#### Example test matrix
```
    strategy:
      matrix:
        cfg:
        - { python-version: '3.9', pkg-version: ""}                                                                                                                                                                                                            
        - { python-version: '3.10', pkg-version: ""} 
        - { python-version: '3.10', pkg-version: 'numba==0.55.1' }
        - { python-version: '3.10', pkg-version: "pandas>=1.3.0"} 
```
* Removed `Sphinx` and doctest due to pip-compile error (can add back in later) 
* By default Unit-tests are skipped in Jenkins, but the piwind tests still trigger  

<!--end_release_notes-->


.github/workflows/oasislmf-unittest.yml
```diff
@@ -0,0 +1,50 @@
...
+          python -m pip install --upgrade pip
+          pip install flake8 pytest pip-tools tox
+
+      - name: Pip Compile
+        run: |
+          rm -f requirements.txt
+          if [ -z "${{ matrix.cfg.pkg-version }}" ]; then
+            pip-compile requirements.in -o requirements.txt
+          else
+            pip-compile --upgrade-package "${{ matrix.cfg.pkg-version }}" requirements.in  ${{ matrix.cfg.numba-version }} -o requirements.txt
```
mtazzari, line 46: This `matrix.cfg.numba-version` is not parsed because `numba-version` is not in the `cfg` dict (See indeed the CI output). Perhaps you meant `matrix.cfg.pkg-version`?
In any case, I wonder if we need that bit at all.

.github/workflows/oasislmf-unittest.yml
```diff
@@ -0,0 +1,50 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        cfg:
+        - { python-version: '3.7', pkg-version: ""}
+        - { python-version: '3.8', pkg-version: ""}
+        - { python-version: '3.9', pkg-version: ""}
+        - { python-version: '3.10', pkg-version: ""}
+        - { python-version: '3.10', pkg-version: 'numba==0.55.1' }
```
mtazzari, line 15: Now that this CI works, I'd add further tests to the matrix for better coverage
```suggestion
        - { python-version: '3.7', pkg-version: ""}
        - { python-version: '3.7', pkg-version: 'numba==0.55.1' }
        - { python-version: '3.8', pkg-version: ""}
        - { python-version: '3.8', pkg-version: 'numba==0.55.1' }
        - { python-version: '3.9', pkg-version: ""}
        - { python-version: '3.9', pkg-version: 'numba==0.55.1' }
        - { python-version: '3.10', pkg-version: ""}
        - { python-version: '3.10', pkg-version: 'numba==0.55.1' }
```

mtazzari (changes requested):
Very nice CI test matrix!
I posted minor comments and some suggestions.

.github/workflows/oasislmf-unittest.yml
sambles: Good catch it was called `numba-version` before it got generalized to any package

tox.ini
```diff
@@ -8,7 +8,8 @@ omit = tests/fm/test_fmpy.py
 
 [testenv]
 deps = -r{toxinidir}/requirements.txt
-commands = pytest -p no:flaky --ignore=fm_testing_tool --ignore=validation --cov-config=tox.ini --cov=oasislmf {posargs} --doctest-modules
+commands = pytest -p no:flaky --ignore=fm_testing_tool --ignore=validation --cov-config=tox.ini --cov=oasislmf {posargs}
+#$commands = pytest -p no:flaky --ignore=fm_testing_tool --ignore=validation --cov-config=tox.ini --cov=oasislmf {posargs} --doctest-modules
```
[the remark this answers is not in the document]
sambles: That test flag needs Sphinx to work which is causing a problem in pip-compile: 
https://github.com/OasisLMF/OasisLMF/runs/7320293263?check_suite_focus=true

Looks to be an open issue here
https://github.com/sphinx-doc/sphinx/issues/10241

.github/workflows/oasislmf-unittest.yml
sambles: I think lets leave this as is for the moment, it feels enough on a per commit basis. We could add a more complete set of checks on the master / main branch only

mtazzari (approved):

PR #1092, opened 2022-07-26
Add check for parquet output before running model

hchagani-oasislmf:
### Add check for parquet output before running model
Optional third party libraries are required for parquet output of `ktools` components (see https://github.com/OasisLMF/ktools/pull/283 for more details). A user can request parquet output by setting the `parquet_format` flag to `true` in the `analysis_settings.json` file. It is possible to compile `ktools` binaries without linking to these optional parquet libraries, as is the case with the Mac OS build. In this case, requesting parquet output will result in an error after all loss calculations have been performed.

A check to determine whether the `ktools` components have been linked with parquet libraries during compilation has been introduced before input generation, yielding an error message if parquet output has been requested but is not supported by the `ktools` build.

### DIFF
--- a/oasislmf/computation/generate/losses.py
+++ b/oasislmf/computation/generate/losses.py
@@ -215,6 +215,27 @@ class GenerateLossesDir(GenerateLossesBase):
 
     ]
 
+    def __check_for_parquet_output(self, analysis_settings, runtypes):
+        """
+        Private method to check whether ktools components were linked to third
+        party parquet libraries during compilation if user requests parquet
+        output.
+        """
+        for runtype in runtypes:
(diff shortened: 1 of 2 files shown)


sambles (changes requested):
Works well with the `oasislmf model run ..` command.  

Just a suggestion, its better to move the **__check_for_parquet_output()** call to the run function in https://github.com/OasisLMF/OasisLMF/blob/develop/oasislmf/computation/generate/losses.py#L188 

That way the check will also trigger when running `oasislmf model generate-losses .. ` Or when calling the loss generation directly from the OasisManager class

sambles (approved):
looks good, thanks for the update

PR #1094, opened 2022-08-04
Disable GroupID hashing for acceptance tests

sambles:
### Disable GroupID hashing for acceptance tests 
* Switched the default of `hashed_group_id` to **False** for the FM acceptance tests 

### (For reference) Update expected output steps (Oasislmf)
1. toggle the setting `self.update_expected` to True in the following files 

```
oasisLMF/tests/fm/test_fm.py
oasisLMF/tests/fm/test_fmpy.py
```
2. Run the tests with pytest (using the code from the PR branch to update)

```
$ pytest 
===================================================== test session starts =====================================================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/sam/repos/core/oasisLMF, configfile: setup.cfg
plugins: cov-3.0.0, ignore-flaky-2.0.0, hypothesis-6.30.1
collected 21 items                                                                                                            

test_fm.py .........                                                                                                    [ 42%]
test_fmpy.py ............                                                                                               [100%]
```
3. Commit the edited files and retest with the flag `self.update_expected` set back to False.


johcarter:
we don't need to test correlation in tests/fm as this test is purely about the gross losses given the ground up inputs. can we introduce a parameter which sets hashed group id to false please @sambles ?

johcarter:
@sambles I actually meant setting the parameter to false only in the test script test_fm.py and test_fmpy.py, and keeping it defaulting to true for exposure run. I think it would be confusing if exposure run produces a different set of inputs to model run. do you agree? Sorry for not being clear.

mtazzari (approved):
Good to go!

PR #1096, opened 2022-08-08
Hashing investigation

maxwellflitton:
This PR solves a bug for which the group_ids produced through hashing (with `hashed_group_id=True`) were different even if they should have been identical given the input parameters of the hashing function were identical.
This PR ensures that group_ids are are equal if the input parameters of the hashing function are identical.

### DIFF
--- a/oasislmf/preparation/gul_inputs.py
+++ b/oasislmf/preparation/gul_inputs.py
@@ -333,7 +333,10 @@ def get_gul_input_items(
 
     # this block gets fired if the hash_group_ids is True
     else:
-        gul_inputs_df["group_id"] = (pd.util.hash_pandas_object(gul_inputs_df[group_id_cols]).to_numpy() >> 33)
+        para_chain = gul_inputs_df.drop_duplicates(subset=group_id_cols).reset_index(drop=True)
+        para_chain["group_id"] = pd.util.hash_pandas_object(para_chain[group_id_cols])
+        para_chain = para_chain[group_id_cols + ["group_id"]]
+        gul_inputs_df = pd.merge(gul_inputs_df, para_chain, how='left', left_on=group_id_cols, right_on=group_id_cols)
 
     gul_inputs_df['group_id'] = gul_inputs_df['group_id'].astype('uint32')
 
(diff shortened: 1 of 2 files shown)


mtazzari (approved):
This is good to be merged in my opinion.

oasislmf/preparation/gul_inputs.py
```diff
@@ -279,7 +279,7 @@ def get_gul_input_items(
 
     # Concatenate chunks. Sort by index to preserve item_id order in generated outputs compared
     # to original code.
-    gul_inputs_df = pd.concat(gul_inputs_reformatted_chunks).sort_index().reset_index()
+    gul_inputs_df = pd.concat(gul_inputs_reformatted_chunks).sort_index().reset_index(drop=True, inplace=True)
```
mtazzari, line 282: Great, this will ensure indices are not repeated by construction.

oasislmf/preparation/gul_inputs.py
```diff
@@ -333,7 +333,7 @@ def get_gul_input_items(
 
     # this block gets fired if the hash_group_ids is True
     else:
-        gul_inputs_df["group_id"] = (pd.util.hash_pandas_object(gul_inputs_df[group_id_cols]).to_numpy() >> 33)
+        gul_inputs_df["group_id"] = pd.util.hash_pandas_object(gul_inputs_df[group_id_cols])
```
mtazzari, line 336: I agree that by introducing `drop=True` in `reset_index()` here we know that indices are not duplicate, but in any case we need to ensure we compute hashes only using the desired columns:
```suggestion
        gul_inputs_df["group_id"] = pd.util.hash_pandas_object(gul_inputs_df[group_id_cols], index=False)
```

mtazzari: I can't comment on the bitshift `>> 33`. I don't understand why we needed it in the first place, so please @sstruzik comment on this. In my view, we might not need to divide by 2**33 anymore, but please confirm.

mtazzari (changes requested):

mtazzari:
@maxwellflitton note that I've updated the branch with the latest develop commits to see if build doesn't break anymore, hopefully.

mtazzari:
Note: the tests don't pass yet because the fix for the tests is in feature/correlated_rng branch. So we need to merge this onto that branch and not directly onto develop.

oasislmf/preparation/gul_inputs.py
mtazzari: @sstruzik suggested that we keep the >> 33 because  "The bit shift is to have value in the int32 range"

Issue #1003, opened 2022-03-08
With alloc_rule=1 do not cap STD_DEV and CHANCE_OF_LOSS values


mtazzari:
## Issue Description
`splittiv` is the function in `ktools:gulcalc.cpp` that caps the losses to the tiv, proportionally to the losses themselves (i.e., alloc_rule=1). `splittiv` is called in `writemode1output` on all sidx numbers, including the negative ones.
You can see below that `splittiv` is applied to `gilv[i]`, which is itself a list of `itemID-loss` pairs for sample `i` and a given `coverage_id`.
Note that `i` ranges `0...5+sample_size+1`, where 5 represents the 5 negative sidx.
Therefore, `i=0` corresponds to `sidx=-5`, `i=1` to `sidx=-4`, etc.
ktools code [here](https://github.com/OasisLMF/ktools/blob/e803eac186bc6fff6d11a257b8c90b5238f0b8cd/src/gulcalc/gulcalc.cpp#L319):
```c++
void gulcalc::writemode1output(const int event_id, const OASIS_FLOAT tiv,
			       std::vector<std::vector<gulItemIDLoss>> &gilv,
			       const bool correlated=false) {

	// Set losses for alloc rule where
	// total peril loss = maximum subperil loss
	if (alloc_rule_ == 2) setmaxloss(gilv);

	std::map<int, std::vector<gulSampleslevelRec>> gxi;

	// Check whether the sum of losses per sample exceed TIV
	// If so, split TIV in proportion to losses
	for (size_t i = 0; i < gilv.size(); i++) {

		splittiv(gilv[i], tiv);
                [...]
```
Definition of `splittiv` (ktools code [here](https://github.com/OasisLMF/ktools/blob/e803eac186bc6fff6d11a257b8c90b5238f0b8cd/src/gulcalc/gulcalc.cpp#L208)):
```c++
void splittiv(std::vector<gulItemIDLoss> &gulitems, OASIS_FLOAT tiv)
{
// if the total loss exceeds the tiv 
// then split tiv in the same proportions to the losses
	if (tiv > 0) {
		OASIS_FLOAT total_loss = 0;
		auto iter = gulitems.begin();
		while (iter != gulitems.end()) {
			total_loss += iter->loss;
			iter++;
		}
		if (total_loss > tiv) {
			iter = gulitems.begin();
			while (iter != gulitems.end()) {
				double percentage = iter->loss / total_loss;
				iter->loss = percentage * tiv;
[...]

PR #1098, opened 2022-08-09
Implement multiplicative method for total loss computation and new `alloc_rule=3`

### LINKED ISSUE #1099: Implement multiplicative method for total loss computation
This issue is to track progress in implementing a new gulcalc allocrule for total peril loss using the multiplicative method in gulpy https://github.com/OasisLMF/ktools/issues/118.

mtazzari:
This PR implements a new way of computing the total losses using multiplicative expression for sub-peril losses and, respectively, a new alloc_rule value: 3.

This PR tracks implementation of this feature in gulpy. Details are in https://github.com/OasisLMF/ktools/issues/118

Fix #1099 

TODO:

- [x] check that gulpy produces same results as before for alloc_rule = 0, 1, 2.

This PR implements a new formula for computing the total losses using multiplicative expression for sub-peril losses.
This implementation is accessible using the new alloc_rule = 3, i.e. with:
```bash
gulpy -a3 ...
```
After computing total losses using the new formula, back-allocation in proportion to the losses is applied to the sub-peril losses, similarly to the back-allocation applied with `alloc_rule=1`.

More details are in https://github.com/OasisLMF/ktools/issues/118

### DIFF
--- a/oasislmf/pytools/gul/core.py
+++ b/oasislmf/pytools/gul/core.py
@@ -127,17 +127,13 @@ def split_tiv_multiplicative(gulitems, tiv):
         gulitems (numpy.array[oasis_float]): array containing losses of all items.
         tiv (oasis_float): total insured value.
     """
-    # gulitems are in tiv units, so it's faster to use a `scale` factor rather than
-    # computing ratios for all gulitems
     Ngulitems = gulitems.shape[0]
-    scale = tiv**Ngulitems
     total_loss = 1.
     for i in range(Ngulitems):
-        total_loss *= tiv - gulitems[i]
+        total_loss *= 1. - gulitems[i] / tiv
 
-    total_loss = scale - total_loss
-    total_loss *= tiv / scale
-
+    total_loss = 1. - total_loss
+    total_loss *= tiv
     if total_loss > 0.:
         # back-allocate proportionally in any case (i.e., not only if total_loss > tiv)
         f = tiv / total_loss
--- a/oasislmf/pytools/gul/core.py
+++ b/oasislmf/pytools/gul/core.py
@@ -128,15 +128,17 @@ def split_tiv_multiplicative(gulitems, tiv):
         tiv (oasis_float): total insured value.
     """
     Ngulitems = gulitems.shape[0]
-    total_loss = 1.
+    undamaged_value = 1.
+    sum_loss = 0.
     for i in range(Ngulitems):
-        total_loss *= 1. - gulitems[i] / tiv
+        undamaged_value *= 1. - gulitems[i] / tiv
+        sum_loss += gulitems[i]
 
-    total_loss = 1. - total_loss
-    total_loss *= tiv
-    if total_loss > 0.:
+    multiplicative_loss = tiv * (1. - undamaged_value)
+
+    if sum_loss > 0.:
         # back-allocate proportionally in any case (i.e., not only if total_loss > tiv)
-        f = tiv / total_loss
+        f = multiplicative_loss / sum_loss
 
         for j in range(Ngulitems):
             # editing in-place the np array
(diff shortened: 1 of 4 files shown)


mtazzari:
@sstruzik FYI, in manager.py I also fix a typo that we left when we refactored gulpy the first time. We changed the `losses` array to store the negative sidx samples at the end of the array, instead of at the beginning using my padding, and we forgot that split_tiv only had to be applied to the positive sidx:
```py
for sample_i in range(1, losses.shape[0]):
```
should have been
```py
for sample_i in range(1, losses.shape[0] - 5):
```

johcarter:
is split_tiv anything to do with the sub-peril backallocation, @mtazzari ? from memory, I think the alloc rule should also apply to sidx -1, -3, -5

oasislmf/pytools/gul/core.py
```diff
@@ -116,6 +116,33 @@ def split_tiv(gulitems, tiv):
...
+    total loss as tiv * (1 - (1-A)*(1-B)*(1-C)...), where A, B, C are damage ratios
+    computed as the ratio between a sub-peril loss and the tiv. Sub-peril losses
+    in gulitems are always back-allocated proportionally to the losses.
+
+    Args:
+        gulitems (numpy.array[oasis_float]): array containing losses of all items.
+        tiv (oasis_float): total insured value.
+    """
+    Ngulitems = gulitems.shape[0]
+    total_loss = 1.
```
sstruzik, line 131: I would call that insure_value_left or undamaged_value or 1_minus_total_loss.
I prefer to avoid confusion when reading as until line 135 it is not the total loss

mtazzari:
@johcarter 
> is split_tiv anything to do with the sub-peril backallocation, @mtazzari ? from memory, I think the alloc rule should also apply to sidx -1, -3, -5

you're right, we had an issue open in ktools and oasislmf as well: https://github.com/OasisLMF/OasisLMF/issues/1003
Current implementation in gulcalc applies split_tiv to all sidx samples.
I suggest we keep it like this in this PR and we address this in general in another PR.

oasislmf/pytools/gul/core.py
```diff
@@ -116,6 +116,33 @@ def split_tiv(gulitems, tiv):
...
+    for i in range(Ngulitems):
+        total_loss *= 1. - gulitems[i] / tiv
+
+    total_loss = 1. - total_loss
+    total_loss *= tiv
+    if total_loss > 0.:
```
sstruzik, line 137: I think this is the same and a bit less computation:
if insure_value_left < 1:
          f = 1/(1-insure_value_left)
          for j in range(Ngulitems):
            # editing in-place the np array
            gulitems[j] *= f
Can you double check?

oasislmf/pytools/gul/manager.py
```diff
@@ -379,17 +379,23 @@ def write_losses(event_id, sample_size, loss_threshold, losses, item_ids, alloc_
...
     """
-    # note that Nsamples = sample_size + NUM_IDX
-
     if alloc_rule == 2:
         losses[1:] = setmaxloss(losses[1:])
 
     # split tiv has to be executed after setmaxloss, if alloc_rule==2.
-    if alloc_rule >= 1 and tiv > 0:
+    if tiv > 0:
         # check whether the sum of losses-per-sample exceeds TIV
```
sstruzik, line 387: remove comments as not relevant anymore

mtazzari: ok done

oasislmf/pytools/gul/core.py
mtazzari: doesn't apply anymore I think

oasislmf/pytools/gul/core.py
mtazzari: changed `total_loss` to `undamaged_value`

sstruzik (approved):
looks good

PR #1101, opened 2022-08-23
Always create a correlations.bin, if missing model_settings file is b…

sambles:
### Fix for running the MDK without a model_settings file
* The file generation step always creates a `correlation.csv` file, when no model_settings file is given, the file is empty. 

Tests conducted:

- [x] when model_settings file is given and no correlations are defined -> correlations.bin and correlations.csv should be created empty
- [x] when model_settings file is given and correlations are defined  but not applied to perils -> correlations.bin and correlations.csv should be created empty
- [x] when model_settings file is given and correlations are defined and applied to perils -> correlations.bin and correlations.csv should be created and populated as expected


mtazzari (approved):
Perfect fix

PR #1104, opened 2022-08-26
Facilitate class event rates in Moment Event Loss Table (MELT) output

hchagani-oasislmf:
### Facilitate classic event rates in Moment Event Loss Table (MELT) output
If the `event_dict.bin` (event dictionary) file exists, it is copied to the `input directory. This file gives event rates for each event ID (see `ktools` PR https://github.com/OasisLMF/ktools/pull/322). In a similar fashion to `events.bin`, an event dictionary file ID can be defined in `analysis_settings.json` so that multiple event dictionary files can be stored in the same model files directory:

```
.
    "model_settings": {
        "event_set": "p",
	"event_dict_set": "p",
    }
.
```
where `event_dict_set` is the ID of the event dictionary file (in this case `event_dict_p.bin`).

If the file does not exist, ktools component `eltcalc` will calculate event rates from the occurrence file, which is the current mode of operation.


hchagani-oasislmf:
`eltcalc` should extract event rates from a dedicated `event_rates.csv` file instead of the event dictionary file. Closing PR to keep things tidy as a significant number of changes have been made.

PR #1107, opened 2022-09-07
Fix/678 logger

sstruzik:
### Fix set logger configuration for oasislmf only
When setting logger, OasisLmf was actually changing the log configuration of all the module by calling logging.getLogger() instead of logging.getLogger('oasislmf')


sstruzik:
fix for https://github.com/OasisLMF/OasisPlatform/issues/678

sambles (approved):

PR #1114, opened 2022-09-21
Facilitate class event rates in Moment Event Loss Table (MELT) output

hchagani-oasislmf:
### Facilitate class event rates in Moment Event Loss Table (MELT) output
If the `event_rates.csv` file exists, it is copied to the `input` directory. This file gives event rates for each event ID (see ktools PR https://github.com/OasisLMF/ktools/pull/327). In a similar fashion to `events.bin`, and event dictionary file can be defined in `analysis_settings.json` so that multiple event dictionary files can be stored in the same model files directory:

```
.
    "model_settings": {
        "event_set": "p",
	"event_rates_set": "p",
    }
.
```
where `event_rates_set` is the ID of the event rates file (in this case `event_rates_p.csv`).

If this file does not exist, ktools component `eltcalc` will calculate event rates from the occurrence file, which is the current mode of operation.


sambles (approved):
Looks good (tested using the example `event_rates.csv` shown in https://github.com/OasisLMF/ktools/pull/327)

PR #1116, opened 2022-09-22
Hide geopandas warning

### LINKED ISSUE #1115: Logs cluttered with warnings from geopandas about GEOS versions
## Issue Description

Lots of clutter in the logs, primarily messages like this one, making them hard to read:
```
...lib64/python3.8/site-packages/geopandas/_compat.py:112: UserWarning: The Shapely GEOS version (3.8.0-CAPI-1.13.1 ) is incompatible with the GEOS version PyGEOS was compiled with (3.10.3-CAPI-1.16.1). Conversions between both will be slow.
```

I think it's made worse when the vendor code in the complex model starts multiple processes because each process will display this warning.

As far as I know there's no simple way to properly fix this but I'm also not aware of any serious consequences for the relatively small use that we (and Oasis?) make of geopandas. [Shapely v2](https://github.com/shapely/shapely/issues/782) will include PyGEOS and may completely bypass this issue.

1) Run an analysis using a "complex model". 
2) Look at the logs.

OasisLMF 1.26.3

carlfischerjba:
### Hide geopandas compatibility warnings from logs
Fixes #1115.
Hide numerous messages like the following from the logs:
```
...lib64/python3.8/site-packages/geopandas/_compat.py:112: UserWarning: The Shapely GEOS version (3.8.0-CAPI-1.13.1 ) is incompatible with the GEOS version PyGEOS was compiled with (3.10.3-CAPI-1.16.1). Conversions between both will be slow.
```

### DIFF
--- a/oasislmf/__init__.py
+++ b/oasislmf/__init__.py
@@ -58,10 +58,11 @@ class MyImport(MetaPathFinder):
             import_path = import_path[1]
             for deprecated in self.depricated_modules:
                 if deprecated == import_path or import_path.startswith(deprecated+'.'):
-                    warnings.simplefilter("always")
-                    warnings.warn(
-                        f"imports from 'oasislmf.{deprecated}' are deprecated. Import by using 'oasislmf.{self.depricated_modules[deprecated]}' instead."
-                    )
+                    with warnings.catch_warnings():
+                        warnings.simplefilter("always")
+                        warnings.warn(
+                            f"imports from 'oasislmf.{deprecated}' are deprecated. Import by using 'oasislmf.{self.depricated_modules[deprecated]}' instead."
+                        )
                     import_path = import_path.replace(deprecated, self.depricated_modules[deprecated])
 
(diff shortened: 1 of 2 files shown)


oasislmf/__init__.py
```diff
@@ -58,10 +58,11 @@ def find_spec(self, fullname, path=None, target=None):
             import_path = import_path[1]
             for deprecated in self.depricated_modules:
                 if deprecated == import_path or import_path.startswith(deprecated+'.'):
-                    warnings.simplefilter("always")
-                    warnings.warn(
-                        f"imports from 'oasislmf.{deprecated}' are deprecated. Import by using 'oasislmf.{self.depricated_modules[deprecated]}' instead."
-                    )
+                    with warnings.catch_warnings():
```
carlfischerjba, line 61: I'm not entirely sure this is necessary and haven't taken the time to confirm that it matters but if I've understood the Python docs this should prevent all warnings being reactivated across the entire application on the next line.

PR #1119, opened 2022-10-14
use correct condpriority to fix cond class exclusion

sstruzik:
### Fix for cond class exclusion when multiple priority are used
Add a loop breaker in case account cond hierarchy lead to cycling cond tag infinitely.

example of oed that had issue treating cond class:
[account.csv](https://github.com/OasisLMF/OasisLMF/files/9784457/account.csv)
[location.csv](https://github.com/OasisLMF/OasisLMF/files/9784458/location.csv)

### DIFF
--- a/oasislmf/preparation/il_inputs.py
+++ b/oasislmf/preparation/il_inputs.py
@@ -572,7 +572,7 @@ def __merge_gul_and_account(gul_inputs_df, accounts_df, fm_terms, oed_hierarchy)
 
     if 'condpriority' not in accounts_df.columns:
         accounts_df['condpriority'] = 0
-    accounts_df['condpriority'].fillna(0)
+    accounts_df['condpriority'].fillna(0, inplace=True)
 
     policy_df = accounts_df.drop_duplicates(subset=[portfolio_num, acc_num, policy_num, 'layer_id']).drop(columns=[cond_tag, 'condpriority'])
     cond_df = accounts_df[[portfolio_num, acc_num, cond_tag, 'condpriority']].drop_duplicates()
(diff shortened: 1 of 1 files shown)


johcarter:
Fixes #1122 , and tests all pass

oasislmf/preparation/il_inputs.py
```diff
@@ -563,9 +563,22 @@ def __merge_gul_and_account(gul_inputs_df, accounts_df, fm_terms, oed_hierarchy)
     cond_class = oed_hierarchy['condclass']['ProfileElementName'].lower()
     loc_num = oed_hierarchy['locnum']['ProfileElementName'].lower()
 
-    policy_df = accounts_df.drop_duplicates(subset=[portfolio_num, acc_num, policy_num, 'layer_id']).drop(columns=cond_tag)
-    cond_df = accounts_df[[portfolio_num, acc_num, cond_tag]].drop_duplicates()
-    all_cond_policy =  pd.merge(policy_df, cond_df, on=[portfolio_num, acc_num])
+    if cond_class in accounts_df.columns:
```
mtazzari, line 566: can you add a small (even one-liner) comment about the logic of what we're doing here below?
either before the `if` or inside each case, so it is clear what action has to be taken depending on whether `cond_class` is in `df.columns`.
e.g.
```py
if cond_class in accounts_df.columns:
    # in this case, we ...
else:
    # ...

File: oasislmf/preparation/il_inputs.py [hunk 2]
```diff
@@ -563,9 +563,22 @@ def __merge_gul_and_account(gul_inputs_df, accounts_df, fm_terms, oed_hierarchy)
...
+                                                  [cond_class]
+                                                  .transform(max))
+    else:
+        accounts_df[['polcondclass', cond_class]] = 0
+
+    if 'condpriority' not in accounts_df.columns:
+        accounts_df['condpriority'] = 0
+    accounts_df['condpriority'].fillna(0, inplace=True)
+
+    policy_df = accounts_df.drop_duplicates(subset=[portfolio_num, acc_num, policy_num, 'layer_id']).drop(columns=[cond_tag, 'condpriority'])
```
mtazzari, line 577: please add a short comment before these manipulations of `accounts_df` to explain what we need to achieve.

mtazzari (changes requested):
Quantitative tests were conducted by Joh, and she reported that they pass as expected.
So here I've just reviewed code style.
It'd be great if you could add perhaps a small comment  as indicated.

mtazzari (approved):

PR #1130, opened 2022-10-20
Remove check for IL + RI files from the run model cmd

sambles:
### Remove redundant IL+RI check
With the addition of https://github.com/OasisLMF/OasisLMF/pull/1112  there are two input file checks, one which can be toggled on/off via the flag ` --check-missing-inputs`  and another which is hard coded in the `oasislmf run model` command. 

Removed the check from `oasislmf run model` which is a 'compound' command of [`generate-files` + `generate-losses`]

### DIFF
--- a/oasislmf/computation/generate/files.py
+++ b/oasislmf/computation/generate/files.py
@@ -164,17 +164,6 @@ class GenerateFiles(ComputationStep):
 
         il = bool(exposure_data.account)
         ri = exposure_data.ri_info and exposure_data.ri_scope and il
-
-        # Send warning if RI files are in args without IL
-        if any([self.oed_info_csv, self.oed_scope_csv]) and not ri:
-            self.logger.warn(
-                'RI option indicated by provision of some RI related assets, but other assets are missing. '
-                'To generate RI inputs you need to provide all of the assets required to generate direct '
-                'Oasis files (GUL + FM input files) plus all of the following assets: '
-                '  * reinsurance info file'
-                '  * reinsurance scope file'
-            )
-
         self.logger.info('\nGenerating Oasis files (GUL=True, IL={}, RIL={})'.format(il, ri))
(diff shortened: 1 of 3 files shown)


mtazzari (changes requested):
Here we remove a check performed during/at the end of the generation of the input files, since the same check can now be carried out with 
```bash
oasislmf model run --check-missing-inputs
```
Avoiding duplication is fine to me, however: now that we remove this check from here, is it carried out anyway as part of the same functionality or are we losing the check when `run()` is called?
I see that `check_missing_inputs` is set to `False` by default, so it seems that we'll lose the check.
Can you please double check that this is the intended behaviour?

sambles:
Check is redundant since its run during ods-tool's exposure loading. Removed and wrapped that exception from ods-tools. 

**Example 1**
```
$ oasislmf model generate-oasis-files 
Warning: Unknown options(s) set in MDK config:
   model_data_dir : model_data/PiWind
   gulmc : True
   gulpy : False
   ktools_alloc_rule_gul : 1
   analysis_settings_json : analysis_settings.json
   modelpy : True

Stating oasislmf command - GenerateOasisFiles
  0%|                                                                                                                                             | 0/1 [00:00<?, ?it/s]
  0%|                                                                                                                                             | 0/1 [00:00<?, ?it/s]
Failed to load OED exposure files, OdsException: in account None
 Exposure needs account if reinsurance is provided account=None
```  

**Example 2**
```
$ oasislmf model run

Stating oasislmf command - RunModel
RI option indicated by provision of some RI related assets, but other assets are missing. To generate RI inputs you need to provide all of the assets required to generate direct Oasis files (GUL + FM input files) plus all of the following assets:     reinsurance info. file path,     reinsurance scope file path.
```

mtazzari (approved):

PR #1136, opened 2022-10-26
Fix: gulpy dynamic buff size

### LINKED ISSUE #1135: gulpy appears to hang when sample size is large
## Issue Description
When sample size is large, `gulpy` appears to hang. The `work/` directory contains empty output files for `summarycalc`, indicating that no data from `gulpy` is being piped to `summarycalc`. The run completes successfully when `getmodel` and `gulcalc` are used instead of `modelpy` and `gulpy`.

Run the PiWind model with at least 5000 samples.

MDK v1.26.3

mtazzari:
This PR fixes #1135 by dynamically increasing the buffer size `buff_size` to accommodate processing larger `coverages`.
`buff_size` is reset to its default value at the beginning of the processing of each event.

In this PR we introduce a feature where the output buffer is dynamically increased to support large number of samples for large coverages (i.e., coverages with many items).
As a reference, using the OasisPiWind model, this command would have resulted in `gulpy` hanging:
```bash
 eve 1 1 | modelpy | gulpy -S5000 -a1 -L0 | gultocsv
```
Now, the command works as expected for any large value of samples and for any large coverage.

### DIFF
--- a/oasislmf/pytools/gul/manager.py
+++ b/oasislmf/pytools/gul/manager.py
@@ -298,13 +298,14 @@ def run(run_dir, ignore_file_type, sample_size, loss_threshold, alloc_rule, debu
                 eps_ij = np.zeros((1, 1), dtype='float64')
 
             last_processed_coverage_ids_idx = 0
+            buff_size = GULPY_STREAM_BUFF_SIZE_WRITE
             while last_processed_coverage_ids_idx < compute_i:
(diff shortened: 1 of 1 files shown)


hchagani-oasislmf (approved):
Looks good, and testing shows it solves the issue.

sstruzik (approved):

PR #1137, opened 2022-10-26
Introducing `gulmc`: full monte carlo loss engine

mtazzari:
This PR introduces `gulmc`, a new tool that uses a "full Monte Carlo" approach for ground up losses calculation that, instead of drawing loss samples from the 'effective damageability' probability distribution (as done by calling `eve | modelpy | gulpy`), it first draws a sample of the hazard intensity, and then draws a sample of the damage from the vulnerability function corresponding to the hazard intensity sample.

### Introducing `gulmc`, full Monte Carlo loss calculation engine
This PR introduces `gulmc`, a new tool that uses a "full Monte Carlo" approach for ground up losses calculation that, instead of drawing loss samples from the 'effective damageability' probability distribution (as done by calling `eve | modelpy | gulpy`), it first draws a sample of the hazard intensity, and then draws a sample of the damage from the vulnerability function corresponding to the hazard intensity sample.

#### Comparing `gulpy` and `gulmc` output
`gulmc` runs the same algorithm of `eve | modelpy | gulpy`, i.e., it runs the 'effective damageability' calculation mode, with the same command line arguments. For example, to run a model with 1000 samples, alloc rule 1, and streaming the binary output to the `output.bin` file, can be done with:
```bash
eve 1 1 | modelpy | gulpy -S1000 -a1 -o output.bin
```
or
```bash
eve 1 1 | gulmc -S1000 -a1 -o output.bin
```
#### Hazard uncertainty treatment
If the hazard intensity in the fooprint has no uncertainty, i.e.:
```csv
event_id,areaperil_id,intensity_bin_id,probability
1,4,1,1
[...]
```
then `gulpy` and `gulmc` produce the same outputs. However, if the hazard intensity has a probability distribution, e.g.:
```csv
event_id,areaperil_id,intensity_bin_id,probability
1,4,1,2.0000000298e-01
1,4,2,6.0000002384e-01
1,4,3,2.0000000298e-01
[...]
```
then, by default, `gulmc` runs the full Monte Carlo sampling of the hazard intensity, and then of damage. In order to reproduce the same results that `gulpy` produces can be achieved by using the `--effective-damageability` flag:
```bash
eve 1 1 | gulmc -S1000 -a1 -o output.bin --effective-damageability
```
#### On the usage of `modelpy` and `eve` with `gulmc`
Due to internal refactoring, `gulmc` now incorporates the functionality performed by `modelpy`, therefore `modelpy` should not be used in a pipe with `gulmc`:
```bash
eve 1 1 | modelpy | gulpy -S1000 -a1 -o output.bin        # wrong usage, won't work
eve 1 1 | gulpy -S1000 -a1 -o output.bin                  # correct usage
```
> **Note:** both `gulpy` and `gulmc` can read the events stream from binary file, i.e., without the need of `eve`, with:
> ```bash
>  gulmc -i input/events.bin -S1000 -a1 -o output.bin
> ```

#### Printing the random values used for sampling
Since we now sample in two dimensions (hazard intensity and damage), the `-d` flag is revamped to output both random values used for sampling. While `gulpy -d` printed the random values used to sample the effective damageability distribution, in `gulmc`:
```bash
gulmc -d1 [...]   # prints the random values used for the hazard intensity sampling
gulmc -d2 [...]   # prints the random values used for the damage sampling
```
> **Note:** if the `--effective-damageability` flag is used, only `-d2` is valid since there is no sampling of the hazard intensity, and the random value printed are those used for the effective damageability sampling.

> **Note:**  if `-d1` or `-d2` are passed, the only valid `alloc_rule` value is `0`. This is because, when printing the random values, back-allocation is not meaningful. `alloc_rule=0` is the default value or it can be set with `-a0`. If a value other than 0 is passed to `-a`, an error will be thrown.

#### Testing suite
This PR introduces:
 - a minimal toy model in `tests/assets/test_model_1/` that can be used to run unit tests on various functionality in the repository. A more detailed description of the content of the model can be found at  `tests/assets/test_model_1/README.md`.
 - a suite of 192 quantitative tests for the `gulmc` output for combinations of input parameters (alloc rule, correlation, etc.). Binary files with the expected outputs are stored at `tests/assets/test_model_1/expected/`.
 - a suite of 48 quantitative tests for the `gulpy` output for combinations of input parameters.
 - tests checking that `ValueErrors` in `gulmc` are raised when expected.


sambles (approved):
* **Nice to have:**  It would be useful if the gulmc testing could output the failed differences between expected and results. At the moment the output notes that files differ (true/false). However, the binary format needs to be converted to something human readable first. 

* **Note:**   gulmc testing has failures when running on a non-ubuntu based distro (arch linux based). We suspect it might be linked to OS libraries, but haven't found the cause. 

>   Marco Tazzari
if we can get tests running on other OSs then we can see what we can support, which would be even better, but being so focused on numerical performance, it's fair to pick a specific OS and rely on its reproducibility
rather than aiming at a Windows-style "we work on any computer", which it's going to be hard to maintain
  
> Stephane Struzik
I don't agree on this one, I think the test should pass on all platform we advertise. If we just have rounding error I would prefer that we convert the result to csv and use dataframe to compare with the added benefit of providing a better diff message

mtazzari:
Thank you @sambles for the comments. 
Regarding the "nice to have", I'm coding up a handy binary-file-check function customized for us so that it converts to csv if the binary differs and shows the differences. I'll add that to PR #1168 which is a continuation of this PR, with additional functionality.

oasislmf/pytools/gulmc/manager.py
```diff
@@ -0,0 +1,886 @@
...
+                        cumsum += haz_pdf[cdf_start + i]
+                        haz_cdf[cdf_start + i] = cumsum
+
+                    # compute effective damageability cdf (internally called `eff_vuln`)
+                    # for each vulnerability function associated to this areaperil_id, compute the product between
+                    # the hazard intensity probability and the damage probability.
+                    areaperil_to_vulns_idx = areaperil_to_vulns_idx_array[areaperil_to_vulns_idx_dict[last_areaperil_id]]
+                    for vuln_idx in range(areaperil_to_vulns_idx['start'], areaperil_to_vulns_idx['end']):
+                        eff_vuln_cdf_cumsum = 0.
+                        Ndamage_bins = 0
```
sstruzik, line 692: should the name Ndamage_bins be more damage_bin_i, even if at the end it will represent the number of damage_bin

mtazzari: I agree it is counter-intuitive as the name `Ndamage_bins` should typically indicate a total number of bins. 
For efficiency purposes, I am using it as the loop counter, and then I use its final value to avoid setting another variable. 
However, if you think that this would be better I can change it to this:
```py
bin_i = 0
while bin_i < Ndamage_bins_max:
    for i in range(Nbins_to_read):
        intensity_bin_i = event_footprint['intensity_bin_id'][last_areaperil_id_start + i] - 1
        eff_vuln_cdf_cumsum += vuln_array[vuln_idx,
                                          bin_i, intensity_bin_i] * haz_pdf[cdf_start + i]
    eff_vuln_cdf[eff_vuln_cdf_start + bin_i] = eff_vuln_cdf_cumsum
    bin_i += 1
    if eff_vuln_cdf_cumsum > 0.999999940:
              break
Ndamage_bins = bin_i
```

oasislmf/pytools/gulmc/manager.py
```diff
@@ -0,0 +1,886 @@
...
+        max_bytes_per_item = (sample_size + NUM_IDX + 1) * gulSampleslevelRec_size + 2 * gulSampleslevelHeader_size
+
+        while True:
+            if not streams_in.readinto(event_id_mv):
+                break
+
+            # get the next event_id from the input stream
+            event_id = event_ids[0]
+
+            event_footprint = (FootprintLayerClient if data_server else footprint_obj).get_event(event_id)
```
sstruzik, line 286: not massively important but I would set the object outside the loop to avoid the if each time

oasislmf/pytools/gulmc/manager.py
```diff
@@ -0,0 +1,886 @@
...
+                            # cap `haz_rval` to the maximum `haz_prob_to` value (which should be 1.)
+                            haz_rval = haz_rndms[rng_index][sample_idx - 1]
+
+                            if debug == 1:
+                                # store the random value used for the hazard intensity sampling instead of the loss
+                                losses[sample_idx, item_i] = haz_rval
+                                continue
+
+                            if haz_rval >= haz_prob_to[Nbins - 1]:
+                                haz_rval = haz_prob_to[Nbins - 1] - 0.00000003
```
sstruzik, line 540: I think haz_rval is not used after you set it here

oasislmf/pytools/gulmc/manager.py
```diff
@@ -0,0 +1,886 @@
...
+                vuln_rndms_base = generate_rndm(vuln_seeds[:rng_index], sample_size)
+
+                # generate the correlated samples for the whole event, for all peril correlation groups
+                if do_correlation:
+                    corr_seeds = generate_correlated_hash_vector(unique_peril_correlation_groups, event_id)
+                    eps_ij = generate_rndm(corr_seeds, sample_size, skip_seeds=1)
+
+                else:
+                    # create dummy data structures with proper dtypes to allow correct numba compilation
+                    corr_seeds = np.zeros(1, dtype='int64')
```
sstruzik, line 312: corr_seeds = np.zeros(1, dtype='int64') is not used later on

oasislmf/pytools/getmodel/manager.py
```diff
@@ -312,12 +313,15 @@ def get_vulns(static_path, vuln_dict, num_intensity_bins, ignore_file_type=set()
                 num_damage_bins = header[0]
             if "vulnerability.idx" in static_path:
                 logger.debug(f"loading {os.path.join(static_path, 'vulnerability.idx')}")
-                vulns_bin = np.memmap(os.path.join(static_path, "vulnerability.bin"), dtype=VulnerabilityRow, offset=4, mode='r')
-                vulns_idx_bin = np.memmap(os.path.join(static_path, "vulnerability.idx"), dtype=VulnerabilityIndex, mode='r')
+                vulns_bin = np.memmap(os.path.join(static_path, "vulnerability.bin"),
```
sstruzik, line 316: code is identical, I wouldn't reformat here as it respect our line length rule

oasislmf/pytools/gulmc/manager.py
sstruzik: why not damage_bin_i as we have multiple variables that are represented by bin?

sstruzik (approved):
Just few minor comment. The logic looks good.

oasislmf/pytools/gul/common.py
```diff
@@ -11,15 +11,22 @@
...
-PIPE_CAPACITY = 65536  # bytes
-GETMODEL_STREAM_BUFF_SIZE = 2 * PIPE_CAPACITY
-
 items_data_type = nb.from_dtype(np.dtype([('item_id', np.int32),
                                           ('damagecdf_i', np.int32),
                                           ('rng_index', np.int32)
                                           ]))
 
-coverage_type = nb.from_dtype(np.dtype([('tiv', np.float),
+# TODO: 'vulnerability_id' should be renamed to 'vulnerability_idx' because it is the
```
sstruzik, line 19: I recently changed my mind on TODOs. I think we should avoid them in the code and create separate issue of what needs to be done.

oasislmf/pytools/gulmc/manager.py
mtazzari: fine with `damage_bin_i`, I'll change to that

oasislmf/pytools/gulmc/manager.py
mtazzari: good catch, it definitely goes outside the loop

oasislmf/pytools/getmodel/manager.py
mtazzari: I agree, it must have been formatted when line length was the default 120. I'll revert it to one-line, it should fit in 150 characters

oasislmf/pytools/gulmc/manager.py
mtazzari: that's right `haz_rval` is not used after here. I'll drop line 540.

oasislmf/pytools/gulmc/manager.py
mtazzari: good catch. Probably in earlier versions I was passing `corr_seeds` to the main function, but not anymore. I'll remove it.

oasislmf/pytools/gulmc/manager.py
mtazzari: done

oasislmf/pytools/gulmc/manager.py
mtazzari: done, I've changed this in two other points of the code.

oasislmf/pytools/gulmc/manager.py
mtazzari: done

oasislmf/pytools/gulmc/manager.py
mtazzari: done

oasislmf/pytools/gul/common.py
mtazzari: actually, the name is correct. in weighted vulnerability we're going to store again the actual vulnerability_id in that field, so no need to rename it.

oasislmf/pytools/getmodel/manager.py
mtazzari: done

PR #1140, opened 2022-11-04
[gulpy] bugfix: use the correct itemsize when creating `int32_mv`

mtazzari:
This PR is a quick bug fix to solve an issue with the `int32_mv` ndarray that covers the memory view  used to store the gul losses before streaming them out.
Before this PR, `int32_mv` was correctly initialised as `i4` dtype (`int32`), but with a wrong number of elements.
This was causing a wrong output of `gulpy` (losses all zero for entire items) for large number of samples, without throwing an error.
 

### gulpy bugfix
This PR solves Issue #1141 where a bug that was causing a wrong output of `gulpy` (losses all zero for entire items) for large number of samples, without throwing an error.

### DIFF
--- a/oasislmf/pytools/gul/common.py
+++ b/oasislmf/pytools/gul/common.py
@@ -13,7 +13,6 @@ gul_header = np.int32(1 | 2 << 24).tobytes()
 
 PIPE_CAPACITY = 65536  # bytes
 GETMODEL_STREAM_BUFF_SIZE = 2 * PIPE_CAPACITY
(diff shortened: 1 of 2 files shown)


sambles (approved):
I think Stephane is the more important reviewer here, looks good at a glance and CI is passing with this fix
https://github.com/OasisLMF/OasisPiWind/actions/runs/3393836962

sstruzik (approved):

PR #1143, opened 2022-11-07
Try to omit /opt/ folder

sambles:
### Filter '/opt/*' from code coverage 
* Tox incorrectly reports a lower coverage due to duplicated test files from the package install dir.


sambles:
change is small, so moving to piwind test update PR https://github.com/OasisLMF/OasisLMF/pull/1159

PR #1146, opened 2022-11-08
Apply PEP8 to the whole codebase and fix some flake8 errors

### LINKED ISSUE #1132: Make code PEP8 compliant
Currently only portions of the codebase are PEP8 compliant.

Extend the PEP8 compliance to the full codebase.

mtazzari:
In this PR:
1) we Fix #1132 as we make all the codebase PEP8 compliant, with the following command ran at the repository root:
```bash
autopep8 --in-place --recursive --max-line-length 150 --ignore 402 ./
```
We used with the following customization:
 - max line length is set to 150 (where 79 was the default);
 - we ignore error `E402` (docs [here](https://www.flake8rules.com/rules/E402.html)) which forces all module imports at top of file. This might clash with some initializations (e.g. the logging module) that we want to perform before importing other modules.

2) We addressed the following flake8 errors throughout the codebase:
```
F401: module imported but unused
F522: .format(...) unused named arguments
F524: .format(...) missing argument
F541: f-string without any placeholders
```
running the following command at the repository root:
```bash
flake8  --select F401,F522,F524,F541  oasislmf/ 
```

3) We set up an automated CI workflow called `Code Quality` that:
 - fails if newly pushed code is not PEP8 compliant (uses `autopep8`).
 - fails if newly pushed code introduces a flake8 error among those above.

Issue #1147 tracks the job required to address all flake8 errors throughout the codebase, which is likely to require some non-obvious refactoring and thus an accurate integration testing.

### Improving Code Quality
This PR makes all the codePEP8 compliant and introduces automatic CI checks to preserve PEP8 compliance.
This PR fixes some flake8 errors and introduces automatic CI checks to avoid that the same errors are introduced in the code in the future.


mtazzari:
@sambles have a look at these changes. What do you think? 
Here I'm not using the `--aggressive` flag, and I've allowed lines 150 characters long.

benhayes21 (approved):
Looks fine to me. 

I've only scanned the changed files quickly but it looks consistent. I find some of the original formatting easier to read, but happy to go with this for enforced consistency.

maxwellflitton (approved):

sambles (commented):
Looks much better without the `--aggressive` flag 

Question on the linting call, using `--exit-zero`,  why not have that fail and cleanup some of the issues it raises too? 

I think it makes sense to ignore some of the formatting warnings (since it should be PEP8 anyway), something like `flake8 --show-source oasislmf/ --ignore W503,W504`.   

-  W504 line break after binary operator
- W503 line break before binary operator
 .. etc with whichever warns we don't care about

mtazzari:
Thank you @sambles I'll have a look now at how we can fix some of the warnings raised by flake8. They seemed quite a few, so I don't know how much time it'd have taken to fix them. But certainly having also flake8 as a 'live' check that makes the workflow fail if it fails is the right direction to go.

mtazzari:
@sambles if you have a look at the failed test `Oasislmf Testing / unittest (3.8) (push)` you'll notice that most of the failures have to do with the fact that I had to insert `#noqa` at the end of a very long line, which generates the bash scripts.
Since the test checks the identity of the bash file to a reference file, it fails.

Any ideas how to get around this?
One way would be updating the reference bash scripts, but I feel it is too much a job for this PR, and I find it risky to do, given that code coverage is only 40%.
We could do a separate PR on flake8, which goes into that level of fixing and requires more thorough testing.
Therefore, I'd be inclined to roll back to commit https://github.com/OasisLMF/OasisLMF/pull/1146/commits/a1fa5999fd9a6afb3fe0630a63f909d6daa3f117
and merge this PR.
What do you think?

sambles:
@mtazzari  

Yup makes sense to role back and merge this for the PEP8 checking, it makes sense to start a new PR for flake if its a bigger job

> We could do a separate PR on flake8, which goes into that level of fixing and requires more thorough testing.
> Therefore, I'd be inclined to roll back to commit https://github.com/OasisLMF/OasisLMF/commit/a1fa5999fd9a6afb3fe0630a63f909d6daa3f117

mtazzari:
@sambles I've reverted back as we discussed. Then, I've just implemented innocuous fixes to flake8 errors:
```
F401: module imported but unused
F522: .format(...) unused named arguments
F524: .format(...) missing argument
F541: f-string without any placeholders
```
I've now enforced flake8 to fail the CI workflow if these 4 errors are introduced in new code.
I've opened Issue #1147 to track fixing all flake8 errors.

sambles (approved):
Looks good, thanks for the changes

PR #1153, opened 2022-11-17
Fix generation of dummy model footprint and vulnerability files in cases where there are no impacted bins

hchagani-oasislmf:
### Fix generation of dummy model footprint and vulnerability files in cases where there are no impacted bins
When generating dummy model files, the vulnerability sparseness is the percentage of bins impacted for a vulnerability at an intensity level and the intensity sparseness is the percentage of bins impacted for an event and areaperil. Setting these at relatively low values can result in cases where an areaperil contains no impacted bins. This can cause problems with division by zero warnings during normalisation.

Entries with no impacted intensity bins have been dropped from the generated vulnerability file. Additionally, in cases where there are no vulnerable damage bins, the first damage bin probability is set to 1 as this is always generated as the zero loss bin.


mtazzari (changes requested):
Hi Hassan, sorry for the unsolicited review - I've just noticed a few things and thought to feed them back.

1) Great job adding proper documentation throughout the code, it's really helpful, especially the math equations used!!

2) Looking at the docstrings, I think it'd be great if you could update them to make them uniform with what we're typically using, e.g.:
```py
def func(par1, par2):
    """
    This function does that.

    Args:
        par1 (int): first parameter.
        part2 (float): second parameter.

    Returns:
        result (float): result.
    
    Raises:
        ValueError: if this happens.
    """
    result = part2 * part 1
    if result < 0:
        raise ValueError("this happened")

    return result
```
I see, e.g.:
```py
def generate_data(self):
    """
    Generate Footprint binary dummy model file data.
    Yields:
        areaperil ID: int
        intensity bin ID: int
        probability: float
    """
```
should be:
```py
def generate_data(self):
    """
    Generate Footprint binary dummy model file data.

    Yields:
        areaperil (int): <description>
        intensity_bin (int): <description>
        probability (float): description
    """
```
where:
 - the format for listing the parameters is `parameter name (data type): description.`, 
 - the name of the parameters is as close as possible to the variable names used.

I think it'd be great if you could update the docstrings to be in line with our format.

3) regarding testing, is there a test that fails if the issue that you're solving here is not solved?
Since you raised the right comment of having tests for new features or bugfixes, let me say that I'm finding it surprisingly easy to add tests using `pytest`.
I think you could add a test, e.g., in `OasisLMF/tests/computation/test_dummy_model.py` that checks that the result is correct (or that the numpy `RuntimeWarning` is not raised anymore).
I'd prefer the first option, i.e. to validate the result quantitatively, as numpy may change the warning in the future.
I guess you can import the dummy model classes from `test_dummy_model.py` and get the result, without requiring to run `oasislmf ...` shell command.
It's not an error, but still, if it leads to inconsistent results, for us is an error. 
Let me know if you have any questions.

Thanks!

oasislmf/computation/data/dummy_model/generate.py
```diff
@@ -105,42 +172,135 @@ def generate_data(self):
...
+    """
+
+    def __init__(self, directory):
+        """
+        Initialise Footprint index file class.
+
+        Args:
+            directory (str): dummy model file destination.
+        """
+        self.dtypes = OrderedDict([
```
sstruzik, line 237: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6
as of python 3.6, dict are ordered (in 3.7 it is a feature), I would use dict here

sstruzik (approved):

benhayes21 (approved):

hchagani-oasislmf:
Hi @mtazzari, given the release schedule and that the docstrings have been changed as you suggested, would it be OK to merge this into `develop` please? You can create a new issue for developing the unit tests, and I would be happy to implement these in the next release.

oasislmf/computation/data/dummy_model/generate.py
mtazzari: I agree with Hassan that's better be explicit and use `OrderedDict` if we need the ordering.

PR #1160, opened 2022-11-24
Don't write temp files during test

carlfischerjba:
### Oasislmf Testing fixes 
* test_reinsurance.py  - removed unused debugging files 
* test_bash.py - fix for running individual tests
* flake8 - Newer versions of flake8 only recognise codes that match this pattern (single letter, three digits)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ python_classes =
 [flake8]
 exclude = .tox,.git,*/migrations/*,*/static/CACHE/*,docs,.ropeproject,.hypothesis
 max-line-length = 150
-ignore = E501,402
+ignore = E501,E402
 
 [coverage:run]
 branch = true
--- a/tests/model_execution/test_bash.py
+++ b/tests/model_execution/test_bash.py
@@ -7,7 +7,9 @@ import shutil
 from tempfile import NamedTemporaryFile
 from unittest import TestCase
 
-from oasislmf.model_execution.bash import genbash, create_bash_outputs, create_bash_analysis, bash_wrapper, bash_params
+from oasislmf.execution.bash import (bash_params, bash_wrapper,
+                                     create_bash_analysis, create_bash_outputs,
+                                     genbash)
 from oasislmf.utils import diff
 
 TEST_DIRECTORY = os.path.dirname(__file__)
@@ -882,6 +884,7 @@ class Genbash(TestCase):
 class Genbash_GulItemStream(Genbash):
     @classmethod
     def setUpClass(cls):
+        super().setUpClass()
         # test dirs
         cls.KPARSE_INPUT_FOLDER = os.path.join(TEST_DIRECTORY, "kparse_input")
         cls.KPARSE_OUTPUT_FOLDER = os.path.join(TEST_DIRECTORY, "itm_kparse_output")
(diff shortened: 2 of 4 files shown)


tests/model_preparation/test_reinsurance.py
```diff
@@ -3055,10 +3047,6 @@ def test_single_pol_level_fac_with_direct_layers(self):
             ri_info_df,
             ri_scope_df)
 
-        ri_inputs[0].ri_inputs.fm_policytc.to_csv('/tmp/fm_policytc.csv', index=False)
-        ri_inputs[0].ri_inputs.fm_programme.to_csv('/tmp/fm_programme.csv', index=False)
-        ri_inputs[0].ri_inputs.fm_profile.to_csv('/tmp/fm_profile.csv', index=False)
-
```
carlfischerjba, line 3061: These look like some debugging that was never deleted. They are not necessary for the tests to pass.

Worse than leaving temporary files lying around, they prevent two users from running this test on a shared machine. The test will fail for all but the first user.
1. User A runs the test which creates `/tmp/fm_....csv` owned by A.
2. User B run the test which tries to create the same files but can't because they already exist and are only (over)writeable by user A. Test fails.

tests/model_execution/test_bash.py
```diff
@@ -882,6 +884,7 @@ def test_analysis_settings_4_0_reins_iters_chunk(self):
 class Genbash_GulItemStream(Genbash):
     @classmethod
     def setUpClass(cls):
+        super().setUpClass()
```
carlfischerjba, line 887: Without calling `super()` we cannot run individual tests or test classes (with pytest or unittest) and we are forced to run the full file.

```
$ pytest ./tests/model_execution/test_bash.py::Genbash_TempDir::test_il_summarycalc_20_partition_chunk
...
        params = bash_params(
            max_process_id=num_partitions,
            analysis_settings=analysis_settings,
            num_reinsurance_iterations=(num_reinsurance_iterations or self.ri_iterations),
            fifo_tmp_dir=(fifo_tmp_dir or self.fifo_tmp_dir),
            stderr_guard=(stderr_guard or self.stderr_guard),
            gul_alloc_rule=(gul_alloc_rule or self.gul_alloc_rule),
            il_alloc_rule=(il_alloc_rule or self.il_alloc_rule),
            ri_alloc_rule=(ri_alloc_rule or self.ri_alloc_rule),
>           num_gul_per_lb=self.num_gul_per_lb,
            num_fm_per_lb=self.num_fm_per_lb,
            event_shuffle=self.event_shuffle,
            bash_trace=(bash_trace or self.bash_trace),
            gul_legacy_stream=(gul_legacy_stream or self.gul_legacy_stream),
            fmpy=(fmpy or self.fmpy),
        )
E       AttributeError: 'Genbash_TempDir' object has no attribute 'num_gul_per_lb'

tests/model_execution/test_bash.py:132: AttributeError
================================================================================= short test summary info =================================================================================
FAILED tests/model_execution/test_bash.py::Genbash_TempDir::test_il_summarycalc_20_partition_chunk - AttributeError: 'Genbash_TempDir' object has no attribute 'num_gul_per_lb'
============================================================================== 1 failed, 2 warnings in 0.47s ==============================================================================
```

setup.cfg
```diff
@@ -11,7 +11,7 @@ python_classes =
 [flake8]
 exclude = .tox,.git,*/migrations/*,*/static/CACHE/*,docs,.ropeproject,.hypothesis
 max-line-length = 150
-ignore = E501,402
+ignore = E501,E402
```
carlfischerjba, line 14: Newer versions of flake8 only recognise codes that match this pattern (single letter, three digits). I assume that `402` was a typo for `E402` [Module level import not at top of file](https://www.flake8rules.com/rules/E402.html) (seems like a common warning to disable).

sambles (approved):
Looks good thanks for the fixes

hchagani-oasislmf (approved):
Looks good.

PR #1161, opened 2022-11-24
Add support for Python 3.11

### LINKED ISSUE #1162: Add suppor for Python 3.11
## Issue Description
`oasislmf` supports Python up to version 3.10. Recently Python 3.11 has been released and it would be worth extending `oasislmf` support to Python 3.11.

Linux
oasislmf 1.27.0rc3

mtazzari:
In this PR we add Python 3.11 to the matrix of tests in `unittest.yml`, and we drop the specific test for python=3.10 and numba=0.55.1, which we introduced when numba 0.55.2 was released to ensure that we were testing the two most recent numba versions.
Since v0.55.2, `numba` has had 4 more releases and is now at v0.55.6, with no issues being reported across our tests, so we drop the specific test for numba 0.55.1.
This PR fixes  #1162 

### Add support for Python 3.11
This PR extends `oasislmf` support to Python version 3.11.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       ktools_branch:
-        description: 'If set, build ktools from scratch and use output to build package: [BranchName]'
+        description: "If set, build ktools from scratch and use output to build package: [BranchName]"
         required: false
         default: "develop"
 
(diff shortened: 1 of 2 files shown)


mtazzari:
The latest numba version 0.56.4 doesn't support Python 3.11 yet, but there is an actively developed PR to add that:
https://github.com/numba/numba/pull/8590

mtazzari:
Temporarily closing this. To be resumed when numba adds support for Python 3.11.

PR #1167, opened 2022-11-28
Add support for codecov.io

### LINKED ISSUE #1169: Track code coverage
## Issue Description
At the moment code coverage reports are produced as part of the unittest.yml workflow, but they are not uploaded to any services, like codecov.io that make it easy to visualize stats.

mtazzari:
This PR Fix #1169 by adding to the CI workflow the capability to upload code coverage reports to codecov.io to have a better graphical visualization of current coverage, and how code coverage changes with any new PR.

We cannot add a badge to the README right now because the codecov.io dashboard doesn't show any data (we need to merge this PR to master for that). A coverage badge can be added to the README after this PR is merged, in a future PR.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -73,5 +73,5 @@ jobs:
       - name: Run tox
         run: tox -e py
       - name: Upload coverage to Codecov
-        if: matrix.cfg.upload-cov == true && contains(fromJson('["pull_request", "master", "develop"]'), github.event_name)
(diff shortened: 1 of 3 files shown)


sambles (approved):
**Note:** Open question about tests running twice.  When a PR is opened the unit test workflow will have two triggers, one on `pull_request` and another on `pull`

We're not sure how  [codecov.io](http://codecov.io/) will handle the double upload, might need a follow up PR post merge

PR #1168, opened 2022-11-28
Implement weighted vulnerability feature in `gulmc`

### LINKED ISSUE #1126: Stochastic disaggregation 2 Vulnerability weightings
## Issue Description

Implement on the fly vulnerability weightings

### LINKED ISSUE #1188: `numpy.genfromtxt` returns a zero-dimensional array when reading 1-line csv files
## Issue Description
`numpy.genfromtxt` returns a zero-dimensional array when reading 1-line csv files, instead of a 1d array of length 1.
The issue is documented on the `numpy` repo [here](https://github.com/numpy/numpy/issues/20494), but it is not fixed yet.

This bug impacts us since we are using `numpy.genfromtxt` in a few places to read model files in `input/` and `static/` directory. For example: 
 - to read [`items.csv`](https://github.com/OasisLMF/OasisLMF/blob/7467b48d063cf579d53e2057d37ced7ccb755097/oasislmf/pytools/getmodel/manager.py#L182) (when `items.bin` is not present or `*.bin` are to be ignored)
 - to read [`vulnerability.csv`](https://github.com/OasisLMF/OasisLMF/blob/7467b48d063cf579d53e2057d37ced7ccb755097/oasislmf/pytools/getmodel/manager.py#L329) (when `vulnerability.bin` is not present or `*.bin` are to be ignored)
 - etc.

Furthermore, we have test model assets with simple 1-line `items.csv` files, etc. that are impacted by this bug and cause an error.

## Steps to Reproduce
Taking the example from https://github.com/numpy/numpy/issues/20451:
```py
from io import StringIO
import numpy as np

def printdata(inputdata):
    data = np.genfromtxt(StringIO(inputdata), names="a, b, c", usecols=("a", "c"))
    for i in data['a']:
        print(i)

# No issue here, column 'a' can be iterated
printdata("1 2 3\n4 5 6")

# Instead of an array, a single element is returned, resulting in a TypeError
printdata("1 2 3")
```
Error message:
```bash
  File "<input>", line 1, in <module>
TypeError: iteration over a 0-d array
```

A quick and effective fix is to replace replace `numpy.genfromtxt` with `numpy.loadtxt` with option `ndmin=1` to force it to always return a 1d arrays. This works. Changing the above example to use `numpy.loadtxt`:
```py
from io import StringIO
import numpy as np

def printdata(inputdata):
  data = np.loadtxt(
    StringIO(inputdata),
    dtype=np.dtype([('a', np.int32), ('b', np.int32), ('c', np.int32)]),
    ndmin=1)

  for i in data['a']:
      print(i)

# No issue here, column 'a' can be iterated
printdata("1 2 3\n4 5 6")

# No more issue here, column 'a' is still an iterable
printdata("1 2 3")
```
produces no errors:
```bash
1
4
1
```

Linux
All OasisLMF versions where `numpy.genfromtxt` is used.
NumPy/Python version information:
1.20.1 3.8.8 (default, Apr 13 2021, 19:58:26)
[GCC 7.3.0]

mtazzari:
This PR Fixes #1126. It improves `gulmc` by adding the functionality to treat aggregated vulnerability functions.
This PR also Fixes #1188 in commit [b9af01a](https://github.com/OasisLMF/OasisLMF/pull/1168/commits/b9af01ad7389c99fa2d78afcaa1f915b17ae3cf3).

### TODOs:

- [x] behaviour checks:
   - [x] when `weights.csv` does not contain weights for all the possible (areaperil_id, vulnerability_id) pairs, assume that the weight is zero for the missing entries.
  - [x] raise an Error if `weights.csv` or `.bin` is missing when `aggregate_vulnerability.csv` or `.bin` is present
  - [x] allow an empty `weights.csv` or `.bin` file to be provided for a non-empty `aggregate-vulnerability.csv`.
  - [x] raise an Error if all weights are zero for an aeraperil_id, aggregate_vulnerability_id (i.e., if the aggregate vuln cdf is identically zero).
- [x] data validation:
  - [x] what should be `gulmc` behaviour if `weights.csv` lacks entries for an `areaperil_id` or for a used `vulnerability_id`?
  - [x] what should be `gulmc` behaviour if `weights.csv` has duplicate entries?
  - [x] shall we implement a data validation step? 
  - [x] separate Issue to track the validation of these tables: #1190 
- [ ] reading/writing functionality for  `aggregate_vulnerability` and `weights` tables:
  - [x] implement csv reader for `aggregate_vulnerability.csv` and `weights.csv`;
  - [x] implement binary reader for `aggregate_vulnerability.bin` and `weights.bin`;
  - [ ] [probably not necessary for this PR to work; DEFERRED, TBC] implement executable for converting the two tables to/from csv/binary format (e.g., see [`disaggregationweighttobin` ](https://github.com/OasisLMF/ktools/tree/disaggregation_test/src/disaggregationweighttobin) in ktools).
    - [ ] if deferred, open an issue to track this. 
- [x] add cli argument for cache size in MB
- [ ] [not necessary for this PR to work; DEFERRED] `process_aggregate_vulnerability` and `process_aggregate_weights` functions in aggregate.py should be refactored to be jit-compiled. Performance will be necessary for large datasets.
  - [ ] open an issue to track this
- [ ] [IN PROGRESS] fix or remove all TODOs from the code
- [ ] [IN PROGRESS] update all docstrings, add comments where necessary
- [x] generate expected results for tests assets 2 to 6:
  - [x] generate all binaries for all the test model files, and correlations.bin using Python API. 
- [ ] [IN PROGRESS] finalize Relase Notes of this PR

Note: "IN PROGRESS" will be completed on 9-10-11 January 2022. 

### Note on test files
Most of the changed files are the new 700+ small binary files containing the expected test results for `test_model_1` to `test_model_6`. Note that all the expected results for 1200 `gulmc` tests only take 24 MB:
```bash
8.9M    ./test_model_1
9.0M    ./test_model_2
1.4M    ./test_model_3
1.5M    ./test_model_4
1.5M    ./test_model_5
1.4M    ./test_model_6
24M     .
```
All binary files can be inspected with `gultocsv`.

### Introducing  support for aggregate vulnerability definitions
With this PR we enable `gulmc` to support aggregate vulnerability functions, i.e., vulnerability functions that are composed of multiple individual vulnerability functions.
`gulmc` now can efficiently reconstruct the aggregate vulnerability functions on-the-fly and compute the aggregate (aka blended, aka weighted) vulnerability function. This new functionality works both in the "effective damageability" mode and in the full Monte Carlo mode.

Aggregate vulnerability functions are defined using two new tables, to be stored in the `static/` directory of the model data: `aggregate_vulnerability.csv` (or `.bin`) and `weights.csv` (or `.bin`). Example tables:
 - an `aggregate_vulnerability` table that defines 3 aggregate vulnerability functions, made of 2, 3, and 4 individual vulnerabilities, respectively:
```csv
aggregate_vulnerability_id,vulnerability_id
100001,1
100001,2
100002,3
100002,4
100002,5
100003,6
100003,7
100003,8
100003,9
```
 - a `weights` table that specifies weights for each of the individual vulnerability functions in all `areaperil_id`: 
```csv
areaperil_id,vulnerability_id,weight
54,1,138
54,2,224
54,3,194
54,4,264
54,5,390
54,6,107
[...]
154,1,1
154,2,97
154,3,273
154,4,296
[...]
```
items.csv (use only two aggregate vulnerability ids):
```csv
item_id,coverage_id,areaperil_id,vulnerability_id,group_id
1,1,154,8,833720067
2,1,54,2,833720067
3,2,154,8,956003481
4,2,54,100001,956003481
5,4,154,100002,2030714556
[...]
```
**Notes**:
 - if `aggregate_vulnerability.csv` or `.bin` is present, then `weights.csv` or `weights.bin` needs to be present too, or `gulmc` raises an error.
 - if `aggregate_vulnerability.csv` or `.bin` is not present, then `gulmc` runs normally, without any definition of aggregate vulnerability.

#### Caching
In order to speed up the calculation of losses in the full Monte Carlo mode, we implement a simple caching mechanism whereby the most commonly used vulnerability functions cdf are stored in memory for efficient re-usage. 
The cache size is set as the minimum between the cache size specified by the user with the new `--vuln-cache-size` argument (default: 200, units: MB) and the amount of memory needed to store all the vulnerability functions to be used in the calculations.
The cache dramatically speeds up the execution when the hazard intensity distribution is narrowly peaked (i.e., when most of the intensity falls in a few intensity bins), which implies a few vulnerability functions are used repeatedly.
The cache only stores individual vulnerability functions cdf, not the aggregate/weighted cdf, which would be too many to be stored.

Example: to allow the vulnerability cache size to grow up to 1000 MB can be done with:
```bash
eve 1 1 | gulmc -S100 -a1 --vuln-cache-size=1000
```

#### Testing
This PR:
 - introduces a suite of test model assets, from `test_model_2` to `test_model_6`, each with different properties that capture different ways of using aggregate vulnerabilities.
 - adds expected results for test_model_1 for zero samples (`-S0` parameter).
 - adds 720 further quantitative tests carried out against the test models. The total number of quantitative tests of `gulmc` that can be carried out with `pytest -v tests/pytools/gulmc` is now 1200.


mtazzari:
commit [9776b65](https://github.com/OasisLMF/OasisLMF/pull/1168/commits/9776b651fb1b9b9b40f403eb41045bcd0f6b6f72)
implements the weighted vulnerability functionality for the full monte carlo case (ie, for `effective_damageability=False`).
The algorithm implements a caching mechanism whereby the cdfs are stored in memory during the damage sampling: after sampling the hazard intensity and getting the hazard intensity bin, the corresponding vulnerability function is selected, and, if its cdf was not computed already, it is computed and stored it in a 1D dynamic array (its size doubles when the next cdf to be stored in it doesn't fit in the remaining free space), together with supporting arrays containing the pointers to that cdf.

If the vulnerability_id is one of the aggregate vulnerability_ids, the above algorithm is applied to each vulnerability function belonging to that aggregate vulnerability (namely, all the corresponding cdfs are computed and stored unless they were already computed), and all the resulting cdfs are sum-weighted using the aggregate weights definitions.

It is not clear, though, how to treat efficiently the computation of sidx=-1 and sidx=-2, namely the expected mean and std of the loss, which requires identical calculations of the effective_damageability=True case. In both cases, we need to compute the cdf of all vulnerability functions, for each intensity bin, so effectively we don't save anything by caching the cdfs.

johcarter (approved):

sambles (approved):

sstruzik:
The issue that Sam mentioned regarding the error with too many open file is solved here
https://github.com/OasisLMF/OasisLMF/pull/1212

PR #1174, opened 2022-12-08
Set pip-compile to backtracking and trim unused requirments

sambles:
### Fix CI bug with pip compile 
* There are several excess packages in `requirements.in` which are not needed for CI testing, these have been removed. 
* Set pip-compile to use backtracking when running with python3.7. 
* Tweaked the default in the manual trigger for `unittest.yml`,  it now won't build ktools unless a branch is given as input.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -60,13 +60,17 @@ jobs:
           python -m pip install --upgrade pip
           pip install pip-tools
 
+      - name: Set pip resolver flag
+        if: matrix.cfg.python-version == 3.7
+        run: |
+          echo "PIP_RESOLVER=--resolver=backtracking" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


.github/workflows/unittest.yml
```diff
@@ -19,7 +19,6 @@ on:
       ktools_branch:
         description: "Build Ktools before tests: [git ref]"
         required: false
-        default: "develop"
```
mtazzari, line 22: is this deletion on purpose?

mtazzari (approved):

.github/workflows/unittest.yml
sambles: yup, i missed that from the PR notes, i'll add it in. 

I noticed when running a manual trigger, ktools is always built because that value is auto-filled with `develop` when it should probably be blank as default.

PR #1175, opened 2022-12-08
Fix drop na in expected data

sambles:
### Fix expected data in test_data.py 
Calling `expected = df.dropna(subset=non_na_cols, axis=0)` is not equivalent to `get_dataframe(non_na_cols=non_na_cols, ..)` because **get_dataframe** overrides which strings pandas views as **NaN**.  

These need to be matched by replacing  strings matching patterns in `oasislmf.utils.data.PANDAS_DEFAULT_NULL_VALUES` with `np.nan`

### DIFF
--- a/tests/utils/test_data.py
+++ b/tests/utils/test_data.py
@@ -576,6 +576,7 @@ class TestGetDataframe(TestCase):
             data[-2]['str_col'] = np.nan
             df = pd.DataFrame(data)
             df.to_csv(path_or_buf=fp, columns=df.columns, encoding='utf-8', index=False)
+            df['str_col'] = df['str_col'].map(lambda x: np.nan if x in PANDAS_DEFAULT_NULL_VALUES else x)
             fp.close()
 
             non_na_cols = ['int_col', 'str_col']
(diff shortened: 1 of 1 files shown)


tests/utils/test_data.py
```diff
@@ -576,6 +576,7 @@ def test_get_dataframe__from_csv_file_with_nulls_in_some_columns__set_non_na_col
             data[-2]['str_col'] = np.nan
             df = pd.DataFrame(data)
             df.to_csv(path_or_buf=fp, columns=df.columns, encoding='utf-8', index=False)
+            df['str_col'] = df['str_col'].map(lambda x: np.nan if x in PANDAS_DEFAULT_NULL_VALUES else x)
```
mtazzari, line 579: Is there a reason why "`str_col`" here is lowercase?

mtazzari (approved):
all good, just one comment

tests/utils/test_data.py
sambles: I'm not sure why it was lower case in this example, i followed the test which was using `data[-2]['str_col'] = np.nan` instead of `data[-2]['STR_COL'] = np.nan` like the others

mtazzari: fair enough'

sstruzik (approved):

Issue #1177, opened 2022-12-09
The run_ktools.sh script does not check if all custom gulcalc processes completed successfully. 


JCrane512:
## Issue Description
The `check_complete()` function in `run_ktools.sh` does not check the log files of custom gulcalc processes. This is because it has no way currently to know what log messages to grep for. At JBA, sometimes our model runs out of RAM. When this happens the Out of Memory killer on Linux systems can kill one of our custom gulcalc processes. This sometimes results in an analysis that completes with only partial results and no error raised. 

To solve this issue I suggest adding two new settings to the file `oasislmf_settings.json` for the log messages when a gulcalc process starts/finishes. 

## Version / Environment information
Platform Linux
Effects all versions of Oasis

PR #1178, opened 2022-12-09
Update the run_ktools.sh script to include a check using the logs that all custom gulcalc process complete

JCrane512:
https://github.com/OasisLMF/OasisLMF/issues/1177

This MR is to add two new settings to the file `oasislmf.json` and to also enable these settings to be configured via the CLI. The two new settings are `model_custom_gulcalc_log_start` and `model_custom_gulcalc_log_finish`. They allow the user to specify the start and end log message strings used in a custom gulcalc process. 

These log messages are then used in the `check_complete()` function in `run_ktools.sh` to grep the log file `gul_stderror.err`. If the number of processes started and finished in the log is not the same an error is raised. 

Two new tests have been added. Each uses one of the two methods for generating a `run_ktools.sh` script, one using `genbash()` and the other `gen_chunked_bash()`. 

I have tested this change internally, and it works well with farmcat and catches the previous problem case where the Out of Memory killer on Linux kills one of our custom gulcalc processes and this is not reported by Oasis. 

### Adding new settings to oasislmf_settings.json
The two new settings are called `model_custom_gulcalc_log_start` and `model_custom_gulcalc_log_finish`, if you specify these two strings then `run_ktools.sh` will now check to ensure that the same number of custom gulcalc process finished that started.


sambles:
PR has moved to https://github.com/OasisLMF/OasisLMF/pull/1179  so it can be rebased on develop branch

PR #1179, opened 2022-12-09
Pr 1178 develop -- Update the run_ktools.sh script to include a check using the logs that all custom gulcalc process complete

sambles:
> Rebase of PR https://github.com/OasisLMF/OasisLMF/pull/1178 from @JCrane512 to fix issue https://github.com/OasisLMF/OasisLMF/issues/1177

### Update the run_ktools.sh script to use logs for checking custom gulcalc process complete
This MR is to add two new settings to the file `oasislmf.json` and to also enable these settings to be configured via the CLI. The two new settings are `model_custom_gulcalc_log_start` and `model_custom_gulcalc_log_finish`. They allow the user to specify the start and end log message strings used in a custom gulcalc process. 

These log messages are then used in the `check_complete()` function in `run_ktools.sh` to grep the log file `gul_stderror.err`. If the number of processes started and finished in the log is not the same an error is raised. 

Two new tests have been added. Each uses one of the two methods for generating a `run_ktools.sh` script, one using `genbash()` and the other `gen_chunked_bash()`. 

I have tested this change internally, and it works well with farmcat and catches the previous problem case 
where the Out of Memory killer on Linux kills one of our custom gulcalc processes and this is not reported by Oasis. 

#### Adding new settings to oasislmf_settings.json
The two new settings are called `model_custom_gulcalc_log_start` and `model_custom_gulcalc_log_finish`, if you specify these two strings then `run_ktools.sh` will now check to ensure that the same number of custom gulcalc process finished that started.


tests/model_execution/test_bash.py
```diff
@@ -926,6 +943,21 @@ def setUpClass(cls):
...
+    def _get_getmodel_cmd(**args):
+        return "custom_gulcalc_command"
+
+    def test_custom_gul_summarycalc_1_partition(self):
+        self.genbash("custom_gul_summarycalc_1_output", 1, _get_getmodel_cmd=self._get_getmodel_cmd)
+        self.check("custom_gul_summarycalc_1_output_1_partition")
+    
+    def test_custom_gul_summarycalc_1_partition_chunk(self):
+        self.gen_chunked_bash("custom_gul_summarycalc_1_output", 1, _get_getmodel_cmd=self._get_getmodel_cmd)
+        self.check_chunks("custom_gul_summarycalc_1_output_1_partition", 1)
```
carlfischerjba, line 959: I like this better than what I did. Adding the custom gulcalc tests to all test classes. Even if it only changes the outputs of the "err" tests, it's good to confirm that other outputs haven't changed.

oasislmf/execution/bash.py
```diff
@@ -1475,7 +1500,12 @@ def bash_params(
     else:
         bash_params['_get_getmodel_cmd'] = _get_getmodel_cmd
 
-    # Set fifo dirs
+    # Set custom gulcalc log statment checks, 
+        bash_params['custom_gulcalc_log_start'] = custom_gulcalc_log_start or analysis_settings.get('model_custom_gulcalc_log_start')
+        bash_params['custom_gulcalc_log_finish'] = custom_gulcalc_log_finish or analysis_settings.get('model_custom_gulcalc_log_finish')
```
carlfischerjba, line 1505: Indentation of comment.
```suggestion
        # Set custom gulcalc log statment checks, 
        bash_params['custom_gulcalc_log_start'] = custom_gulcalc_log_start or analysis_settings.get('model_custom_gulcalc_log_start')
        bash_params['custom_gulcalc_log_finish'] = custom_gulcalc_log_finish or analysis_settings.get('model_custom_gulcalc_log_finish')
```

oasislmf/execution/bash.py
```diff
@@ -1475,7 +1500,12 @@ def bash_params(
...
+
+
+    ## Set fifo dirs
```
carlfischerjba, line 1508:
```suggestion
    # Set fifo dirs
```

oasislmf/execution/bash.py
```diff
@@ -2285,7 +2327,13 @@ def genbash(
     # remove the file if it already exists
     if os.path.exists(filename):
         os.remove(filename)
-
-    with bash_wrapper(filename, bash_trace, stderr_guard):
+    
+    with bash_wrapper(  
+        filename, 
+        bash_trace, 
+        stderr_guard,
```
carlfischerjba, line 2334: Trailing whitespace here and a number of other places in this file. VSCode will let you trim trailing whitespace on save.

```suggestion

    with bash_wrapper(
        filename,
        bash_trace,
        stderr_guard,
```

tests/model_execution/test_bash.py
carlfischerjba: I hadn't spotted that these tests were only added to the "error guard" test class. That's probably fine for this piece of work but we could benefit from having tests for a custom gulcalc executable even without the error guard enabled. Should we add them to the base class instead?

With the current implementation, we've got `custom_gulcalc_log_start` and `custom_gulcalc_log_finish` added to the base test class but `_get_getmodel_cmd` only added to one subclass. Apparently it works but makes it unclear what we're testing.

sambles:
Thanks for the review @carlfischerjba  I've added in the recommended changes. 

> Is it worth handling the case where `log/gul_stderror.err` doesn't exist, perhaps a warning instead of crashing? The code is complex enough that it's hard to guarantee that the situation will never arise.

I'm not sure what to do here, a missing `log/gul_stderror.err` while the Error Guard is on feels like something has gone wrong in the execution, are there cases where this can happen where the output is valid?

carlfischerjba:
> Thanks for the review @carlfischerjba I've added in the recommended changes.
> 
> > Is it worth handling the case where `log/gul_stderror.err` doesn't exist, perhaps a warning instead of crashing? The code is complex enough that it's hard to guarantee that the situation will never arise.
> 
> I'm not sure what to do here, a missing `log/gul_stderror.err` while the Error Guard is on feels like an execution error, are there cases where this can happen but the output is valid?

In the whole codebase the only references to `gul_stderror.err` are when we check the contents, but I don't see the file being created anywhere. Doesn't this mean that the reference scripts are invalid and would crash if they were used, e.g. `tests/model_execution/err_kparse_reference/custom_gul_summarycalc_1_output_1_partition.sh`? There may be a `2>> log/gul_stderror.err` missing from the gulcalc lines.

To actually answer your question, it's too complicated for me to be sure but fine if you're happy with it. Just trying to avoid the case where there's a legitimate error causing the run to fail but then the real error gets masked by the fact that the log file doesn't exist but I agree that something would have to be badly messed up for that to happen.

sambles:
> In the whole codebase the only references to gul_stderror.err are when we check the contents, but I don't see the file being created anywhere. Doesn't this mean that the reference scripts are invalid and would crash if they were used, e.g. tests/model_execution/err_kparse_reference/custom_gul_summarycalc_1_output_1_partition.sh? There may be a 2>> log/gul_stderror.err missing from the gulcalc lines.

oh! i see what you mean, I missed that, yes that file should redirect into `log/gul_stderror.err`  but its instead using
```
 ( custom_gulcalc_command > fifo/gul_P1  ) 2>> $LOG_DIR/stderror.err &
```

Yes that looks wrong, we split the custom gul off to `gul_stderror.err` in case custom components to add some separation between the 'standard' ktools component errors.  I'll double check whats happening there before a merge

sambles:
@carlfischerjba   I think the issue was the **_get_getmodel_cmd(**kwargs)**  used in the tests, which only returns a flat `"custom_gulcalc_command"` irrespective of the kwargs. 

while the runner's default **_get_getmodel_cmd( .. )** checks if the error guard is set and adds an extra subshell to redirect any output to  log/gul_stderror.err

https://github.com/OasisLMF/OasisLMF/blob/PR-1178-develop/oasislmf/execution/runner.py#L78-L79

carlfischerjba (approved):
Looks good to me. 

@JCrane512 mentioned that this change might no longer be needed because the next PR is robust enough to catch all possible failures, but merging this wouldn't hurt either and might catch some corner cases we haven't considered.

PR #1189, opened 2022-12-21
Collect coverage in numba jit compiled functions

mtazzari:
This PR Fixes #  by introducing a specific test run to collect coverage information. In that run, numba jit compilation is disabled so that coverage can collect information on the lines hit by the tests. 

As documented here: https://community.codecov.com/t/numba-jitted-methods-are-not-captured-by-codecov/2649 coverage doesn't collect coverage from jit compiled functions, which results in a possibly considerable underestimate of our code coverage.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -36,6 +36,7 @@ jobs:
           - { python-version: "3.7", pkg-version: "", upload-cov: false }
           - { python-version: "3.8", pkg-version: "", upload-cov: false }
           - { python-version: "3.9", pkg-version: "", upload-cov: false }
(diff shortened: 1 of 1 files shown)


mtazzari:
@sambles I've added a conditional step to disable numba jit compilation for one of the tests, in order to be able to collect coverage in functions that normally are jit compiled.
Judging from the coverage uploaded to codecov for this branch, it seems that the functions are compiled anyway.
Now, numba compiles the functions the first time they're called.
Is it possible that the wheel that it is used to install oasislmf in the unittests contains compiled versions of the functions?

mtazzari:
It doesn't look like it's working. Won't go ahead with this.

PR #1192, opened 2023-01-09
Track pid of gulcalc processes and wait on them in `run_ktools.sh`

JCrane512:
When running a JBA analysis using Oasis, it is possible for a custom gulcalc process to run out of RAM and for a process to be killed by the OS. This is not always correctly reported and can result in an analysis completed with partial results with no error being raised. By explicitly waiting on these custom processes it is possible to use the existing `run_ktools.sh` error handling to also check these processes.

I previously discussed a solution with an implementation relating to `GNU parallel` with Oasis members. This is effectively the same thing but is a much cleaner solution. 

By using the existing `process_counter` variable, I have added it as a variable to the functions 'get_main_cmd_ri_stream()', 'get_main_cmd_il_stream()', and 'get_main_cmd_gul_stream()' in `bash.py`. 

I have personally tested that this solution works for the JBA model `farmcat` running out of RAM, but I have tested it on no other models.


JCrane512:
`PiWind MDK / PiWind / run_mdk (3.9)` is failing due to an unbound variable error. Unsure as to why for now, this did not come up with testing the JBA model.

JCrane512:
I failed to assign the value of the variable `main_cmd` in the function `get_main_cmd_ri_stream`. Which resulted in an unbound variable. 

`main_cmd = add_pid_to_shell_command(main_cmd, process_counter)`

JCrane512:
After fixing the previous error I had to update the unit tests. Some of the `run_ktools.sh` scripts have a double-blank line at the bottom. My vscode session was stripping one of them out when I was opening the files. Leading to failed tests.

JCrane512:
It is failing on the unit tests for CPython (3.8.15). A single test failure: 

`tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file__use_default_options`

I tested on `Python 3.8.10`

JCrane512:
The failing unit test `tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file__use_default_options` Is testing the function `get_dataframe()`.

I have made no changes to this function and so I am currently unsure why this test is failing.

JCrane512:
> The failing unit test `tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file__use_default_options` Is testing the function `get_dataframe()`.
> 
> I have made no changes to this function and so I am currently unsure why this test is failing.

The failure is due to "different" columns, and as far as I can tell they are the same. 
```
 DataFrame.iloc[:, 0] (column name="str_col") values are different (10.0 %)
E   [index]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
E   [left]:  [a, a, a, a, a, a, a, a, a, nan]
E   [right]: [a, a, a, a, a, a, a, a, a, nan]
```

sambles:
Hi @JCrane512   It looks similar to this https://github.com/OasisLMF/OasisLMF/pull/1175 

Does the develop branch on your fork have that commit?

JCrane512:
@sambles, thanks for your comment. I had a look at the commit you mentioned, I noticed that the same method used in that commit is not being applied to the test that is failing in this instance. So I have added it to the test. I suspect a similar problem with different types of `nan`.

JCrane512:
Looks like applying the change from https://github.com/OasisLMF/OasisLMF/pull/1175 into this unit test has fixed the issue. Not exactly sure why it passed before, but its fixed now!

JCrane512:
@sambles looks like everything is passing now. Thanks for your suggestion, I was able to fix the error. The other last few commits were just tidying up and resolving the merge conflicts.

oasislmf/execution/bash.py
```diff
@@ -1246,6 +1246,19 @@ def get_getmodel_cov_cmd(
...
     return cmd
 
+def add_pid_to_shell_command(cmd, process_counter):
+    """
+    Adds a variable onto the end of the cmd which tracks the process ID and increments the counter.
+    :param cmd: The command that you want to store the pid of in a variable.
+    :type cmd: string
+    :param process_counter: The number of process IDs that are being tracked.
+    :type process_counter: Counter()
+    """
```
mtazzari, line 1256: rewording of docstring in line with adopted style.

```suggestion
    """
    Add a variable to the end of a command in order to track the ID of the process executing it. 
    Each time this function is called, the counter `process_counter` is incremented.
    
    Args:
        cmd (str): the command whose process ID is to be stored in a variable.
        process_counter (Counter or dict): the number of process IDs that are being tracked.
    
    Returns:
        cmd (str): the updated command string.
    """
```

mtazzari (approved):

mtazzari:
This is merged now in develop, it'll go into the next release. Thank you @JCrane512 for the contribution!

PR #1209, opened 2023-01-23
use Policy CondClass to assess if full filter (exclusion) needs to be used

sstruzik:
### fix (FM) CondClass 1 as second priority condition
have CondClass exclusion work even if it is not the first priority condition


sambles (approved):
Change to using `PolCondClass` is more for @johcarter

 Only one comment on code changes, it might be worth using the oed-tool's 
 `OedExposure(location=Path('<path>')) `  when combining exposure in [validation/combine.py](https://github.com/OasisLMF/OasisLMF/pull/1209/files#diff-13933ca90f28c1c26a1c54c27253db8afa9644283b956d84a932c19284b02981)
 
Its not a big deal, but it might be safer in the long run.

johcarter (approved):

sambles:
Merge conflicts when cherry-picking to `1.26`,  needs adjusting for that backport

PR #1212, opened 2023-01-30
use io.open for footprint bin instead of os.open to take advantage of…

sstruzik:
… automatic closure

### fix Footprint open file leakage in `gulmc` and `getmodel`
Closure of `footprint.bin` and `footprint.bin.z` file was not handled properly in `oasislmf/pytools/getmodel/footprint.py`.
This code is used in both `gulmc` and `getmodel` and was keeping the file open until the process finished.


sambles (approved):
Fixes the OS file error when running https://github.com/OasisLMF/OasisLMF/blob/develop/tests/pytools/gulmc/test_gulmc.py

mtazzari (approved):
Good catch at fixing this!

PR #1218, opened 2023-02-01
Fix missing default collumn for RI

sstruzik:
Add certain column with default value when loading RI files as they are needed later on in the RI preparation.

### DIFF
--- a/oasislmf/utils/data.py
+++ b/oasislmf/utils/data.py
@@ -26,7 +26,9 @@ __all__ = [
     'PANDAS_BASIC_DTYPES',
     'PANDAS_DEFAULT_NULL_VALUES',
     'reduce_df',
-    'set_dataframe_column_dtypes'
+    'set_dataframe_column_dtypes',
+    'RI_SCOPE_DEFAULTS',
+    'RI_INFO_DEFAULTS'
 ]
 
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


oasislmf/utils/data.py
```diff
@@ -109,6 +111,29 @@
 # Load schema json dir
 SCHEMA_DATA_FP = os.path.join(os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir)), 'schema')
 
+RI_INFO_DEFAULTS = {
```
benhayes21, line 114: Should the defaults not come from the OED spec, rather than be hard-coded in oasislmf?

oasislmf/utils/data.py
```diff
@@ -109,6 +111,29 @@
...
+    'CededPercent': 1.0,
+    'RiskLimit': 0.0,
+    'RiskAttachment': 0.0,
+    'OccLimit': 0.0,
+    'OccAttachment': 0.0,
+    'TreatyShare': 1.0
+}
+
+RI_SCOPE_DEFAULTS = {
```
benhayes21, line 123: As above - defaults feel like they should be inherited?

benhayes21 (commented):
default values for RI fields are specified in utils/data. Can these not be taken form the OED spec in ODS tools?

benhayes21 (approved):

sambles:
> Added chat discussion for keeping track

**Stephane:** We expect it for our computation but they are not required in the schema.
that's why I need to add them if missing.

**Ben Hayes:** actually no - they shouldn't be required really, so not right to set them in the spec
not sure why our computation needs them, but it's fine I think

PR #1220, opened 2023-02-02
Fix mapping of `vuln_idx` to `vuln_i` and implement `eff_vuln_cdf` as a dynamic array to work with large footprints

mtazzari:
This PR fixes a bug in `gulmc` which was causing `gulmc` not to work when vulnerability functions are missing so that their numeration is non-contiguous, e.g., `3, 5, 200`. This PR also implements a dynamic array storage for effective vulnerability cdf which allows processing large footprint files.

### Bugfix in `gulmc` to deal with vulnerability tables with missing ids and with large fooprint files.
This PR fixes a bug in `gulmc` which was causing `gulmc` not to work when vulnerability functions are missing so that their numeration is non-contiguous, e.g., `3, 5, 200`. This PR also implements a dynamic array storage for effective vulnerability cdf which allows processing large footprint files.


sambles (approved):
Unable to reproduce the original bug with PiWind
```
gulmc doesn't work properly when vulnerability functions are numbered in a non-contiguous way, e.g., 3, 5, 200.
```
gulmc from (1.27.0) doesn't seem to fail running piwind with [vulnerability.csv](https://github.com/OasisLMF/OasisLMF/files/10697102/vulnerability.csv)
which is both non-contiguous, and has missing vulnerability_id's. 

not failing is strange and worth looking into, however the logic change looks good and results are consistent

sstruzik (approved):

benhayes21 (approved):
tested with the model that the problem was showing in, and it all works now

PR #1224, opened 2023-02-08
Backport 1.26 for PR-1209 

sambles:
### fix (FM) CondClass 1 as second priority condition
have CondClass exclusion work even if it is not the first priority condition

> **Backport for** https://github.com/OasisLMF/OasisLMF/pull/1209


sstruzik:
to delete see https://github.com/OasisLMF/OasisLMF/pull/1225

PR #1236, opened 2023-03-13
Switch changelog builder from "build" repo to "OasisPlatform"

sambles:
### Switch changelog builder from "build" repo to "OasisPlatform"
The repository [OasisLMF/build](https://github.com/OasisLMF/build) is planned for removal, switch the Changelog script to run from [OasisPlatform/develop/scripts/update-changelog.py](https://github.com/OasisLMF/OasisPlatform/blob/develop/scripts/update-changelog.py) instead of build

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -55,14 +55,12 @@ jobs:
     secrets: inherit
     with:
       oasislmf_version: ${{ inputs.release_tag }}
-      ktools_version: "" # !! TESTING ONLY !!
 
   build:
     uses: ./.github/workflows/build.yml
     secrets: inherit
     needs: update
     with:
-      ktools_branch: "" # !! TESTING ONLY !!
       build_osx: ${{ inputs.build_osx }}
 
   release:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,16 +41,8 @@ on:
         required: false
         default: 'false'
 
-      # TESTING ONLY
-      # This is Risky and should be exposed by default --> Run ./.github/workflows/version.yml separately and wait for testing workflows to pass before publish
-      #ktools_version_set:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -196,8 +188,8 @@ jobs:
     - name: Create Release notes
       working-directory: ${{ env.WORKSPACE }}
       run: |
-        ${{ github.workspace }}/auto_changelog.py build-release \
-          --repo OasisLMF \
+        ${{ github.workspace }}/update-changelog.py build-release \
+          --repo ${{ github.event.repository.name }} \
```
mtazzari, line 192: by dynamically specifying `--repo` instead of `--repo OasisLMF` does it mean that we can build changelogs for different repos for which this workflow is triggered?

mtazzari (approved):
All fine to me.

benhayes21 (approved):

.github/workflows/publish.yml
```diff
@@ -41,16 +41,8 @@ on:
...
-      #  description: '(Optional) Pin a new ktools version, semvar (without the "v") For ktools release v3.0.0 input [3.0.0]'
-      #  required: false
-      #  default: ''
-
 
 env:
-  BUILD_SCRIPTS: "master"    # build scripts branch
   WORKSPACE: ${{ github.workspace }}/OasisLMF
   RELEASE_TAG: ${{ inputs.release_tag }}
   PREV_RELEASE_TAG: ${{ inputs.prev_release_tag }}
```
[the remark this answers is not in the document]
sambles: yup, these should also go, I'll remove them

.github/workflows/publish.yml
sambles: This `publish.yml` file is specific to oasislmf, so in practice `github.event.repository.name` will always have the value `OasisLMF`. 

But overall it makes the script a little more robust, (in case the repo name ever needs to change.. etc)

Issue #1259, opened 2023-04-18
Duplicate summary_ids in outputs


sambles:
## Issue Description
When loading the summary map file if the selected field, for example `accnumber`, has a mix of numbers and strings in the CSV

pandas can incorrectly infer the dtypes as a mix of strings and int due to how `get_dataframe` loads the summary map file (without dtypes set) 

```
In [x]: df_bug = pd.read_csv('BUG_gul_S1_summary-info.csv')
In [x]: df_bug[df_bug.accnumber.duplicated(keep=False)]
Out[x]: 
     summary_id accnumber           tiv
351         352         5  2.354872e+11
352         353         5  5.776238e+10
353         354        84  1.642087e+13
354         355        84  2.333112e+10
378         379        23  2.487472e+10
379         380        23  5.593756e+09
```

```
ipdb> summary_desc_df[summary_desc_df.accnumber.duplicated(keep=False)]
Empty DataFrame
Columns: [accnumber]
Index: []

ipdb> summary_desc_df.accnumber.loc[351]
'5'
ipdb> summary_desc_df.accnumber.loc[352]
5
ipdb> type(summary_desc_df.accnumber.loc[352])
<class 'int'>
ipdb> type(summary_desc_df.accnumber.loc[351])
<class 'str'>
```

## Fix 
* Pass explicit dtypes when loading the summary map file (1.27.x) 
* Use the option `low_memory=False` when loading the dataframe, when true the df could be a mix of types
```
Generating Oasis files (GUL=True, IL=True, RIL=False)
/root/.local/lib/python3.10/site-packages/oasislmf/utils/data.py:551: DtypeWarning: Columns (1) have mixed types. Specify dtype option on import or set low_memory=False.
  headers = list(pd.read_csv(src_fp, encoding=use_encoding).head(0))
```

## Affected Versions (LTS) 
* 1.15.26 (and all versions below) 
* 1.23.16 (and all versions below)    
* 1.26.6 (and all versions below)  

> Note: This is not a problem in `1.27.2` but a fix is still required for safer loading of data

PR #1260, opened 2023-04-18
Use low_memory=False in get_dataframe

sambles:
### Fix issue where Datafame is loaded with mixed types 
* Loading a dataframe in 'chunks' can cause a column to have mixed dtypes, causing #1259

### DIFF
--- a/oasislmf/utils/data.py
+++ b/oasislmf/utils/data.py
@@ -415,7 +415,7 @@ def get_dataframe(
                 # Find flexible fields in loc file and set their data types to that of
                 # FlexiLocZZZ
                 if 'FlexiLocZZZ' in col_dtypes.keys():
-                    headers = list(pd.read_csv(src_fp, encoding=use_encoding).head(0), low_memory=low_memory)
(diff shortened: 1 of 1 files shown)


oasislmf/utils/data.py
```diff
@@ -409,13 +415,14 @@ def get_dataframe(
                 # Find flexible fields in loc file and set their data types to that of
                 # FlexiLocZZZ
                 if 'FlexiLocZZZ' in col_dtypes.keys():
-                    headers = list(pd.read_csv(src_fp, encoding=use_encoding).head(0))
+                    headers = list(pd.read_csv(src_fp, encoding=use_encoding).head(0), low_memory=low_memory)
```
sstruzik, line 418: I guess you saw in the error but you put the arg in the wrong place
headers = list(pd.read_csv(src_fp, encoding=use_encoding, low_memory=low_memory).head(0))

sambles: oupf! thanks for the spot

sstruzik (approved):

sambles:
Overriding missing review, `@` me @benhayes21 if something needs fixing

PR #1262, opened 2023-04-24
specify dtype for useful summary column reading

sstruzik:
### specify dtype for useful summary column reading
During the loss computation preparation step, we read the summaries files using a simple pandas.read_csv.
This change provide the read with the dtype of each needed columns improving read speed and avoiding potential wrong data type issues.
We also remove the case lowering of the column name as the column are expected to have the proper case in the files and we need them to match properly with the other files.


mtazzari (approved):
All ok for me. The only issue I see is that we should infer the data types instead of defining them once again in `MAP_SUMMARY_DTYPES`.
Data types for the same fields are defined multiple times in the whole code base. This PR makes the issue worse, in that respect.
If there's a quick way to infer the datatypes instead of re-defining them it would be great.

sambles (approved):

PR #1265, opened 2023-05-02
add combine function to lookup built in

sstruzik:
### add combine functionality to lookup builtin function 
add a function that combine several strategies trying to achieve the same purpose by different mean into one.
for example, finding the correct area_peril_id for a location with one method using (latitude, longitude) and one using postcode.
each strategy will be applied sequentially on the location that steal have OASIS_UNKNOWN_ID in their id_columns after the precedent strategy

### DIFF
--- a/oasislmf/lookup/builtin.py
+++ b/oasislmf/lookup/builtin.py
@@ -198,8 +198,8 @@ class Lookup(AbstractBasicKeyLookup, MultiprocLookupMixin):
 
     def set_step_function(self, step_name, step_config, function_being_set=None):
         """
-        set the strategy as a function of the lookup object if it's not already done and return it.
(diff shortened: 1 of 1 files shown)


sstruzik:
[lookup_config.zip](https://github.com/OasisLMF/OasisLMF/files/11374543/lookup_config.zip)
example of lookup with two strategy to determine area_peril_id using lat long  or using postcode

johcarter (approved):
Works well and implemented in https://github.com/OasisLMF/OasisModels/tree/piwind-postcode, tested in test_6

sambles (approved):

sambles:
**Note:** rebased with develop to fix test failure from ods-tools update

PR #1277, opened 2023-05-11
Fix invalid columns in FM tests

sambles:
### Fix invalid columns in FM tests
* Test failures raised with https://github.com/OasisLMF/ODS_Tools/pull/29 `Pol6AllDed` -> `PolDed6All`


sstruzik (approved):

johcarter (approved):
Good find. Doesnt change results because there are no policy deductibles applied in these test cases

PR #1280, opened 2023-05-11
Drop py3.7 in testing and add py3.11

sambles:
### Update unit test python versions 
Numba 0.57 has moved its minimum python version to 3.8, update testing to track for and replace with py3.11

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -46,10 +46,10 @@ jobs:
     strategy:
       matrix:
         cfg:
-          - { python-version: "3.7", pkg-version: "", upload-cov: false }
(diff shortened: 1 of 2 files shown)


mtazzari (changes requested):
Great news that Python 3.11 is now supported by numba. 
I've also noticed they dropped support for Python 3.7.
This PR is all good, but we should fix the main README of the repo, and change
```
Also for this release (and all future releases) a **minimum of Python 3.7 is required**.
```
to
```
Also for this release (and all future releases) a **minimum of Python 3.8 is required**.
```
Can you fix that?

sstruzik (approved):

mtazzari (approved):

PR #1286, opened 2023-05-19
Improve code coverage accuracy

sambles:
### Improve code coverage accuracy
* Run the CI unit test workflow by installing the oasislmf package using editable mode. This is so code coverage will pickup execution checks for pytools (for example [test_fm.py](https://github.com/OasisLMF/OasisLMF/blob/develop/tests/fm/test_fm.py) and  [test_fmpy.py](https://github.com/OasisLMF/OasisLMF/blob/develop/tests/fm/test_fmpy.py)) 
* Fixed JIT functions not picked up in code coverage
* Removed the [utils/forex.py](https://github.com/OasisLMF/OasisLMF/pull/1286/commits/7de6b852d68b12de0f278b2b03d39b79ea43eb8c#diff-fdccee93025f26c36aab2dfbbbb281aa0a3485132eace922ef835026d20bc856) file as the same functionality has been moved over to the ods-tools package. 
* Added `--gul-rtol` and `--gul-atol` override options as testing flags, this is needed for a larger difference in expected tolerance between running with and without JIT.


mtazzari (approved):
Great job Sam, it's really great to now extend coverage to the jit-compiled parts of the code, which are a sizeable fraction of the codebase.
As discussed, it's far from ideal and a bit puzzling that we have to increase relative and absolute tolerance for tests to 1e-3 when disabling numba, but there is not much we can do about it.
Hopefully in the future we will be able to restrict tolerance again. 
For now, this is good to go.

sstruzik (approved):

PR #1289, opened 2023-05-22
Fix/fillna on str

sstruzik:
### support setting default value for all kinds of blank value
replace fillna by ods_tools fill_empty in order to set default to all kind of blank values including empty string.


sambles (approved):

mtazzari (approved):
All good for me.

sambles:
**Note:** Don't forget to pin ods-tools to `3.2.0` on next release

PR #1290, opened 2023-05-22
Added platform client testing

sambles:
### Add platform client testing
* Added checks for the Platform API session manager 
* Added check for the Platform API client
* Added fixes and tidy up to API client and session manager code 

> **Follow up note:** clean up exception handling, for example `session.py` should either raise the original exception which is then wrapped in `client.py` or directly wrap all unhandled exceptions to `OasisException`


mtazzari (approved):
All fine for me. You may want to link an Issue to this PR.

sstruzik (approved):

PR #1291, opened 2023-05-24
remove obsolete fm compute

sstruzik:
remove obsolete fm compute file


sambles (approved):

mtazzari (approved):
Fine with that. For future reference, can you confirm whether the functionality in `compute.py` has been dropped, or moved to some other module(s)?

sambles:
> Fine with that. For future reference, can you confirm whether the functionality in `compute.py` has been dropped, or moved to some other module(s)?

Yup, its was replaced by `oasislmf/pytools/fm/compute_sparse.py`

PR #1294, opened 2023-05-25
Removed obsolete Cookiecutter from oasislmf 

sambles:
### Removed Cookiecutter from oasislmf 
The template Cookiecutter repos have been removed, so the code in oasislmf is now redundant. 
Removed the deadcode


sstruzik (approved):

mtazzari (approved):
Looks good to me.

PR #1299, opened 2023-06-05
Support OED Peril terms and coverage specific terms for all levels

sstruzik:
### Support OED Peril terms and coverage specific terms for all level
-  support OED Peril terms (adding a filter so only the loss from correct perils are part of the policy)
-  full revamp of fm file generation step in order to preserve memory.
-  support coverage specific term for condition
-  have the condition logic able to handle graph structure (not just tree structure)

Also, to be able to run our tests using exposure run, Peril need to be taken from LocPerilCovered
 in exposure run add option to use LocPerilCovered for peril id and use only certain peril
During an exposure run, the perils used were determine base on num_subperils and their id were 1 to num_subperils
With this change user can specify the peril covered by the deterministic model via --model-perils-covered
if nothing is given all peril in LocPerilCovered will be attributed a key and will receive a loss from the model.

it is also now possible to specify extra summary column so they can be seen in the loss summary at the end of exposure run using --extra-summary-cols 

example:
```
oasislmf exposure run -s ~/test/peril_test -r ~/OasisLMF/runs/peril_test --extra-summary-cols peril_id --model-perils-covered WTC
```


oasislmf/computation/run/exposure.py
johcarter, line 1: I would keep this as 'item', because item is already peril specific

sstruzik: solved, I've kept item in exposure run and added the possibility to add certain extra columns in the summary with --extra-summary-cols peril_id

sambles (approved):

benhayes21 (approved):
Tested peril terms, cond coverage terms and peril cond coverage terms. All looks good to me

PR #1306, opened 2023-06-12
lookup: remove extra column after failed combine step

sstruzik:
### lookup: remove extra column after failed combine step
In builtin lookup, for locations where a combine step is not valid, some extra column could be added on top of the id one.
With this change only the original column are passed to the next combine step


sambles:
Overriding min reviewer limit to include in patched release

PR #1321, opened 2023-06-26
Cleanup/fix fm validation cr

sstruzik:
### make validation oed file comply with oed conditional requirement
Add conditionally required column to oed validation files


sambles (approved):
Tested with 3.0.4rc2 and https://github.com/OasisLMF/ODS_Tools/pull/48

PR #1328, opened 2023-07-12
Implement Post Loss Amplification

hchagani-oasislmf:
### Implement Post Loss Amplification
Major events can give rise to inflated costs as a result of the shortage of labour, materials and other factors. Conversely, in some cases the costs may be lower as the main expenses may be shared among the sites that are hit in the same area. To account for this, the ground up losses from `gulpy` are multiplied by post loss amplifications by a new component `plapy`.

`plapy` requires the files `static/lossfactors.bin` and `input/amplifications.bin` to assign loss factors to event ID-item ID pairs from `gulpy`. More details on these files are available in the `ktools` PR https://github.com/OasisLMF/ktools/pull/351. Losses are then multiplied by their corresponding factors. Loss factors that are not found in the loss factors file are assumed to be 1. The output is identical to that of `gulpy`: event ID, item ID, sample ID (sidx); and loss.

The file `static/lossfactors.bin` is supplied by the model provider, and maps event ID-amplification ID pairs to loss factors. The file `input/amplifications.bin` is generated from the keys file. A strategy to assign amplification IDs to fields in the source locations file can be supplied by the model provider. If present, amplification IDs are assigned to keys. If there are no amplification IDs,

If `input/amplifications.bin` is present, the Ground Up Loss (GUL) output is piped through `plapy`.


benhayes21 (approved):

oasislmf/preparation/gul_inputs.py
```diff
@@ -537,6 +570,12 @@ def write_gul_input_files(
         if oasis_files_prefixes.get('complex_items'):
             oasis_files_prefixes.pop('complex_items')
 
+    # If no amplification IDs then remove corresponding file name from files
+    # prefixes dict
+    if 'amplification_id' not in gul_inputs_df:
+        if oasis_files_prefixes.get('amplifications'):
```
sstruzik, line 576: if 'amplification_id' not in gul_inputs_df:
    oasis_files_prefixes.pop('amplifications', None)
will pop the key if present

oasislmf/pytools/pla/structure.py
```diff
@@ -0,0 +1,158 @@
...
+            N_PAIRS, buffer=factors_buffer, dtype=amp_factor_dtype
+        )
+        f.readinto1(factors_buffer[:FILE_HEADER_SIZE])   # Ignore first 4 bytes
+
+        cursor = 0
+        valid_buffer = 0
+        count = 0
+        event_id = 0
+        while True:
+            len_read = f.readinto1(factors_buffer[valid_buffer:])
```
sstruzik, line 142: I think you need readinto here because readinto1 can return less that ask even if the file has more data.
then it would error if it cut in the middle of a line.
readinto read the number of byte asked unless the stream ends
readinto1 read from 1 up to the number of byte asked unless the stream ends (then it can return 0)

sstruzik (approved):

PR #1342, opened 2023-07-19
Redefine key_columns as local variable

hchagani-oasislmf:
### Redefine key_columns as local variable
Making changes the global variable `key_columns`, which is a list of location file columns used in the lookup process, can lead to errors. As the variable is only used in the method `builtin.py::Lookup::process_locations`, it can be defined local to that method instead.

### DIFF
--- a/.github/workflows/piwind-test.yml
+++ b/.github/workflows/piwind-test.yml
@@ -24,7 +24,11 @@ on:
 
   workflow_dispatch:
     inputs:
-      platform_version:
+      platform_1_version:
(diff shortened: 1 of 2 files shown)


sambles:
CI is checking vs the `latest` docker image tag, which is platform 1 and why this got missed on release. 
Running same workflow here using version `2.2.0` https://github.com/OasisLMF/OasisLMF/actions/runs/5597713137

sambles (approved):

sstruzik (approved):

PR #1345, opened 2023-07-20
put complex_model_config into lookup_config

sstruzik:
### Add complex model config into model config if both present
If both complex model config and  model config are present, add the json dict from the complex config into the model config
as below
                config['complex_config_dir'] = complex_config_dir
                config['complex_config'] = complex_config


sambles (approved):
Merging in this fix, with the aim of renaming `complex_lookup_config_fp` to `analysis_settings` in a follow up PR

PR #1346, opened 2023-07-21
Revamp fm test

sstruzik:
### Update all fm test to use AA1 as peril in all peril columns
Work is in progress to have perils columns such as LocPerilsCovered, LocPeril, ... supported in oasislmf. This change aim at changing all perils to AA1 as they represent generic test. some more test specific to peril covered will be added later on with the feature.

also improve the split combine scripts used to add fm unit test by adding support for reinsurance files


sambles (commented):
I'm not sure how relevant this comment is, but there are still lots of files with `OO1` under **oasisLMF/validation** 

Is the goal to update some tests where needed, or every file in the test suite?

[list_of_files_matching_OO1.txt](https://github.com/OasisLMF/OasisLMF/files/12193681/list_of_files_matching_OO1.txt)
[grep_files_for_OO1.txt](https://github.com/OasisLMF/OasisLMF/files/12193683/grep_files_for_OO1.txt)

sstruzik:
> I'm not sure how relevant this comment is, but there are still lots of files with `OO1` under **oasisLMF/validation**
> 
> Is the goal to update some tests where needed, or every file in the test suite?
> 
> [list_of_files_matching_OO1.txt](https://github.com/OasisLMF/OasisLMF/files/12193681/list_of_files_matching_OO1.txt) [grep_files_for_OO1.txt](https://github.com/OasisLMF/OasisLMF/files/12193683/grep_files_for_OO1.txt)

good catch
I've updated the files complete file in expected.

I haven't touch certain unit test, the unit test in todo, to be done if necessary when intergrated.
the expected for unit test, I think they should be removed

johcarter:
Yes I agree the expected folders underneath individual units should be removed. 

I think we could remove the todo units under reinsurance1, as the existing tests have good coverage.

sstruzik:
> Do you think that the OEDVersion in the test files should be updated to v3? Since we are using ODS tools v3.x?

It's not really used at the moment so we could but it will go out of sync as soon as we release a new version of ODS.
With the current implementation I don't see the added value

sambles (approved):
There are still references to the Peril code `OO1` in the the sub unit's expected dirs, but as Joh mentioned, should probably just be deleted.

johcarter (approved):

Issue #1322, opened 2023-06-28
Step policies: Allow BI ground up loss through to gross losses


johcarter:
## Issue Description

In step policies, business interruption total insured value 'BITIV' is not carried through the financial calculations. If there is a value, and if it is modelled as in independent coverage under a step policy, ground up losses may be generated but gross loss will be zero.

This is because additional financial losses in step polices (debris removal, extra expenses) are defined as uplift factors applied to the main structure loss, or to a combination of the main structure and contents loss.

However I suggest that if there is a model that models BI as an independent coverage and uses step policy fields to model gross losses, then BI ground up loss should be carried through to gross losses and not zeroed out.

Technically, this involves applying a pass through calcrule (100) to any ground up loss of coverage type 4 in the FM for step policies, rather than applying a 0 limit calcrule.

Raph692:
Hi Joh,

Just FYI, I can confirm that we are also seeing that BI is set to zero when using a step function (even if setting a regular deductible & limit) - see screenshot attached. Are you aware of any workarounds in the meantime whilst your above solution is implemented?

<img width="312" alt="image" src="https://github.com/OasisLMF/OasisLMF/assets/72822393/b947cfc9-255b-4420-ac85-79883566e9b9">

johcarter:
You could split each location into 2 locations, one for Cov1-3 and one for Cov 4, then include locations for Cov1-3 in Account 1 and use a step policy, and include locations for Cov4 in Account 2 and use regular policy terms.  Then use OED fields to group them together for reporting, i.e. for summing gross coverages losses across the accounts  . "AccGroup", for instance.

PR #1351, opened 2023-07-28
let BI coverage pass through in step policy

sstruzik:
### Step policies: Allow BI ground up loss through to gross losses
https://github.com/OasisLMF/OasisLMF/issues/1322


johcarter (approved):
works as expected

sambles (approved):

PR #1352, opened 2023-08-03
Support multiple identifiers for footprint files

hchagani-oasislmf:
### Support multiple identifiers for footprint files
To enable the storage of footprints in multiple files rather than a single master file, optional identifiers in the form of footprint file suffixes are now supported. This is executed in a similar way to that currently in place to distinguish multiple events and event occurrences files. The `footprint_set` model settings option in the analysis settings file can be set to the desired file suffix for the footprint files to be used. A symbolic link to the desired footprint set is created in the `static/` directory within the model run directory. Footprint file priorities are identical to those set by `modelpy` and `gulmc`, which in order of descending priority are: parquet; zipped binary; binary; and csv.

### DIFF
--- a/oasislmf/execution/bin.py
+++ b/oasislmf/execution/bin.py
@@ -39,9 +39,9 @@ from ..utils.defaults import STATIC_DATA_FP
 from .files import TAR_FILE, INPUT_FILES, GUL_INPUT_FILES, IL_INPUT_FILES
 from .bash import leccalc_enabled, ord_enabled, ORD_LECCALC
 from oasislmf.pytools.getmodel.common import footprint_file_formats
-from oasislmf.pytools.getmodel.footprint import (
-    FootprintParquet, FootprintBinZ, FootprintBin, FootprintCsv
-)
+from oasislmf.pytools.getmodel.footprint import (   # noqa: F401
+    FootprintParquet, FootprintBinZ, FootprintBin, FootprintCsv   # noqa: F401
+)   # noqa: F401
 
 logger = logging.getLogger(__name__)
 
--- a/oasislmf/execution/bin.py
+++ b/oasislmf/execution/bin.py
@@ -11,7 +11,8 @@ __all__ = [
     'create_binary_tar_file',
     'csv_to_bin',
     'prepare_run_directory',
-    'prepare_run_inputs'
+    'prepare_run_inputs',
+    'set_footprint_set'
 ]
 
 
@@ -37,6 +38,12 @@ from ..utils.log import oasis_log
 from ..utils.defaults import STATIC_DATA_FP
 from .files import TAR_FILE, INPUT_FILES, GUL_INPUT_FILES, IL_INPUT_FILES
 from .bash import leccalc_enabled, ord_enabled, ORD_LECCALC
+from oasislmf.pytools.getmodel.common import footprint_file_formats
+from oasislmf.pytools.getmodel.footprint import (
+    FootprintParquet, FootprintBinZ, FootprintBin, FootprintCsv
+)
+
+logger = logging.getLogger(__name__)
 
 
 @oasis_log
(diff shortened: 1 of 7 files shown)


oasislmf/execution/bin.py
```diff
@@ -392,6 +399,44 @@ def prepare_run_inputs(analysis_settings, run_dir, ri=False):
...
+    :type setting_val: string
+
+    :param run_dir: model run directory
+    :type run_dir: string
+    """
+    format_to_class = {
+        'parquet': FootprintParquet, 'binZ': FootprintBinZ,
+        'bin': FootprintBin, 'csv': FootprintCsv
+    }
+    priorities = [format_to_class[fmt] for fmt in fp_format_priorities if fmt in format_to_class]
```
mtazzari, line 418: This bit is already defined in oasislmf/pytools/getmodel/footprint.py
I'd say we should re-use the definition and avoid repetition.

oasislmf/execution/bin.py
```diff
@@ -392,6 +399,44 @@ def prepare_run_inputs(analysis_settings, run_dir, ri=False):
...
+def set_footprint_set(setting_val, run_dir):
+    """
+    Create symbolic link to footprint file set that will be used for output
+    calculation.
+
+    :param setting_val: identifier for footprint set
+    :type setting_val: string
+
+    :param run_dir: model run directory
+    :type run_dir: string
```
mtazzari, line 412: Minor: args and returns don't follow usual format with Args:... and Returns:...

mtazzari (approved):
All good, just one minor comment to improve maintainability.

sambles (approved):

oasislmf/execution/bin.py
hchagani-oasislmf: I agree that it would be better to avoid repetition. However, as the footprint format classes are defined in `oasislmf/pytools/getmodel/footprint.py`, we can't define `format_to_class` as a global variable before the declaration of these classes. I'm open to suggestions of how to make it so we only have to define it once.

oasislmf/execution/bin.py
hchagani-oasislmf: I followed the style that was already used in other places in the file. I notice that there are a few other files that follow the same style. Perhaps we should create a new issue to standardise all these files?

oasislmf/execution/bin.py
mtazzari: I'd then move def of `format_to_class` in oasislmf/pytools/getmodel/footprint.py after the classes are defined.

oasislmf/execution/bin.py
mtazzari: Yes, there is confusion all over the place. Fine if you want to keep them like they are.
Anyway, some time ago we introduced [CONTRIBUTING.md](https://github.com/OasisLMF/OasisLMF/blob/main/CONTRIBUTING.md) guidelines where we decided to follow Google-style docstrings, see the referred [Chapter 3.8](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).

Please go ahead and open an issue to standardise all the docstrings, I'm in favour of that.

oasislmf/execution/bin.py
hchagani-oasislmf: We would also need to move the `Footprint` class to the end of the file too then.

mtazzari: Well, actually I've just seen that `priorities` is already defined in the `load` method of `Footprint`. I think the code would benefit from a small refactoring, defining a priorities list and `format_to_class` mapping once in the `footprint.py` file, and using it in `Footprint` and everywhere else, including in `execution/bin.py`. For the sake of closing this PR quickly, probably it makes sense to leave it as it is, and remember/add an issue to refactor the footprint priorities. 
Right now we hard-code the priorities order and mapping three times in the code, which is bad for maintainability.

hchagani-oasislmf: Please see issue https://github.com/OasisLMF/OasisLMF/issues/1355.

oasislmf/execution/bin.py
hchagani-oasislmf: Please see issue https://github.com/OasisLMF/OasisLMF/issues/1356.

Issue #1317, opened 2023-06-21
Add post-analysis hook


philipoldham:
## Issue Description
We wish to have access to a post-processing hook. The post-analysis hook would have the same capabilities as the pre-analysis hook but be called after the analysis stage.

benhayes21:
I think we need more information on this. From conversations, it sounded more like it was a "combining results"?

JBADougBurns:
It would be good to have a conversation on this. Our intention for now is to use the OasisLMF output summaries to generate some additional metrics, specifically things like sensitivity indices.

philipoldham:
We would want the post-analysis hook to be triggered _after_ writing results to disk, as the last thing to be run in an analysis. The hook should have the capability to access the input directory, the output directory, the model settings, and the model data (i.e. same as pre-hook but with the addition of the output directory).

PR #1371, opened 2023-09-18
Post analysis hook

carlfischerjba:
https://github.com/OasisLMF/OasisLMF/issues/1317

### Implement post analysis hook
Model vendors can supply a custom Python module that will be run after the analysis has completed. This module will have access to the run directory, model data directory and analysis settings. It could for instance modify the output files, parse logs to produce user-friendly reports or generate plots.

The two new Oasis settings required to use this feature are similar to the ones used for the pre analysis hook.
- `post_analysis_module`: Path to the Python module containing the class.
- `post_analysis_class_name`: Name of the class.

The class must have a constructor that takes kwargs `model_data_dir`, `model_run_dir` and `analysis_settings_json`, plus a `run` method with no arguments. For example:
```python
class MyPostAnalysis:
    def __init__(self, model_data_dir=None, model_run_dir=None, analysis_settings_json=None):
        self.model_data_dir = model_data_dir
        self.model_run_dir = model_run_dir
        self.analysis_settings_json = analysis_settings_json

    def run():
        # do something
```


oasislmf/computation/hooks/post_analysis.py
```diff
@@ -1,16 +1,43 @@
...
+        {'name': 'model_run_dir', 'is_path': True, 'pre_exist': False,
+         'help': 'Model run directory path'},
+        {'name': 'analysis_settings_json', 'is_path': True, 'pre_exist': True,
+         'help': 'Analysis settings JSON file path'},
+    ]
+
+    run_dir_key = 'post-analysis'
+
+    def run(self):
+        output_dir = os.path.join(self.model_run_dir, "output")
```
JCrane512, line 32: Do we not use pathlib in oasislmf?

tests/computation/test_post_analysis.py
```diff
@@ -0,0 +1,97 @@
+import json
+import pytest
+from oasislmf.manager import OasisManager
+
+
+def write_simple_post_analysis_module(module_path, class_name='PostAnalysis'):
+    with open(module_path, 'w') as f:
+        f.write(f'''
+from pathlib import Path
```
JCrane512, line 9: Is there a reason these modules are imported inside of this test, rather than at the top of the file?

JCrane512: Ah, I see it is used due to the way the module is imported.

tests/computation/test_post_analysis.py
```diff
@@ -0,0 +1,97 @@
...
+
+    kwargs = {'post_analysis_module': (tmp_path / 'post_analysis_simple.py').as_posix(),
+              'model_run_dir': tmp_path.as_posix()}
+
+    write_simple_post_analysis_module(kwargs['post_analysis_module'])
+
+    OasisManager().post_analysis(**kwargs)
+
+    assert (post_processed_output_dir / 'my_file.txt').read_text() == "Sample Data for Testing"
+    assert (post_processed_output_dir / 'gul_S1_aalcalc.csv').read_text() == "999"
```
JCrane512, line 53: If you want the test to be strict, should you check that no other files are present in this folder?

tests/computation/test_post_analysis.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_postanalysis_non_defaults(tmp_path):
+    model_run_dir = tmp_path
```
JCrane512, line 57: Technically the variable `model_run_dir` is unused, and you could have just used `tmp_path`. I assume this is done for readability?

tests/computation/test_post_analysis.py
```diff
@@ -0,0 +1,97 @@
...
+    raw_output_dir = model_run_dir / 'output'
+    raw_output_dir.mkdir(parents=True, exist_ok=True)
+    (raw_output_dir / 'gul_S1_aalcalc.csv').write_text("999")  # Create one output file.
+
+    post_processed_output_dir = raw_output_dir / 'postprocessed'
+
+    kwargs = {'post_analysis_module': (tmp_path / 'post_analysis_simple.py').as_posix(),
+              'model_run_dir': tmp_path.as_posix(),
+              'post_analysis_class_name': 'MyClass'}
```
JCrane512, line 66: Should you add a test, where you set `'post_analysis_class_name': 'MyClass_Does_Not_Exists' to check that an error is correctly raised?`

oasislmf/computation/hooks/post_analysis.py
carlfischerjba: We considered it but a lot of these things are strings and require some work to change them all. Separate piece of work?

tests/computation/test_post_analysis.py
carlfischerjba: Could do.

oasislmf/computation/hooks/post_analysis.py
```diff
@@ -1,16 +1,43 @@
...
+    step_params = [
+        {'name': 'post_analysis_module', 'required': True, 'is_path': True, 'pre_exist': True,
+         'help': 'Post-Analysis module path'},
+        {'name': 'post_analysis_class_name', 'default': 'PostAnalysis',
+         'help': 'Name of the class to use for the post_analysis'},
+        {'name': 'model_run_dir', 'is_path': True, 'pre_exist': False,
+         'help': 'Model run directory path'},
+        {'name': 'analysis_settings_json', 'is_path': True, 'pre_exist': True,
+         'help': 'Analysis settings JSON file path'},
+    ]
```
carlfischerjba, line 27: Add the `model_data_dir` to future proof this. If there anything else we or others may need in a postprocessing function?

oasislmf/computation/hooks/post_analysis.py
```diff
@@ -1,16 +1,43 @@
...
+                  'analysis_settings_json': self.analysis_settings_json}
+
+        _module = get_custom_module(self.post_analysis_module, 'Post-Analysis module path')
+        try:
+            _class = getattr(_module, self.post_analysis_class_name)
+        except AttributeError as e:
+            raise OasisException(f"Class {self.post_analysis_class_name} "
+                                 f"is not defined in module {self.post_analysis_module}") from e.__cause__
+
+        _class(**kwargs).run()
```
carlfischerjba, line 43: Why do we need a class? Couldn't we simplify and just call a function?
```suggestion
        _post_analysis_func(**kwargs)
```

tests/computation/test_post_analysis.py
JCrane512: I wrote a test for this

JCrane512: `test_postanalysis_wrong_module_name`

oasislmf/computation/hooks/post_analysis.py
```diff
@@ -1,16 +1,48 @@
...
+
+
+class PostAnalysis(ComputationStep):
+    """Computation step that is called after loss calculations.
+
+    It passes the output directory to a customisable function that might modify or add to the
+    standard output files.
+    """
+    step_params = [
+        {'name': 'post_analysis_module', 'required': True, 'is_path': True, 'pre_exist': True,
```
carlfischerjba, line 19: Can we easily support a Python dotted module path instead of a file system path? I think it would make configuration easier.

tests/computation/test_post_analysis.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def write_simple_post_analysis_module(module_path, class_name='PostAnalysis'):
+    with open(module_path, 'w') as f:
+        f.write(f'''
+from pathlib import Path
+import shutil
+import json
+
+class {class_name}:
```
carlfischerjba, line 16: This does the job but is rather convoluted.

sambles (changes requested):
Hi @carlfischerjba @JCrane512 @ksenijavasiljeva 

Thanks for the PR!  I've created a testing PR over at https://github.com/OasisLMF/OasisLMF/pull/1375/ To check CI. 

Here's some requested changes from that: 

1. https://github.com/OasisLMF/OasisLMF/pull/1375/commits/19cc7f8f934059035594b5161031ff8874a23a5b  Make the `post_analysis_module` argument not required for the **oasislmf model run** command. Otherwise running models without at post analysis module will fail with 
```
Required argument post_analysis_module could not be found in the command args or the MDK config.
``` 

2. https://github.com/OasisLMF/OasisLMF/pull/1375/commits/9dd2e66eea3c25b6329120a025dc051049663b41  Minor adjustment to make flake8 happy 

3. https://github.com/OasisLMF/OasisLMF/pull/1375/commits/e19224f51482d18e41c751cd8d8d5491022c7531 Add the post analysis command as a sub step in the **oasislmf generate-losses** command.  The worker in Platform 1 calls this command so its needed to run on the platform side.

4. https://github.com/OasisLMF/OasisLMF/commit/62a31a6299b22a00dc5f7163e5ab1ca209715c31 Fix for an existing unit test.

sambles (approved):

hchagani-oasislmf (approved):

PR #1373, opened 2023-09-26
store extra aggregation level BuildingId in summary_map and use it to…

sstruzik:
… compute correct tiv in summaries

### Fix Tiv calculation when NumberOfBuilding is >1 in location file
The Tiv calculated in the output summaries was incorrect as the granularity has change after the implementation of stochastic dis-aggregation (when NumberOfBuilding > 1).
Only 'loc_id', 'coverage_type_id' were taken in account leading to detect duplicate leading to lower TIV than it should
With this change, we add 'building_id' and 'risk_id' to the summary_map and add building_id in the key to detect duplicate when we calculate the TIV


sambles (approved):

sambles:
Overriding 2 reviewer limit for release

PR #1378, opened 2023-10-05
Support output of overall average period loss without standard deviation calculation

hchagani-oasislmf:
### Support output of overall average period loss without standard deviation calculation
The new ktools component `aalcalcmeanonly` (see PR https://github.com/OasisLMF/ktools/pull/357) calculates the overall average period loss but does not include the standard deviation. As a result, it has a faster execution time and uses less memory than `aalcalc`.

Support for executing this component as part of a model run has been introduced through the `aalcalc_meanonly` (legacy output) and `alt_meanonly`(ORD output) flags in the analysis settings file.


sambles:
> CI build for testing: https://github.com/OasisLMF/OasisLMF/actions/runs/6419834256

sambles:
> Retest with fixed error_guard and log text: https://github.com/OasisLMF/OasisLMF/actions/runs/6421223952

sambles (approved):

PR #1397, opened 2023-11-08
Add output zeros flag to summarycalc for all reinsurance loss computes

hchagani-oasislmf:
### Assign output zeros flag to summarycalc for all reinsurance loss computes
The `ktools` component `summarycalc` does not output zero loss events by default. These zero loss events are required when net loss is called in `fmpy`. Currently, net loss is called in all reinsurance instances, so the `-z` flag has been assigned to all executions of`summarycalc` when computing reinsurance losses.

### DIFF
--- a/oasislmf/execution/bash.py
+++ b/oasislmf/execution/bash.py
@@ -837,7 +837,7 @@ def do_summarycalcs(
     summarycalc_directory_switch = ""
     if runtype == RUNTYPE_REINSURANCE_LOSS:
         i = num_reinsurance_iterations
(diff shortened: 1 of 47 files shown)


benhayes21 (approved):

sambles:
https://github.com/OasisLMF/OasisLMF/actions/runs/6800990831

sambles (approved):
PiWind needs a results update to pass: 
https://github.com/OasisLMF/OasisLMF/actions/runs/6801135056

https://github.com/OasisLMF/OasisPiWind/pull/165

Issue #1384, opened 2023-10-27
Flag required in oasislmf 1.28.X to turn off disaggregation. 


JCrane512:
## Issue Description
`oasislmf 1.28.3`  introduced disaggregation. As JBA's complex flood model already performs disaggregation itself, this has lead to disaggregation being run twice, once by JBA's code and once again by `oasislmf`.

## Steps to Reproduce (Bugs only)

1. Run any JBA Flood Model with aggregate rows in the portfolio. 

## Version / Environment information

It will happen on any platform. The affected `oasislmf` version is `1.28.X` and higher. The last working configuration is `1.27.X`. 

## Example data / logs

We were finding that this line was suggesting the problem to us, it was taking multiple hours, instead of the normal >10 seconds. 
`2023-10-26 17:14:16,700 oasislmf log@128:wrapper [1386675:140364176003072] [INFO] COMPLETED: oasislmf.preparation.gul_inputs.get_gul_input_items in XXX`

johcarter:
Can we use this flag to disable the IsAggregate logic too? additional requirement from JBA

benhayes21:
https://github.com/OasisLMF/OasisLMF/pull/1401

PR #1398, opened 2023-11-10
Allow disaggregation to be disabled

carlfischerjba:
#1384

### Add option to turn off disaggregation
 Oasis settings now include an optional boolean `do_disaggregation` which defaults to `True`. By setting this to `False`, OasisLMF will no longer expand locations based on the field `NumberOfBuildings`.


johcarter:
this can be tested using the following test;
https://github.com/OasisLMF/OasisModels/tree/develop/PiWindPostcode/tests/test_5
This test has all six possible combinations of NumberOfBuildings and IsAggregate in the location file.

To run:
Default case (do_disaggregate=true)
oasislmf model run -C oasislmf.json -r runs/do_disagg

do_disaggregate = false
oasislmf model run -C oasislmf.json --do-disaggregation=false -r runs/do_not_disagg

In inputs, compare items (40 when true (default), 24 when false)
compare fm_summary_map output id (80 when true (default), 48 when false)

LocNumber Case2 has NumberOfBuildings=2 with IsAggregate=1 and increments the risk_id in fm_summary_map to 2 when the financial terms are split, but they are not split when the do_disaggregate=false (risk_id is 1 for all locations)

Conclusion:
The flag works correctly for all OED Agg cases. When not using the flag, the outputs are exactly the same as under 1.28.3.

johcarter (commented):
Works as expected against my test.

sstruzik (approved):

sambles (approved):

PR #1400, opened 2023-11-10
Testing PR for #1398 - Allow disaggregation to be disabled

sambles:
### Testing PR for #1398 - Allow disaggregation to be disabled


johcarter:
this can be tested using the following test; 
https://github.com/OasisLMF/OasisModels/tree/develop/PiWindPostcode/tests/test_5
This test has all six possible combinations of NumberOfBuildings and IsAggregate in the location file.

To run:
Default case (do_disaggregate=true)
oasislmf model run -C oasislmf.json -r runs/do_disagg

do_disaggregate = false
oasislmf model run -C oasislmf.json --do-disaggregation=false -r runs/do_not_disagg

In inputs, compare items (40 when true (default), 24 when false)
compare fm_summary_map output id (80 when true (default), 48 when false)

LocNumber Case2 has NumberOfBuildings=2 with IsAggregate=1 and increments the risk_id in fm_summary_map to 2 when the financial terms are split, but they are not split when the do_disaggregate=false (risk_id is 1 for all locations)

Conclusion:
The flag works correctly for all OED Agg cases.

carlfischerjba:
> Conclusion: The flag works correctly for all OED Agg cases.

@johcarter I'm surprised this works following what @sstruzik found during his testing. I've added further changes to https://github.com/OasisLMF/OasisLMF/pull/1398 to address those issues but they haven't yet been included in this PR (I get lost between all the PRs and copies of PRs - I don't have permission to update this one or the underlying branch).

This is what @sstruzik did to find that there was a problem:

> In OasisLMF mdk we have a command call exposure that bypass the model and just run 100% loss through the fm
> I wasn't sure if do_disaggregation is an available param so I just manually able or disable it in `oasislmf/computation/generate/files.py` code
> Because its 100% loss, it shouldn't matter if do_disaggregation is true or false but:
> - with true on you get total il=1,050,000
> - with False you get total il=570,000 (because the limits are still divided by number of building)
> I also disable the split manually and I get total il=1,050,000.
> command line I ran:
> ```
>  oasislmf exposure run -s ~/test/piwind_number_of_builing/ -r ./runs/piwind_number_of_builing_disagg
> ```

johcarter:
Yes, I don't understand this without looking in detail at Stephanes test.
I have performed the same test on your PR, @carlfischerjba , and also find it works correctly. Lets see if Stephane finds any issues there too.

sambles:
https://github.com/OasisLMF/OasisLMF/pull/1398

Issue #1088, opened 2022-07-20
Correlation options for the user


johcarter:
## Issue Description

Part of the correlation functionality described in #910  is to give the user the option to override correlation settings. This involves mirroring the correlation settings of model settings in analysis settings and overriding the default correlation factors.

#### Example 
In analysis settings 
```
“correlation_ settings”:[
  {”peril_correlation_group”:1, “correlation_value”:0.7},
  {”peril_correlation_group”:2, “correlation_value”:0.4},
]
```
in model settings
```
“correlation_ settings”:[
  {”peril_correlation_group”:1, “correlation_value”:0.5},
  {”peril_correlation_group”:2, “correlation_value”:0.2},
]
```
Then use 0.7 and 0.4 in place of 0.5 and 0.2 in oasislmf model run

johcarter:
Updated requirements for upcoming changes to correlation parameters in model settings json.
@maxwellflitton 

Correlations file format is changing, PR reference:
https://github.com/OasisLMF/OasisLMF/pull/1181

In analysis settings (optional)
```
“correlation_ settings”:[
  {”peril_correlation_group”:1, “damage_correlation_value”:0.7, "hazard_correlation_value":0.8},
  {”peril_correlation_group”:2, “damage_correlation_value”:0.4, "hazard_correlation_value":0.3},
]
```
in model settings
```
“correlation_ settings”:[
  {”peril_correlation_group”:1,  “damage_correlation_value”:0.5, "hazard_correlation_value":0.6},
  {”peril_correlation_group”:2, “damage_correlation_value”:0.2, "hazard_correlation_value":0.1},
]
```
When correlation settings appear in analysis settings,  override the correlation values populated in the correlations file with those in analysis settings instead of those specified in model settings.

Correlations file example (new format ) 


item_id | peril_correlation_group | damage_correlation_value | hazard_group_id | hazard_correlation_value
-- | -- | -- | -- | --
1 | 1 | 0.7 | 1008141099 | 0.8
2 | 2 | 0.4 | 354519032 | 0.3
.. | ..  | ...   | ..                 | ..

mtazzari:
I started working on this but right now we're moving the analysis settings schema to the ODS repo, and new Python API is being implemented to deal with the schema, including validation, so it's better to wait until the move is done, i.e. when these are merged:
- #995 
- https://github.com/OasisLMF/ODS_Tools/pull/1

Issue #910, opened 2021-10-05
improve correlation functionality available through the engine


johcarter:
## Issue Description

improve correlation functionality available through the engine
Nazare Workflow_Id 1000 Task_Id 1003

johcarter:
Needs clarification from steering group

johcarter:
[correlation_proposals_v1_shared.pdf](https://github.com/OasisLMF/OasisLMF/files/8762052/correlation_proposals_v1_shared.pdf)

mtazzari:
Done

PR #1401, opened 2023-11-14
Correlations info in analysis settings takes precedence over that in model settings

ncerutti:
### Correlations info in analysis settings takes precedence over that in model settings
The default model correlation factors can be overwritten by specific "correlation_settings" added to the analysis settings file.

Described in Issue #1088 (and related to Issue #910).

#### Notes:
Nothing will be done in case no model settings are present. Also note that only the correlation factors will be overwritten (i.e., the supported perils still must be set in the model settings).


johcarter (commented):
Works perfectly, @ncerutti thank you! 

Although it would work exactly the same, I wonder whether the correct place for the correlation_settings would be nested in the model_settings{} blob within the analysis settings json file? Since specifying the correlation factors depends on what the peril correlation groups are, which are model specific and must also be specified in model_settings json.

johcarter:
what do you think about the above question please @sstruzik ?

johcarter (approved):
Works as expected when analysis correlation settings are nested in model_settings blob in analysis settings json.

As long as peril correlation groups are defined by the modeller, analysis correlations setting are used to overwrite model correlation settings.  Model correlation_settings do not need to be defined (model settings json, correlation_settings{} blob present ) for analysis correlation settings to be applied.  I think this behaviour is fine.

ncerutti:
> Works as expected when analysis correlation settings are nested in model_settings blob in analysis settings json.
> 
> As long as peril correlation groups are defined by the modeller, analysis correlations setting are used to overwrite model correlation settings. Model correlation_settings do not need to be defined (model settings json, correlation_settings{} blob present ) for analysis correlation settings to be applied. I think this behaviour is fine.

Thank you, @johcarter! If the need arises in the future, further adjusting the fields to overwrite and their sources should be straightforward.

sstruzik (approved):

PR #1407, opened 2023-11-22
Added tests for condition coverages 1-5 financial terms

johcarter:
### Added tests for condition coverages 1-5 financial terms
Completed all units in validation/insurance_policy_coverages


johcarter:
@sstruzik the condition coverages 1-5 tests are units Q5 9-16 in validation/insurance_policy_coverages. I've added these units to the test and they all pass.

The 7 fmpy validation tests that still fail on this branch are due to the 'TypeError: boolean value of NA is ambiguous' caused by empty CondPriority field. Once that is fixed, the tests should all pass.

PR #1420, opened 2023-12-15
Fix/double condtag

sstruzik:
### support location with an empty condtag and a priority > 1 condtag
When a location doesn't have a specific condition for (level 1) but is group with other location for a condition with higher priority, it is not required to specify an empty condtag in the location on top of the priority > 1 condtag.
However, if the user does it, it should still be handle correctly and not raise an error.


sambles (approved):

johcarter:
added more tests sc18 and sc19 to investigate loss results

Issue #1425, opened 2024-01-09
Customise specific vulnerabilities (without providing full replacement data)


ncerutti:
## Issue Description

Currently, adjustments to individual vulnerabilities require the complete replacement of the associated data for that vulnerability (see PR #1406 and #1415). This process involves either specifying a distinct file for usage or incorporating the data via an additional file or field within the settings file.

To simplify the process of tweaking vulnerability parameters, there should be a straightforward way to adjust individual vulnerabilities without replacing the whole vulnerability function.

PR #1426, opened 2024-01-09
Adjustments to specific vulnerabilities (by tweaking generated random numbers)

ncerutti:
### Factor adjustments to specific vulnerabilities
By specifying adjustments to specific vulnerabilities in the analysis settings, an adjustment can be applied to the probabilities of that vulnerability.

This PR solves Issue #1425.
It provides with a tool to adjust specific vulnerabilities so that they are more or less likely to lead to a higher or lower damage bin. All the random numbers generated for the specific vulnerability id will be multiplied by the factor present in the "vulnerability_adjustments", "adjustments" key in the analysis_settings file.

### Example

the key 
```
"vulnerability_adjustments": {
    "adjustments": {"2": 0.95, "4": 1.10}
}
```
is in the analysis_settings.json file.
Whenever a random number is generated for the vulnerability id 2, it will be multiplied by 0.95 before being used to sample the damage bin. This means it is more likely fall into a lower damage bin. As in the case of vulnerability_id 4, a number >1 can be chosen, instead, to increase the likelihood of it falling into a higher damage bin.


ncerutti:
The names of the keys in analysis settings were changed to accommodate the fact that for the model settings schema it is better not to have a single field with different valid values (dict OR path).

benhayes21 (approved):

sstruzik (approved):

PR #1427, opened 2024-01-15
fix make Othertiv optional

sstruzik:


sambles:
Added with https://github.com/OasisLMF/OasisLMF/pull/1424

Issue #1350, opened 2023-07-27
model settings - correlation settings - allow optional hazard or damage correlation value 


johcarter:
## Issue Description

model settings caused this error:
ValueError: Expect correlation settings in model settings file to contain a `hazard_correlation_value` in each `peril_correlation_group` entry, got 
 [{'peril_correlation_group': 1, 'damage_correlation_value': '0'}, {'peril_correlation_group': 2, 'damage_correlation_value': '0'}].

However the specification of correlation value for either damage or hazard should be optional and default to zero if not entered.
model settings file causing the error attached for OasisModels/PiWind

## Steps to Reproduce (Bugs only)

1. Run OasisModels/PiWind with the attached model settings
2.
3.

## Version / Environment information

1.28
## Example data / logs

[error.log](https://github.com/OasisLMF/OasisLMF/files/12181264/error.log)
[model_settings_0_correlation.zip](https://github.com/OasisLMF/OasisLMF/files/12181269/model_settings_0_correlation.zip)

PR #1456, opened 2024-02-27
Allow optional hazard or damage correlation value

ncerutti:
### Hazard and damage correlation values in model settings now optional
The correlation value for either damage or hazard is now optional and defaults to zero if not entered.

Solves Issue #1350.


sstruzik (approved):

johcarter (approved):
All good

PR #1465, opened 2024-02-28
Bring RI file headers in line with those of FM

hchagani-oasislmf:
### Bring RI file headers in line with those of FM
In the FM files `fm_policytc.csv` and `fm_profile.csv`, the field name `policytc_id` has been replaced by `profile_id`. This brings these files in line with those of RI.

~~In the RI files `fm_policytc.csv` and `fm_profile.csv`, the field name `profile_id` has been replaced by `policytc_id`. This brings these files in line with those of FM.~~


benhayes21:
@hchagani-oasislmf  - sorry, I think I might have suggested we go the wrong way with policy_tc id over profile_id. I've just realised that fmprofiletocsv also has profile_id as the header, so that is probably the correct one to have

benhayes21 (commented):
sorry - I think we should switch to profile_id as the standard. This can wait until 2.3.2

sambles (approved):

benhayes21 (approved):

Issue #1258, opened 2023-04-17
Document oasislmf.json configuration options 


sambles:
## Issue Description
Add missing documentation to list valid `oasislmf model run` options that can be set in a `oasislmf.json` configuration file. 
Should be automatically generated using sphinx

## Example Python to generate Markdown

```
from oasislmf.cli.model import RunCmd

RunCmd().arg_parser._actions
cmd_opts = RunCmd().arg_parser._actions
cmd_opts.pop(0)  # remove help command 


# markdown template 
def cmd_to_markdown(name, desc, default):
    return([
        f'## {name}',
        '',
        f'* Description: {desc}',
        f'* Default value: {default}',
        '* Example use',
        '```',
        '{',
        f' "{name}": <value>,',
        '}',
        '```',
        ''
])

markdown_output = []
for opt in cmd_opts:
    markdown_output += cmd_to_markdown(opt.dest, opt.help, opt.default)

print('\n'.join(markdown_output))
```  
 
## Example Markdown and html
[valid_oasislmf_options.md](https://github.com/OasisLMF/OasisLMF/files/11249542/valid_oasislmf_options.md)
[valid_oasislmf_options.html.zip](https://github.com/OasisLMF/OasisLMF/files/11249543/valid_oasislmf_options.html.zip)

PR #1490, opened 2024-04-26
Add .json configuration to the sphinx docs generation process 

ncerutti:
### oasislmf.json options now included in docs generated by sphinx

Solves #1258. When sphinx creates the documentation, now along with Installation, Building and Running Models, and API Reference, there is also an entry detailing the options that can be inserted in the oasislmf.json configuration file (Options for the JSON Configuration File).


ncerutti:
Old output (pre-PR) can be seen at: https://oasislmf.github.io/OasisLMF/index.html

New output can be seen in https://github.com/OasisLMF/OasisLMF/actions/runs/8907250506/artifacts/1463327218 (`Options for the JSON Configuration File` section)


Only question is: should I delete the `example use` after each entry? I would, unless we expect some to be different, eventually.

sambles:
The linked output asset expired, re-ran job and attaching here: 
[github-pages.zip](https://github.com/OasisLMF/OasisLMF/files/15197850/github-pages.zip)

sambles:
> Only question is: should I delete the example use after each entry? I would, unless we expect some to be different, eventually. 

Yes I agree, its not all that useful, something like expected types would be better. 
`type=<function str2bool at 0x7f4f6194c180>`  --> expected type are ['True', 'False', 1, 0 .. etc]
`type=<class 'int'>`  -->  expected type Int

But I'm noticing many flags don't have a `type` set, (like the filepath stuff)

sambles:
I think the **Default value** is not populated either (Wrong from the original example! so that one is on me) 

They all seem to be listed as `None`, this information must be somewhere, but maybe not added into `RunCmd().arg_parser._actions` ?

ncerutti:
Here is the output:

[github-pages.zip](https://github.com/OasisLMF/OasisLMF/files/15414940/github-pages.zip)

sambles (approved):

benhayes21 (approved):

PR #1504, opened 2024-05-10
Added guard for summarycalc when output groupings over the max of 9 

sambles:
### Added guard for summarycalc when output groupings over the max of 9 
* With the addition of `summarypy` more than 9 output groups are now supported. However summarycalc will crash, 
added a check to catch this before going to execution. 

* Updated the bash script error trap to only exclude scripts starting with `*startup.sh`. This to prevent workers going down, but making sure any subshells `*/run_ktools.sh` are not left up which could cause an execution to hang.

> This guard is required because of https://github.com/OasisLMF/ODS_Tools/pull/113

### DIFF
--- a/oasislmf/computation/generate/losses.py
+++ b/oasislmf/computation/generate/losses.py
@@ -260,7 +260,6 @@ class GenerateLossesDir(GenerateLossesBase):
                     '\nThis can be set using the flag "--summarypy True", or by setting `"summarypy": True` in the oasislmf.json file.'
                 )
 
-
     def run(self):
         # need to load from exposure data info or recreate it
         model_run_fp = self._get_output_dir()
@@ -671,7 +670,6 @@ class GenerateLosses(GenerateLossesDir):
         script_fp = os.path.join(os.path.abspath(model_run_fp), 'run_ktools.sh')
         ri_layers = self._get_num_ri_layers(analysis_settings, model_run_fp)
         model_runner_module, package_name = self._get_model_runner()
-
 
         with setcwd(model_run_fp):
             try:
--- a/oasislmf/computation/generate/losses.py
+++ b/oasislmf/computation/generate/losses.py
@@ -244,6 +244,23 @@ class GenerateLossesDir(GenerateLossesBase):
                         )
                     return  # Only need to find a single request
 
+    def __check_summary_group_support(self, analysis_settings, runtypes):
+        """
+        Private method to check the max number of summary groups selected.
+        If that value is greater than 9 then ktools 'summarycalc' will crash.
+
+        Stop execution if the summarypy flag is not set as True.
+        """
+        for runtype in runtypes:
+            summary_num = len(analysis_settings.get(f'{runtype}_summaries', []))
+            if summary_num > 9 and not self.summarypy:
+                raise OasisException(
+                    'More than 9 summaries groups are not supported in summarycalc.'
+                    f'\nEither enable summarypy or reduce the number of groups set in "{runtype}_summaries".'
+                    '\nThis can be set using the flag "--summarypy True", or by setting `"summarypy": True` in the oasislmf.json file.'
+                )
+
+
     def run(self):
         # need to load from exposure data info or recreate it
         model_run_fp = self._get_output_dir()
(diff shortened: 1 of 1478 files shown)


oasislmf/computation/generate/losses.py
```diff
@@ -244,6 +244,22 @@ def __check_for_parquet_output(self, analysis_settings, runtypes):
                         )
                     return  # Only need to find a single request
 
+    def __check_summary_group_support(self, analysis_settings, runtypes):
+        """
+        Private method to check the max number of summary groups selected.
+        If that value is greater than 9 then ktools 'summarycalc' will crash.
```
hchagani-oasislmf, line 250: It's not that `summarycalc` would crash. For example:
```
$ summarycalc -m -i -10 output_file.bin < input_file.bin
```
would create the files `0` and `0.idx`, for summary group ID 1. Also:
```
$ summarycalc -m -i -20 output_file.bin < input_file.bin
```
would return an error if there was no data with for summary group ID 2.

Perhaps something like this would be more accurate:
```suggestion
        Private method to check the max number of summary groups selected.
        Setting this value greater than 9 will lead to undesirable
        behaviour in ktools 'summarycalc'.
```

hchagani-oasislmf (approved):
Just one minor suggestion. Otherwise, looks good.

oasislmf/computation/generate/losses.py
sambles: I think its actually a bit worse, the execution seems to hang after something detects something has gone wrong
```
compute_info : {'allocation_rule': 2, 'max_level': 3, 'max_layer': 1, 'node_len': 72, 'children_len': 102, 'parents_len': 70, 'profile_len': 32, 'loss_len': 111, 'extra_len': 0, 'compute_len': 77, 'start_level': 0, 'items_len': 40, 'output_len': 40, 'stepped': False}
RUNNING: oasislmf.execution.runner.run
/home/sam/repos/models/piwind/runs/losses-20240510110824/run_ktools.sh: line 30: 60453 Killed                  ktools_monitor.sh $$ $LOG_DIR
grep: warning: * at start of expression
grep: warning: * at start of expression

```

sambles: thats from using: 
```
{                                                                                                                                             
    "version": "3",
    "analysis_tag": "base_example",
    "source_tag": "MDK",
    "model_name_id": "PiWind",
    "model_supplier_id": "OasisLMF",
    "gul_threshold": 0,
    "gul_output": true,
    "model_settings": { 
        "event_set": "p",
        "event_occurrence_id": "lt"
    },
    "gul_summaries": [ 
        { "eltcalc": true, "id": 1 },
        { "eltcalc": true, "id": 2 },
        { "eltcalc": true, "id": 3 },
        { "eltcalc": true, "id": 4 },
        { "eltcalc": true, "id": 5 },
        { "eltcalc": true, "id": 6 },
        { "eltcalc": true, "id": 7 },
        { "eltcalc": true, "id": 8 },
        { "eltcalc": true, "id": 9 },
        { "eltcalc": true, "id": 10 },
        { "eltcalc": true, "id": 11 },
        { "eltcalc": true, "id": 12 } 
    ],
    "il_output": false,
    "il_summaries": [ ],
    "ri_output": false,
    "ri_summaries": [ ] 
}

```

hchagani-oasislmf: Ah, interesting. I just ran `summarycalc` from the command line to see what it would do.

sambles: I was seeing this in `stderror.err` which is what triggers the kill: 
```
FATAL: openpipe: Summary set ID 1 requested at least twice                                                                                    
FATAL: openpipe: Summary set ID 1 requested at least twice
FATAL: openpipe: Summary set ID 1 requested at least twice
FATAL: openpipe: Summary set ID 1 requested at least twice
FATAL: openpipe: Summary set ID 1 requested at least twice
FATAL: openpipe: Summary set ID 1 requested at least twice
FATAL: openpipe: Summary set ID 1 requested at least twice
  ...
```

hchagani-oasislmf: That makes sense, because it sees `-1` and `-10` as equivalent, i.e. as `-1`. Then I think what you have written in the docstring is fine.

sstruzik (approved):

PR #1508, opened 2024-05-17
fix pla reader

sstruzik:
### Fix plapy reader creating negative event in output
Plapy reader is a bit different from the other reader using the generic approach because it return all events at ones even if not finished.
This was not handle properly for big even. This fix remediate it by keeping in memory the last event and item id and updating the read and write memory view

### DIFF
--- a/oasislmf/pytools/pla/streams.py
+++ b/oasislmf/pytools/pla/streams.py
@@ -54,7 +54,6 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id, items_amps, plaf
                 break
             event_id, cursor = mv_read(byte_mv, cursor, oasis_int, oasis_int_size)
             item_id, cursor = mv_read(byte_mv, cursor, oasis_int, oasis_int_size)
-
             ##### do new item setup #####
             factor = plafactors.get((event_id, items_amps[item_id]), default_factor)
             ##########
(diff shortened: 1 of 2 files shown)


benhayes21 (approved):

oasislmf/pytools/pla/streams.py
```diff
@@ -70,13 +69,18 @@ def __init__(self, items_amps, plafactors, default_factor):
...
         self.logger = logger
 
     def read_buffer(self, byte_mv, cursor, valid_buff, event_id, item_id):
-        return read_buffer(
-            byte_mv, cursor, valid_buff, event_id, item_id,
+        cursor, self.event_id, self.item_id, yield_event = read_buffer(
+            byte_mv, cursor, valid_buff, self.event_id, self.item_id,
             self.items_amps, self.plafactors, self.default_factor, self.out_byte_mv, self.out_cursor
         )
+        return cursor, self.event_id, self.item_id, yield_event
```
hchagani-oasislmf, line 83: It looks fine to me @sstruzik, but just a question: are `event_id` and `item_id` are unused variables in this method?

hchagani-oasislmf (approved):

oasislmf/pytools/pla/streams.py
sstruzik: yes, but it is in the generic function signature read_buffer(self, byte_mv, cursor, valid_buff, event_id, item_id)
That's why it is there.
Normally each type a event is fully read, the read return the event but in pla we don't need that we can read the full buffer update the loss as we read and return the full buffer regardless of the item. however we need to keep track of event_id and item_id that are normally reset to 0 if we tell the upper function to yield the event

PR #1513, opened 2024-05-30
fix fm files generation for terms with non 0 default value

sstruzik:
### fix fm files generation for terms with non 0 default value
Some terms such as LayerParticipation and AccParticipation have a non 0 default value that was not taken into account during the fm file generation.
In this fix we use ods_tools to get the default value for the term and add it if necessary.


sambles (approved):

sambles:
> Note: Must pin ods-tools for this PR

Issue #1402, opened 2023-11-14
Adding support for ranges of events, not just individual IDs


aiste-kalinauskaite:
Currently an analysis can be run using one or more individual IDs, as a comma separated list. However, if a user is trying to a few consecutive events, it can be rather tedious to specify individual IDs. Could you add ability to specify event IDs as ranges, as well as individual IDs?

Analysis_settings.json could be updated as follows:

```
{ 
   ...
   "event_ids": [1-3, 20-22, 400, 1000-1002, 9999, 10000] 
}

```
So in this case the following event IDs would be considered: 1,2,3,20,21,22,400,1000,1001,1002,9999,10000.

PR #1517, opened 2024-06-18
event ranges

benhayes21:
#1402 

### Support for ranges of events in analysis settings
Adding support for additional ranges of events in analysis settings, in addition to the list of event ids which can already be provided.

Event id ranges should be submitted using the new "event_ranges" option in analysis settings

The string data will take the format "1-4,8,89-94" which will be parsed and an event list created.

It can be used in combination with the existing list, and unique values will be established


ncerutti (approved):

sambles:
Don't we need this?

sstruzik (approved):

PR #1534, opened 2024-07-18
Feature/1468 intermediate ri inuring priorities

sstruzik:
### Allow to have loss output at intermediate RI inuring priorities

-  create new rl_outputs and rl_summaries parameters in analysis settings to drive all reinsurance loss perspective outputs

-  relabel the existing reinsurance loss outputs to use the code rl rather than ri, but otherwise keep the same naming convention

-  use existing ri_inuring_priorities parameter in analysis settings to drive new intermediate reinsurance net loss perspectives with output files labelled ri. This involves some genbash work to 'tee' the fmpy ri net stream to files as well as passing it down the pipe to the next RI calculation

As part of this work and to prepare for the UI controls, we are considering making changes to the run dir;

-  move ri_layers.json into the inputs folder (so that the available net perspectives may be read by the UI)

-  move RI_[X] folders from the root run dir to be nested in the inputs dir, and make the necessary changes to genbash to change the relative folder path of the RI_[X] folders in the run_ktools script.


sstruzik:
need https://github.com/OasisLMF/OasisPiWind/pull/185 to have run_mdk CI work

benhayes21 (approved):

sambles (approved):

PR #1552, opened 2024-08-30
Make analysis and model setting able to modify any computation step parameters (MDK parameters)

sstruzik:
### Make analysis and model setting able to modify any computation step parameters (MDK parameters)
To be accessible to the user and modeler via analysis and model settings, MDK parameters use to need a specific treatment in the code to allow the overwrite.
This PR makes it automatic by using the ods_tools settings object that help merge the different settings and parameters in a coherent way.


sambles:
Overriding 2nd review for platform testing

PR #1555, opened 2024-09-05
disable full_correlation if running gulmc

sstruzik:
### Disable full_correlation when running gulmc
Gulmc provide a more fine grained correlation feature and make full_correlation flag not useful and created an issue where the run would hang
This change disable full_correlation in that case fixing the hanging issue.


sambles (approved):

sambles:
Overriding two review requirement - 3 line change, looks good

PR #1558, opened 2024-09-12
support for having several pla sets

sstruzik:
### support for having several pla sets
Allow user to specify a different pla set from the default one by adding loss_factors_set in the model_settings part of analysis_settings.json
if "loss_factors_set": "2" is added
the loss factor that will be taken into account will be lossfactors_2.bin in the static folder


sambles:
Note: update name to `pla_loss_factors_set`

sambles (approved):

PR #1584, opened 2024-10-25
Fix: undo 100% correlation if no correlation setting provided in model settings

sstruzik:
### Fix: undo 100% correlation if no correlation setting provided in model settings
The hazard_group_id where not generated if the correlation settings were not provided, causing a 100% correlation across all location.
This fix assure the  hazard_group_id are generated and used correctly in gulmc.

### DIFF
--- a/oasislmf/computation/generate/files.py
+++ b/oasislmf/computation/generate/files.py
@@ -181,7 +181,6 @@ class GenerateFiles(ComputationStep):
         location_profile = get_json(src_fp=self.profile_loc_json) if self.profile_loc_json else self.profile_loc
         accounts_profile = get_json(src_fp=self.profile_acc_json) if self.profile_acc_json else self.profile_acc
         oed_hierarchy = get_oed_hierarchy(location_profile, accounts_profile)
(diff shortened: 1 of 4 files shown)


benhayes21 (approved):

johcarter:
The correlation file is being generated when correlation_settings are missing from model settings on the fixed branch, which is the expected behaviour.

I see another separate issue here, which is the damage_group_id and hazard_group_ids are identical when data settings are the same for hazard and damage.  This would mean that the same random numbers would be used to seed hazard and damage sampling, which would induce unwanted correlation between hazard and damage per sample.  Although we may want the logical groupings to be the same, we want a different hash to prevent this correlation from occurring.

proposed fix is to add a constant to the hashing algorithm to force hazard group_ids to be different to damage group ids, but leave the damage group id hashing as it is.
FYI @benhayes21

johcarter (changes requested):
issue fixed, but also need to hash hazard_group_id differently to damage_group_id when data settings are identical

johcarter (approved):
even though hazard_group_id and damage_group_id may be the same, we've checked that the random numbers are seeded differently, which is what we need. No changes needed

Issue #1528, opened 2024-07-02
Output Calc Python Rewrite


benhayes21:
## Issue Description
Rewrite output calc components in python

- [ ] AAL Calc
- [ ] LEC Calc
- [ ] ELT Calc
- [ ] PLT Calc
- [ ] Full Summary Output

PR #1590, opened 2024-11-07
Feature/pltpy

SkylordA:
### Pltpy, python replacement for pltcalc

This PR introduces the possibility of creating some summary files (SPLT, MPLT, QPLT) with the python replacement of the original pltcalc found in ktools.

Solves PLT in https://github.com/OasisLMF/OasisLMF/issues/1528


benhayes21 (approved):

oasislmf/pytools/common/input_files.py
```diff
@@ -0,0 +1,186 @@
...
+        integer_part = int(pos)
+        fractional_part = pos - integer_part
+        intervals.append((q, integer_part, fractional_part))
+
+    # Convert to numpy array
+    intervals = np.array(intervals, dtype=quantile_interval_dtype)
+    return intervals
+
+
+def read_occurrence(run_dir, filename=OCCURRENCE_FILE):
```
sstruzik, line 80: Do we need to read sequentially here?
For me the occurrence file is a binary file with a csv like structure
You know at the beginning if it is granular or not
then you just put the rest of the bin in a np.array

oasislmf/pytools/plt/cli.py
```diff
@@ -0,0 +1,35 @@
...
+import argparse
+import logging
+
+from . import manager, logger
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Process period loss table stream')
+    parser.add_argument('--run_dir', type=str, default='.', help='path to the run directory')
+    parser.add_argument('--files_in', type=str, nargs='+', required=False, help='Input files')
```
sstruzik, line 12: Can you also add -i as a short flag

oasislmf/pytools/common/input_files.py
SkylordA: No, I'll update pytools-common to just use np.frombuffer instead of the for loop and merge it into the other branches.

oasislmf/pytools/plt/cli.py
SkylordA: Sure, I'll update the other branches too

sstruzik (approved):

PR #1592, opened 2024-11-07
Fix/dynamic footprint intensity

### LINKED ISSUE #1591: Dynamic footprint has incorrect type 1 losses
## Issue Description
In the dynamic footprint, intensity (instead of intensity_bin_id) is being used to calculate type 1 losses. Unless intensity and intensity_bin_id are the same, this results in incorrect type 1 losses (run may or may not fail, results are often non-repeatable).

1. Run a dynamic footprint model where intensity =/= intensity_bin_id in the intensity bin dictionary.

2.3.6+

siobhancampbell:
Closes #1591 

The effective vulnerability CDF used to generate type 1 losses is created before compute_event_losses is called (which is where intensities are adjusted and mapped to bins). FootprintParquetDynamic has been updated to return both intensity and intensity bin, so the intensity bin can be used to create the effective vulnerability CDF, rather than the physical intensity

Type 1 losses **do not** account for intensity adjustments with this implementation. This will require some additional refactoring to move the creation of the effective vulnerability CDF to inside compute_event_losses

### DIFF
--- a/oasislmf/pytools/getmodel/footprint.py
+++ b/oasislmf/pytools/getmodel/footprint.py
@@ -400,11 +400,15 @@ class FootprintParquetDynamic(Footprint):
             df_footprint['from_intensity'] = df_footprint['from_intensity'].fillna(0)
 
             if len(df_footprint.index) > 0:
-                df_footprint['intensity_bin_id'] = np.floor(df_footprint.from_intensity + (
+                df_footprint['intensity'] = np.floor(df_footprint.from_intensity + (
                     (df_footprint.to_intensity - df_footprint.from_intensity) * df_footprint.interpolation))
-                df_footprint['intensity_bin_id'] = df_footprint['intensity_bin_id'].astype('int')
+                df_footprint['intensity'] = df_footprint['intensity'].astype('int')
+                intensity_bin_dict = pd.read_csv('static/intensity_bin_dict.csv')
+                intensity_bin_dict.rename(columns={'intensity_bin': 'intensity_bin_id'}, inplace=True)
+                df_footprint = df_footprint.merge(intensity_bin_dict, on='intensity')
                 df_footprint['probability'] = 1
             else:
+                df_footprint.loc[:, 'intensity'] = []
                 df_footprint.loc[:, 'intensity_bin_id'] = []
                 df_footprint.loc[:, 'probability'] = []
 
--- a/oasislmf/pytools/getmodel/footprint.py
+++ b/oasislmf/pytools/getmodel/footprint.py
@@ -16,7 +16,7 @@ import numba as nb
 from oasis_data_manager.df_reader.config import clean_config, InputReaderConfig, get_df_reader
 from oasis_data_manager.df_reader.reader import OasisReader
 from oasis_data_manager.filestore.backends.base import BaseStorage
-from .common import (FootprintHeader, EventIndexBin, EventIndexBinZ, Event, EventCSV,
+from .common import (FootprintHeader, EventIndexBin, EventIndexBinZ, Event, EventCSV, EventDynamic,
                      footprint_filename, footprint_index_filename, zfootprint_filename, zfootprint_index_filename,
                      csvfootprint_filename, parquetfootprint_filename, parquetfootprint_meta_filename,
                      event_defintion_filename, hazard_case_filename, fp_format_priorities)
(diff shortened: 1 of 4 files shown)


oasislmf/pytools/getmodel/footprint.py
```diff
@@ -400,15 +400,33 @@ def get_event(self, event_id: int):
             df_footprint['from_intensity'] = df_footprint['from_intensity'].fillna(0)
 
             if len(df_footprint.index) > 0:
-                df_footprint['intensity_bin_id'] = np.floor(df_footprint.from_intensity + (
+                df_footprint['intensity'] = np.floor(df_footprint.from_intensity + (
                     (df_footprint.to_intensity - df_footprint.from_intensity) * df_footprint.interpolation))
-                df_footprint['intensity_bin_id'] = df_footprint['intensity_bin_id'].astype('int')
+                df_footprint['intensity'] = df_footprint['intensity'].astype('int')
+                intensity_bin_dict = pd.read_csv('static/intensity_bin_dict.csv')
```
sstruzik, line 406: We have an abstraction to read what is in static. This should work and be in line with the rest of the code
intensity_bin_dict =  self.get_df_reader('intensity_bin_dict.csv').as_pandas()

sstruzik (changes requested):
Thanks for your PR, I made a couple of suggestion

oasislmf/pytools/getmodel/footprint.py
sstruzik: thanks I'll have a look

oasislmf/pytools/getmodel/footprint.py
```diff
@@ -400,15 +400,33 @@ def get_event(self, event_id: int):
...
+                intensity_bin_dict.rename(columns={'intensity_bin': 'intensity_bin_id'}, inplace=True)
+                df_footprint = df_footprint.merge(intensity_bin_dict, on='intensity')
                 df_footprint['probability'] = 1
             else:
+                df_footprint.loc[:, 'intensity'] = []
                 df_footprint.loc[:, 'intensity_bin_id'] = []
                 df_footprint.loc[:, 'probability'] = []
 
-            numpy_data = self.prepare_df_data(data_frame=df_footprint[['areaperil_id', 'intensity_bin_id', 'probability', 'return_period']])
```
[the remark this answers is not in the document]
siobhancampbell: Done

siobhancampbell (commented):
Made the buffer change as requested, but leaving the pd.read_csv for the intensity_bin_dict, due to bug in Oasis Data Manager

siobhancampbell:
I'm seeing the same issue PiWind was having when running a non flood model

If jit is enabled, the code in `if dynamic_footprint` is being evaluated, even if dynamic footprint was not set (it defaults to false in `gulmc.manager.run()`). If jit is disabled, it works as expected.

I tried `if dynamic_footprint == True` and  `if bool(dynamic_footprint)`, but same issue. Is there something else I can try?

sambles:
> I'm seeing the same issue PiWind was having when running a non flood model
> 
> If jit is enabled, the code in `if dynamic_footprint` is being evaluated, even if dynamic footprint was not set (it defaults to false in `gulmc.manager.run()`). If jit is disabled, it works as expected.
> 
> I tried `if dynamic_footprint == True` and `if bool(dynamic_footprint)`, but same issue. Is there something else I can try?

This rings a bell, its something to do with branch prediction, i'm pretty sure @sstruzik has hit this before.  I think there is a directly you can use in the python decorator

siobhancampbell:
Adding `error_model="numpy"` to the decorator did not help unfortunately.

`if dynamic_footprint: ...` was added to `gulmc.manager.compute_event_losses()` and `gulmc.manager.process_areaperils_in_footprint()` here - but `compute_event_losses()` already has a boolean used for an if statement (`if effective_damageability: ...`). Is that one working as expected when JIT is enabled?

siobhancampbell:
Removing cache=True and fast_math=True from the decorator also does not help...

I also tried this:
```
dynamic_footprint=False
if dynamic_footprint:
    haz_cdf ...
```
but the expression evaluated anyway.

However, when I changed the variable name to something random, it worked as expected:
```
zzzzzzz=False
if zzzzzzz:
    haz_cdf ...
```

Does that help pin down the issue at all?

siobhancampbell:
Forcing Numba to compile in object mode - `@jit(cache=True, fastmath=True, forceobj=False)` - does fix the issue, but this means it's no longer compiling in nonpython mode / machine code.

This change would have a performance impact, but Numba's nonpython mode seems to be struggling with control flow based on Boolean flags.

sstruzik (approved):

sambles (approved):

PR #1594, opened 2024-11-15
improve the memory performance of il layer number continuity step

sstruzik:
### OOM issue for accounts with lots of layers for several levels
Improve the memory performance of level matching between levels of accounts that have lots of layers.


benhayes21 (approved):

johcarter (approved):
checked that fm validation tests all pass

PR #1609, opened 2024-12-10
refactor/pytools-common

SkylordA:
### refactor/pytools-common
Include common code functionality refactor for pytools output calc feature branches
- feature/eltpy
- feature/pltpy
- feature/aalpy
- feature/lecpy


oasislmf/pytools/common/data.py
```diff
@@ -141,3 +156,183 @@ def load_as_array(dir_path, name, _dtype, must_exist=True):
...
+            return np.array([], dtype=oasis_int), np.array([], dtype=oasis_float)
+        unique_event_ids = data['event_id']
+        event_rates = data['rate']
+
+        # Make sure event_ids are sorted
+        sort_idx = np.argsort(unique_event_ids)
+        unique_event_ids = unique_event_ids[sort_idx]
+        event_rates = event_rates[sort_idx]
+        return unique_event_ids, event_rates
+    except Exception as e:
```
sstruzik, line 185: load_as_ndarray has a must_exist parameter that handle the case where the files are missing.
So I think we should just allow missing but remove the try except to avoid ignoring other potential issue that should raise exception.

oasislmf/pytools/common/data.py
```diff
@@ -141,3 +156,183 @@ def load_as_array(dir_path, name, _dtype, must_exist=True):
...
+        return np.array([], dtype=quantile_interval_dtype)
+    try:
+        data = load_as_ndarray(run_dir, filename[:-4], np.float32)
+        for q in data:
+            # Calculate interval index and fractional part
+            pos = (sample_size - 1) * q + 1
+            integer_part = int(pos)
+            fractional_part = pos - integer_part
+            intervals.append((q, integer_part, fractional_part))
+    except Exception as e:
```
sstruzik, line 217: same I would use must_exist parameter

oasislmf/pytools/common/data.py
```diff
@@ -141,3 +156,183 @@ def load_as_array(dir_path, name, _dtype, must_exist=True):
...
+            event_id, cursor = mv_read(fin, cursor, np.int32, np.dtype(np.int32).itemsize)
+            period_no, cursor = mv_read(fin, cursor, np.int32, np.dtype(np.int32).itemsize)
+            if granular_date:
+                occ_date_id, cursor = mv_read(fin, cursor, np.int64, np.dtype(np.int64).itemsize)
+            else:
+                occ_date_id, cursor = mv_read(fin, cursor, np.int32, np.dtype(np.int32).itemsize)
+            occ_map[i] = (event_id, period_no, occ_date_id)
+
+        return occ_map, date_algorithm, granular_date, no_of_periods
+    except Exception as e:
```
sstruzik, line 293: Just let the error be raised, this exception handling doesn't add any information

oasislmf/pytools/common/data.py
```diff
@@ -141,3 +156,183 @@ def load_as_array(dir_path, name, _dtype, must_exist=True):
...
+        # Sort by period_no
+        period_weights = np.sort(data, order="period_no")
+
+        # Identify any missing periods
+        expected_periods = np.arange(1, no_of_periods + 1)
+        actual_periods = period_weights['period_no']
+        missing_periods = np.setdiff1d(expected_periods, actual_periods)
+        if len(missing_periods) > 0:
+            raise RuntimeError(f"ERROR: Missing period_no in period binary file {periods_fp}.")
+    except Exception as e:
```
sstruzik, line 336: same, just let the error be raised

oasislmf/pytools/common/data.py
```diff
@@ -141,3 +156,183 @@ def load_as_array(dir_path, name, _dtype, must_exist=True):
...
+
+def read_occurrence(run_dir, filename=OCCURRENCE_FILE):
+    """Read the occurrence binary file and returns an occurrence map
+    Args:
+        run_dir (str | os.PathLike): Path to input files dir
+        filename (str | os.PathLike): occurrence binary file name
+    Returns:
+        occ_map (ndarray[occ_map_dtype]): numpy map of event_id, period_no, occ_date_id from the occurrence file
+    """
+    from .event_stream import mv_read
```
sstruzik, line 233: Because you have to import here to avoid circular import, I would put all the specific file reader in an other python file (this is a personal preference here, so as you like, totally ok if you prefer it like this).

oasislmf/pytools/common/event_stream.py
```diff
@@ -254,7 +254,7 @@ def register_streams_in(selector_class, streams_in):
         """
         main_selector = selector_class()
         stream_data = []
-        for stream_in in streams_in:
+        for file_id, stream_in in enumerate(streams_in):
```
sstruzik, line 257: I generally have tried to differentiate id from index,
I think we should call this one file_i or file_idx

sstruzik (approved):

SkylordA:
No need to merge as these changes were in the mentioned branches.

PR #1576, opened 2024-10-11
1250 support risk attaching ra basis in reinsurance

SkylordA:
### Support Risk Attaching 'RA' basis in reinsurance
Implemented #1250  as shown below. Policies which meet the RA requirement are not set to the PASSTHROUGH_PROFILE_ID, and left as NO_LOSS_PROFILE_ID.
UseReinsDates is currently not considered for checking RA, only AttachmentBasis (LO/RA).

RAImplementation

PR #1614, opened 2025-01-14
fixes error message for missing PolInceptionDate

SkylordA:
### fixes error message for missing PolInceptionDate
For RA based reinsurance introduced in 2.3.11 https://github.com/OasisLMF/OasisLMF/pull/1576
Solution is to look for both non '_x' and '_x' variants of these column numbers in the row if they exist and output them.
```python
acc_info = {
    field: row[f'{field}_x'] if f'{field}_x' in row else row[f'{field}']
    for field in RISK_LEVEL_FIELD_MAP[oed.REINS_RISK_LEVEL_ACCOUNT]
    if f'{field}_x' in row or f'{field}' in row
}
```

Should output the correct error message as shown in the example
```
Error: PolInceptionDate missing for {'PortNumber': '1', 'AccNumber': 'A11111'}, cannot use AttachmentBasis [RA]. Please check the account file
```


benhayes21 (approved):

benhayes21:
this should be back-ported to 2.3.12 too

johcarter (approved):
Tested that fix works as described

PR #1627, opened 2025-02-03
test: Improvements for test execution

sr-murthy:
@sambles I'm requesting a review for this evaluation PR with the following changes:

* Speed up unit tests, which currently seem to be sequentially run and take up to 30 minutes in the pipeline, by distributing the load across multiple (available) physical CPUs on the runner, using [pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/distribution.html), which has been added to `requirements.in`. This is implemented by using the `-n/--numprocesses auto` pytest flag combined with `--dist worksteal` to ensure the load is distributed as evenly as possible among all available physical CPUs. The use of `-n logical` would enable a load distribution over available logical CPUs, but `auto` has been preferred here. This should, ideally, reduce the overall runtime by a factor of about N, where N is the number of physical CPUs. (Not sure what the speedup factor would be for logical CPUs, due to hyperthreading issues.) An additional improvement, **not included**, is to use an environment variable to set`auto` (for physical CPU distribution) or `logical` (for logical CPU distribution), depending on the platform and/or runner. 

* A Makefile with targets to run unit tests, with JIT enabled (`make unittests_jit`) and JIT disabled (`make unittests_nojit`), plus updating the `unittest.yml` workflow to use these Makefile targets in test execution. The same Makefile targets can be used to run the tests locally in a more compact way.

* Code coverage configuration updated to include branches in measurement: this may reduce overall coverage because it exposes missing branches. This is the recommended approach.

* Other minor improvements include the display of missing source line numbers in the code coverage report (source lines in the `cov` target not covered by test cases), while skipping covered lines, with `--cov-report=term-missing:skip-covered`, setting the verbosity of test execution output to level 3 (`--verbosity=3`), and also ensuring code highlighting (`--code-highlight=yes`) and colorisation (`--color=yes`) in the output.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -107,13 +107,13 @@ jobs:
           pip uninstall ods_tools -y
           pip install ${{ needs.ods_tools.outputs.whl_filename }}
 
-      - name: run pytest (JIT enabled)
+      - name: run unit tests (JIT enabled)
         if: matrix.cfg.upload-cov != true
-        run: pytest -p no:flaky --ignore=fm_testing_tool --ignore=validation --cov-config=tox.ini --cov=oasislmf --cov-report=xml --cov-report=term
+        run: make unittests_jit
 
-      - name: run pytest (JIT disabled)
+      - name: run unit tests (JIT disabled)
         if: matrix.cfg.upload-cov == true
-        run: pytest -p no:flaky --ignore=fm_testing_tool --ignore=validation --cov-config=tox.ini --cov=oasislmf --cov-report=xml --cov-report=term --gul-rtol 1e-3 --gul-atol 1e-3
+        run: make unittests_nojit
         env:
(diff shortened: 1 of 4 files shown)


sr-murthy:
Hmm ... not sure why 47 cases failed, but the remaining 3.3K+ pass [?] The tests and test data haven't been changed, only the commands used to run them.

sambles:
This is a nice idea @sr-murthy  but i'm worried its going to cause issues in the long run. 
(no idea why its failing now for example) ~ it feels like it should all pass 

Closing this for the moment, but could be useful to capture this in an issue.

sr-murthy:
@sambles No problem, but did you see the error - it looked like a network error for the Code QL job, not sure about the other. A re-run may well have worked. The changes are all straightforward - I've implemented these in many other projects, and they help scalability and speed.

PR #1628, opened 2025-02-04
Update pages workflow

sambles:
### Update pages workflow
CI fix


sambles:
workflow only runs on merge, try it and hope for best

PR #1643, opened 2025-02-27
automate release branch prep

sambles:
### automate release branch prep

- [ ] Extract version from the name `release/<version>`  and update the package to match 
- [ ] auto open pull request formatted in a fixed way, so title '### Release <version>'   targeting the correct stable branch.


sambles:
look at some other time

PR #1648, opened 2025-03-06
in fm add polnumber information to level with no terms

sstruzik:
### Fix create_financial_structure issue in fmpy when last fm level has no terms
When last level of insurance has terms on only a part of the accounts and there is RI
An error occur during the creation of fmpy financial files (create_financial_structure)
We now keep track correctly the PolNumber used to merge with re-insurance.
The value to check if there is a term or not is also fixed, we now consider there is a term if the value is different from the default.
In particular, layer participation set to 0 where ignored and no policy were apply. This will change insurance loss output where this term is present


benhayes21 (approved):

sambles:
Overriding missing review for release.

PR #1649, opened 2025-03-11
Non-NumBa Error Logging

Ha-Ree:
Added functionality to disable numba upon finding an error and rerunning the command causing the error to get a better error message with more details


oasislmf/execution/runner.py
```diff
@@ -109,6 +110,43 @@ def custom_get_getmodel_cmd(
...
+    ktools_pipeline = ''
+
+    with open("run_ktools.sh", "r") as bash_script:
+        for line in bash_script:
+            if "( ( eve" in line:
+                ktools_pipeline = line.split('|')
+                break
+
+    gul_cmd = [cmd.strip() for cmd in ktools_pipeline if cmd.strip().startswith(('gul'))].pop(0)
+    fm_cmd = [cmd.strip() for cmd in ktools_pipeline if cmd.strip().startswith(('fm'))].pop(0)
```
sstruzik, line 136: there may be multiple extra fmpy when there is reinsurance, does this take into account this possibility
To produce a fake error only for ri you can add an exception in load_net_value

oasislmf/execution/runner.py
```diff
@@ -109,6 +110,43 @@ def custom_get_getmodel_cmd(
...
+    pipe_output = "/tmp/il_P1"
+    summary_output = "/tmp/il_S1_summary_P1"
+    single_event_pipe = f"{eve_cmd} | {gul_cmd} | {fm_cmd} -o {pipe_output}"
+    summary_pipe = f"summarypy -t il -m -1 {summary_output} < {pipe_output}"
+    try:
+        with open("rerun_errors.log", "w") as error_log:
```
sstruzik, line 142: in order to have intermediary output, to ease bug reproduction, can you save each intermediary output.
{eve_cmd} | {gul_cmd} > {event_id}_gul.bin
then if fmpy is present 
 {fm_cmd} -o {pipe_output} -i {event_id}_gul.bin 
 and same for the reinsurance fmpy

oasislmf/execution/runner.py
```diff
@@ -109,6 +110,43 @@ def custom_get_getmodel_cmd(
...
+        for line in bash_script:
+            if "( ( eve" in line:
+                ktools_pipeline = line.split('|')
+                break
+
+    gul_cmd = [cmd.strip() for cmd in ktools_pipeline if cmd.strip().startswith(('gul'))].pop(0)
+    fm_cmd = [cmd.strip() for cmd in ktools_pipeline if cmd.strip().startswith(('fm'))].pop(0)
+    pipe_output = "/tmp/il_P1"
+    summary_output = "/tmp/il_S1_summary_P1"
+    single_event_pipe = f"{eve_cmd} | {gul_cmd} | {fm_cmd} -o {pipe_output}"
```
sstruzik, line 139: need to also take into account if there is no fmpy just the Gul

oasislmf/execution/runner.py
```diff
@@ -109,6 +110,43 @@ def custom_get_getmodel_cmd(
...
+    summary_pipe = f"summarypy -t il -m -1 {summary_output} < {pipe_output}"
+    try:
+        with open("rerun_errors.log", "w") as error_log:
+            subprocess.run(single_event_pipe, shell=True, env=env, stderr=error_log)
+        with open("summary_errors.log", "w") as error_log:
+            subprocess.run(summary_pipe, shell=True, env=env, stderr=error_log)
+    finally:
+        env['NUMBA_DISABLE_JIT'] = "0"
```
sstruzik, line 147: the env is a copy, so I don't think you need that line

oasislmf/pytools/common/event_stream.py
```diff
@@ -337,33 +338,37 @@ def read_event(self, stream_in, main_selector, stream_selector, mv, byte_mv, cur
...
                     cursor = 0
-                return event_id, cursor, valid_buff
-            else:
-                mv[:valid_buff - cursor] = mv[cursor: valid_buff]
-                valid_buff -= cursor
-                cursor = 0
+
+        except Exception as e:
+            self.logger.error(str(e))
+            raise OasisException("Error reading stream", e)
```
sstruzik, line 371: I think we want a subclass of OasisException , OasisStreamException to be sure this is specific.
otherwise later on in the try except it means the loop inside pytool cannot raise OasisException

oasislmf/execution/runner.py
Ha-Ree: Are you sure this works? I threw a 
`return 0/0`
In load_net_value and it wont throw an error so I don't know if its called?

Ha-Ree: Turns out I wasn't using my local oasislmf  and wondering why I couldnt force errors🤦‍♂️ ignore

sstruzik (approved):

sambles (approved):

Issue #1651, opened 2025-03-11
Broken tests after hypothesis package update


sambles:
Tests failing in **tests/utils/test_data.py** after
`hypothesis==6.127.2`   -->  `hypothesis==6.128.2`

```
============================================================= short test summary info =============================================================
FAILED tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file__set_required_cols_option_and_use_defaults_for_all_other_options - AssertionError: DataFrame.iloc[:, 0] (column name="str_col") are different
FAILED tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file_with_mixed_case_cols__set_lowercase_cols_option_to_false_and_required_cols_option_and_use_defaults_for_all_other_options - AssertionError: DataFrame.iloc[:, 0] (column name="STR_COL") are different
FAILED tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file_with_mixed_case_cols__set_required_cols_option_and_use_defaults_for_all_other_options - AssertionError: DataFrame.iloc[:, 0] (column name="str_col") are different
FAILED tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file_with_mixed_case_cols__set_sort_cols_option_on_single_col_and_use_defaults_for_all_other_options - AssertionError: DataFrame.iloc[:, 0] (column name="str_col") are different
FAILED tests/utils/test_data.py::TestGetDataframe::test_get_dataframe__from_csv_file_with_mixed_case_columns___set_lowercase_cols_option_to_false_and_use_defaults_for_all_other_options - AssertionError: DataFrame.iloc[:, 0] (column name="STR_COL") are different
============================================= 5 failed, 44 passed, 3 skipped, 364 warnings in 27.84s ==============================================
```

PR #1652, opened 2025-03-11
Fix workaround for failing data null string tests 

sambles:
### Fix workaround for failing data null string tests 
fix for https://github.com/OasisLMF/OasisLMF/issues/1651


Ha-Ree (approved):
LGTM 🚀

sstruzik (approved):

PR #1654, opened 2025-03-18
Improve loading and memory usage of Aggregate Vulnerability

sstruzik:
…nerability

### Improve loading and memory usage of Aggregate Vulnerability
During the loading of the aggregate weight, all possible combination of areaperil and vulnerability where loaded.
This change use the item areaperil and vulnerability as filter to only load relevant weight

note:
Also convert the loading function to numba for extra time performance.
Locally on Bangladesh model, the total gulmc time went from more than 4 min to 9 seconds

### DIFF
--- a/oasislmf/pytools/gulmc/aggregate.py
+++ b/oasislmf/pytools/gulmc/aggregate.py
@@ -10,8 +10,6 @@ import pandas as pd
 from numba import njit
 from numba.typed import Dict, List
 from numba.types import int32 as nb_int32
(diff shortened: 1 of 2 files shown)


johcarter (approved):

johcarter:
Blending is running at 9 seconds for the 20 location portfolio, and < 5 mins for 100,000 locations with stable memory usage. Looks very good!

Ha-Ree (approved):

PR #1659, opened 2025-03-25
feature/fmtcsv-C

SkylordA:
### feature/fmtcsv-C
Experiments with custom C code to write formatted csv files to improve runtime of np.savetxt

### DIFF
--- a/oasislmf/pytools/kat/oasisnumpy.c
+++ b/oasislmf/pytools/kat/oasisnumpy.c
@@ -1,82 +1,189 @@
 #include <Python.h>
 #include <complex.h>
 #include <fcntl.h>
+#include <listobject.h>
 #include <numpy/arrayobject.h>
+#include <stdarg.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


SkylordA:
Putting some notes about implementation/ideas for reference below:

NOTE 22/04/25: Using the numpy C iterator api is not faster than using for loops for savefmttxt for non-structured arrays, and probably would not be faster for structured arrays as well.

NOTE 22/04/25: Using the fmtarray function to format one row at a time is slower than savefmttxt to format all rows in C. Have not tested with iterators or structured arrays, but I would expect a similar runtime based on savefmttxt/iterator runtimes.


CHANGING HOW CASTING IS CHECKED:
current version checks structured/named arrays for the np.dtype, and non-structured ones for just their overall type. However this fails to format into a string properly the same way that python would for the non-structured version, as if we had an array of floats, casting to an int string would always give the wrong value. We therefore need to check and cast the type based on the formatted string, not the np array dtype. This might be possible to do in python and passed to C.

23/04/25
kcachegrind shows ~88% of time spend in cast_value, with calls to snprintf. Tried other variations of printf but all take same amount of time.

PR #1668, opened 2025-03-31
feature/pytools-optimisation

SkylordA:
### feature/pytools-optimisation
Optimisations to pytools functions to improve runtime or memory.


oasislmf/pytools/plt/manager.py
```diff
@@ -304,67 +304,66 @@ def _get_mean_and_sd_loss():
                 cursor += oasis_float_size  # Read extra 0 for end of record
                 # Update MPLT data (sample mean)
                 if state["compute_mplt"]:
-                    filtered_occ_map = occ_map[occ_map["event_id"] == event_id]
                     firsttime = True
+                    if event_id in occ_map:
```
sstruzik, line 308: Does it make sense to check "event_id in occ_map" when reading the event id and go into a "skip loss" state if it is not in the occ_map?
"skip loss" meaning I skip until sidx==0

sstruzik (approved):

oasislmf/pytools/plt/manager.py
SkylordA: So when event_id not in occ_map, you still update certain state variables when sidx != 0 on line 396 and 423-429. And when sidx == 0 you call _reset_state() which also sets "reading_losses" to False to read a new event_id, which basically is what you mean by a "skip loss" state, since you only get sidx == 0 at the end of the record. Ktools also does this "skip loss" implicitly by the looks of it.

I can move the event_id check at the top of "reading losses" and add a "skip loss" loop, which would stop updating state variables unnecessarily, but I think it might be better not to add a for loop inside a while loop and possibly have to duplicate the valid_buff check. Let me try something else quickly and I'll push it up soon

SkylordA: See changes [here](https://github.com/OasisLMF/OasisLMF/pull/1668/commits/6baa7037813e90d960e247e3f28b6252cb9aa73b)

PR #1672, opened 2025-04-07
eltpy sum loss and sum loss square need to be float64 to avoid to much precision error

sstruzik:
### Increase eltpy precision to avoid too much rounding error in standard deviation
Standard deviation is calculated using the difference between two squared sum. This exacerbate the relative error, and more precision is needed on intermediary values to keep the rounding error to a more predictable value.

For example, in our Piwind example one of the absolute error was 369 monetary unit
 12786.774414 vs 13155.693359

May impact Eltpy standard deviation result


SkylordA:
I'm fine having this change, but just as a note ktools eltcalc does use OASIS_FLOAT to define these accumulators which can be either float32 (float) or float64(double). If we always want to default to float64 for these variables then it's fine.

Definition of [OASIS_FLOAT](https://github.com/OasisLMF/ktools/blob/a3c17d620a57f66208802f823647449f610e43a4/src/include/oasis.h#L71)
```cpp
//#define OASIS_FLOAT_TYPE_DOUBLE
#ifdef OASIS_FLOAT_TYPE_DOUBLE
typedef double OASIS_FLOAT;
#else
typedef float OASIS_FLOAT;
#endif
```

Definition of eltcalc [accumulators](https://github.com/OasisLMF/ktools/blob/a3c17d620a57f66208802f823647449f610e43a4/src/eltcalc/eltcalc.cpp#L328)
```cpp
OASIS_FLOAT sumloss = 0.0;
OASIS_FLOAT sample_mean = 0.0;
OASIS_FLOAT analytical_mean = 0.0;
OASIS_FLOAT sd = 0;
OASIS_FLOAT sumlosssqr = 0.0;
```

SkylordA (approved):

benhayes21 (approved):

PR #1676, opened 2025-04-14
Simplify gulmc

sstruzik:
### reorganize gulmc
reorganize gulmc to make the code simpler to understand and improve the order of each step of the calculation. Found several issues in gulmc that are being fixed
it now:
- take into account intensity adjustment before the effective damageability cdf is calculated.
- apply the intensity adjustment on aggregate vulnerability
- use equi-weight as default for aggregate vulnerability if no vuln_id is present in the weight file for this aggregate
- fix an issue with aggregate vulnerability sampled loss when some vuln id have no probability of loss
- fix gulmc hazard cdf when both hazard and vuln corellation are present
- improve performance of get_corr_rval using precomputed factor

this change improve the performance of gulmc in particular for correlated run:
-> it cut ~25% of total time on pywind 10K loc
-> gulmc with correlation on is almost 2x faster (54s-> 29s)
-> gulmc with correlation off is around 20% faster (33s-> 27s)

**This PR has an impact on the loss for several reason**
- the "aggregate vulnerability fix" will reduce the loss of aggregate vulnerability when one of the sub vulnerability has no damage
- due to the issue related to "gulmc hazard cdf fix", the hazard cdf was equal to the vuln cdf if both had correlation impacting losses
- intensity adjustment  is now applied on effective damageability cdf impactinf mean loss and all type 1 loss in ord_output


johcarter:
I tested the 'use equi-weight' as default feature for Bangladesh cyclone with
* partially complete weights file
* empty weights file (header only)

The full weights file is all equally weighted so I expected to get the same results for the above cases, and I did (AAL agrees to about 5 s.f)

The only thing that doesn't work is a completely missing weights file which generates a runtime error, but it is reasonable to expect it to be present if aggregate_vulnerability is present.

SkylordA (approved):

benhayes21 (approved):

PR #1683, opened 2025-04-23
Add parquet support to new pytools

SkylordA:
### Add parquet support to new pytools
Adds parquet output option to new pytools - ELTpy, PLTpy, AALpy, LECpy, KATpy.

### DIFF
--- a/oasislmf/pytools/aal/cli.py
+++ b/oasislmf/pytools/aal/cli.py
@@ -4,11 +4,7 @@ import argparse
 import logging
 
 from . import manager, logger
-
-
-def validate_flags(args):
-    if args.binary and args.parquet:
-        raise RuntimeError("Cannot output both parquet and binary flags at the same time.")
+from .data import VALID_EXT
 
 
 def main():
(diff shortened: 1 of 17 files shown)


oasislmf/pytools/aal/manager.py
```diff
@@ -901,20 +918,48 @@ def run(run_dir, subfolder, aal_output_file=None, alct_output_file=None, meanonl
                     sample_size,
                     file_data["no_of_periods"],
                 )
-            write_ndarray_to_fmt_csv(output_files["aal"], np.array(aal_data, dtype=AAL_dtype), AAL_headers, AAL_fmt)
-        if output_alct:
+            aal_data = np.array(aal_data, dtype=outmap["aal"]["dtype"])
```
sstruzik, line 921: One suggestion to avoid those ifs
in the init phase create a write function for each output format that take out_data and outmap[<type>] and do the writing.
for example for parquet:
```
def write_out(out_data, out_type):
    data_df = pd.DataFrame(out_data)
    data_table = pa.Table.from_pandas(data_df)
    out_type["file"].write_table(data_table)
```
then to write you just call write_out(aal_data, outmap["aal"])

oasislmf/pytools/aal/cli.py
```diff
@@ -6,6 +6,11 @@
 from . import manager, logger
 
 
+def validate_flags(args):
+    if args.binary and args.parquet:
+        raise RuntimeError("Cannot output both parquet and binary flags at the same time.")
```
sstruzik, line 11: If we only allow one output, can we have just one arg for example [-f, --output-format] that take csv, bin or parquet as value to decide on the format of the output?
It would differ from the c++ version but we already have a specific command I think its ok

benhayes21 (approved):

PR #1677, opened 2025-04-15
fix gulmc hazard cdf when both hazard and vuln corellation are present

sstruzik:
### fix gulmc hazard cdf when both hazard and vuln corellation are present
The same cdf buffer was use for hazard and vuln cdf when both had correlation, this fix the issue by using different buffer

Analysis with both hazard and damage/vulnerability correlation will have their loss impacted

PR #1686, opened 2025-04-24
Backport (1.28.x) ~ fix gulmc hazard cdf when both hazard and vuln corellation are presen…

sambles:
### fix gulmc hazard cdf when both hazard and vuln corellation are present
The same cdf buffer was use for hazard and vuln cdf when both had correlation, this fix the issue by using different buffer

Analysis with both hazard and damage/vulnerability correlation will have their loss impacted

> Changes from: https://github.com/OasisLMF/OasisLMF/pull/1677


sambles:
CI failing due to branch issues (merge and test from release branch)

PR #1687, opened 2025-04-24
Backport 1.28.x ~ add keys in error for loc_id not modelled (#1667)

sambles:
### Fix error when no peril is modelled from peril covered
When checking that all loc_id where processed after lookup, an error was arising when an exposure had only peril covered that was not modelled.
This change adds the loc_id to the error key file with message:
perils_covered + " have no perils modelled"

> from https://github.com/OasisLMF/OasisLMF/pull/1667


sambles:
CI failing due to branch issues (merge and test from release branch)

PR #1690, opened 2025-05-09
support uploading byte stream

vinulw:
### Support uploading file via byte stream 
- Update platform api upload function to allow byte stream in addition to file path.


sambles (approved):
Only a minor request ~   it would be good to have an example of the `upload_data` dict in the doc string but not a big deal, just for clarity 

```
upload_portfolio_file(<portfolio_id>, "location_file", {'bytes': <byte_stream>, 'name': <filename>})
```

PR #1693, opened 2025-05-16
Feature/conversion tools

SkylordA:
### csv to bin and bin to csv conversion tools
For ktools input file conversion tools
Also includes new conversion tools for pytools outputs, including parquet to bin and bin to parquet


oasislmf/preparation/il_inputs.py
```diff
@@ -908,7 +908,7 @@ def write_fm_policytc_file(il_inputs_df, fm_policytc_fp, chunksize=100000):
     """
     try:
         fm_policytc_df = il_inputs_df.loc[(il_inputs_df['agg_id'] > 0) & (il_inputs_df['level_id'] > 0),
-                                          ['layer_id', 'level_id', 'agg_id', 'profile_id', 'orig_level_id']]
+                                          ['level_id', 'agg_id', 'layer_id', 'profile_id', 'orig_level_id']]
```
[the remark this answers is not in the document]
SkylordA: Because ktools reads and writes the binary data in order of ["level_id", "agg_id", "layer_id", ...] which is the order of the struct defined in [oasis.h](https://github.com/OasisLMF/ktools/blob/b4474e85a58228330dad1e26ca15a6d8321ec001/src/include/oasis.h#L225).

But when reading from or writing to csv it uses the order ["layer_id", "level_id", "agg_id", ...], and I'm pretty sure this is the only file where the order inconsistent. For example the [fmpolicytctobin](https://github.com/OasisLMF/ktools/blob/b4474e85a58228330dad1e26ca15a6d8321ec001/src/fmpolicytctobin/fmpolicytctobin.cpp#L60) reads the csv in the order instead, but writes in the original struct order.

I just changed the csv ordering in the pytools code and test files to be the same as in the bin order because when I was first writing this tool I got tripped up by the different column order, and I can't find any valid reason why the order is different only for this file.

tests/pytools/converters/test_converters.py
vinulw, line 1: its a matter of taste but could use `@pytest.mark.parametrize` to avoid repeatedly calling `case_runner`

oasislmf/computation/data/dummy_model/generate.py
```diff
@@ -502,7 +502,7 @@ def generate_data(self):
             bin_from (float): damage bin lower limit.
             bin_to (float): damage bin upper limit.
             interpolation (float): damage bin midpoint.
-            interval_type (int): interval_type (deprecated).
+            damage_type (int): damage_type (deprecated).
```
vinulw, line 505: damage tyupe is not deprecated anymore, it's used to set the `type` of the bins, see https://github.com/OasisLMF/OasisLMF/pull/1721 for the different types available. For the tests its fine to leave it as generating `0` (this is a default value which interprets damage type as `1` if the bin is less that 1 or `2` if the bin is greater than 1)

validation/reorder_csv_columns.py
vinulw, line 1: Is this needed?

oasislmf/preparation/il_inputs.py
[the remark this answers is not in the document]
vinulw: okay no worries, i think it this should be fine it just caused a load of the `fm_polictc.csv` files to change in the tests

validation/reorder_csv_columns.py
SkylordA: not really, I just had it as a helper script when I was updating the fm_policytc.csv files. I can remove it if you like or just leave it there for future use if needed.

oasislmf/computation/data/dummy_model/generate.py
SkylordA: forgot to correct the comment there, will update it, thanks

vinulw: great 👍🏾 , might be worth adding a warning in the damagebintobin method if the damage bin type is not one of the above?

validation/reorder_csv_columns.py
vinulw: i dont think its needed, but not sure if @benhayes21 has a preference?

oasislmf/pytools/converters/bintocsv/utils/vulnerability.py
```diff
@@ -0,0 +1,96 @@
...
+            raise OasisException(f"Error: Cannot read zip files without provided idx_file_in zip path, currently {idx_file_in}")
+        data = _read_bin_idx_zip(file_in, idx_file_in, dtype)
+    else:
+        if idx_file_in is not None:
+            data = _read_bin_idx(file_in, idx_file_in, dtype)
+        else:
+            data = _read_bin(file_in, dtype)
+
+    num_rows = data.shape[0]
+    buffer_size = 1000000
```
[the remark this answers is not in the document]
SkylordA: I agree, a lot of the existing pytools code also had similar buffer_size stuff so I'll update those too, I'll add a DEFAULT_BUFFER_SIZE var to common/data.py for now.

tests/pytools/converters/test_converters.py
SkylordA: I think for these tests it's a tiny bit better to leave it as is instead of using parametrize

oasislmf/computation/data/dummy_model/generate.py
SkylordA: would I be correct in assuming that the damage_type can only be {0, 1, 2, 3}? I can add this as a check to the _validate function for csvtobin/utils/damagebin.py

SkylordA: or rather just check they are {1, 2, 3}

PR #1710, opened 2025-06-05
fix/correct missing `StepNumber` in test case

vinulw:
### fix/correct missing `StepNumber` in test case
- Needed for fix to https://github.com/OasisLMF/ODS_OpenExposureData/issues/247


sstruzik:
done in https://github.com/OasisLMF/OasisLMF/pull/1713

PR #1715, opened 2025-06-05
fix/lecpy `bin` tests assert equality with rounding

vinulw:
### fix/lecpy `bin` tests assert equality with rounding
- comparing bytes of binary files does not account for rounding
- loads `bin` files as dataframes instead to confirm equality

### DIFF
--- a/tests/pytools/lecpy/test_lecpy.py
+++ b/tests/pytools/lecpy/test_lecpy.py
@@ -4,6 +4,7 @@ import pandas as pd
 from tempfile import TemporaryDirectory
 
 import numpy as np
+from oasislmf.pytools.lec.data import EPT_dtype
 from oasislmf.pytools.lec.manager import main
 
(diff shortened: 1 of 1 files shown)


tests/pytools/lecpy/test_lecpy.py
```diff
@@ -90,8 +92,9 @@ def case_runner(sub_folder, test_name, out_ext="csv", use_return_period=False):
                 actual_psept_data = pd.read_parquet(actual_psept)
                 pd.testing.assert_frame_equal(expected_psept_data, actual_psept_data)
             if out_ext == "bin":
-                with open(expected_psept, 'rb') as f1, open(actual_psept, 'rb') as f2:
-                    assert f1.read() == f2.read()
+                expected_ept_data = pd.DataFrame(np.fromfile(expected_ept, dtype=EPT_dtype))
```
SkylordA, line 95: These should use expected_psept and actual_psept as well as the PSEPT_dtype

SkylordA (changes requested):

SkylordA (approved):

PR #1718, opened 2025-06-12
remove empty cdf bin

sstruzik:
### fix issue in gul calulation leading to np.inf loss
inf loss in the gul loss calculation hapend when the probability of the bin is 0, Normally this doesn't happen because the bin wouldn't be selected if the prob is 0.
However when it is the right side of the cdf, this is not guaranteed, particularly if the final probability is less that 1.
To remedy that, this fix remove the all the bin in the right end side  of cdf that have a probability of 0.

### DIFF
--- a/oasislmf/pytools/gulmc/manager.py
+++ b/oasislmf/pytools/gulmc/manager.py
@@ -415,8 +415,8 @@ def run(run_dir,
             norm_inv_cdf, norm_cdf = np.zeros(1, dtype='float64'), np.zeros(1, dtype='float64')
 
         # buffer to be re-used to store all the correlated random values
-        vuln_z_unif = np.zeros(sample_size, dtype='float64')
-        haz_z_unif = np.zeros(sample_size, dtype='float64')
+        vuln_z_unif = np.zeros(sample_size, dtype='float32')
+        haz_z_unif = np.zeros(sample_size, dtype='float32')
 
         if effective_damageability is True:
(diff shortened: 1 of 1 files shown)


oasislmf/pytools/gulmc/manager.py
```diff
@@ -595,6 +595,23 @@ def get_haz_cdf(item_event_data, haz_cdf, haz_cdf_ptr, dynamic_footprint, intens
     return haz_cdf_prob, haz_cdf_bin_id
 
 
+@nb.njit(fastmath=True)
+def get_last_non_empty(cdf, bin_i):
```
SkylordA, line 599: is the intention of this function to just remove values that are equal to 0 from the right side of the array, or is it to remove all non changing values from the right. Or is it to do both.
```
- So for this it would remove all but the last 0 element = [0.1, 0.2, 0.3, 0.4, 0.0, 0.0, 0.0] -> [0.1, 0.2, 0.3, 0.4, 0.0]
- And for this it would remove duplicate values at the end = [0.1, 0.3, 0.6, 0.6, 0.6, 0.6] -> [0.1, 0.3, 0.6]
```

sstruzik: As this is a cumulative distribution in the input, what I mean by "empty bucket" is to have the same value as the one before. In that case, the corresponding probability distribution is 0.

SkylordA (approved):

benhayes21 (approved):

Issue #1705, opened 2025-06-02
damage_bin_dict redundant interval_type should be removed


SkylordA:
## Issue Description
interval_type is not actually used anywhere in oasislmf and is redundant. It is currently just set to a default of 0 when writing binary files and should ideally be entirely removed

## Example data / logs

Comments in this function suggest it is deprecated
https://github.com/OasisLMF/OasisLMF/blob/9ecb12b71d00481ac19c987f4d98918deda4652f/oasislmf/computation/data/dummy_model/generate.py#L522

SkylordA:
Solved by https://github.com/OasisLMF/OasisLMF/pull/1721
interval_type is now changed to damage_type and used in code.

benhayes21:
closing - done

PR #1719, opened 2025-06-12
feature/damage_type column to damage_bin_dict.csv

vinulw:
### feature/damage_type column to damage_bin_dict.csv
- Adds a column called `damage_type` which distinguishes between relative and absolute bin intervals in the `damage_bin_dict` 
- Remove deprecated `interval_type` from the `get_damage_bins`


vinulw:
may partially resolve #1705

PR #1720, opened 2025-06-13
FIX: type inconsistencies when using pytools and ktools with 8 byte flags

MrQQish:
### FIX: type inconsistencies when using pytools and ktools with 8 byte flags

Fix multiple inconsistencies between KTools and OasisLMF when using user defined types 
specifically areaperil_int=u8 and oasis_float=u8 (required for the COMBUS models).  

Also highlighted potential areas of inconsistencies. Alot of these types have been caught
in a bug driven manner but we hope with this final review and expanded internal testing
the last have them have been squashed. 

However in the interest of completeness, would it be possible to request that user defined 
types are tested as part of the oasis test suite. This would cover changes in oasis_int (which
we dont test or modify) and prevent future development of from introducing bugs. 

This can treated separately in a future issue at your convenience.


sstruzik:
Thanks for looking at this and cleaning it up
I don't know why our pep8 CI is unhappy but a simple
`autopep8 --exit-code --recursive --max-line-length 150 --ignore E402 ./oasislmf --in-place`
in the root dir should fix the issue

as you pointed out:
- in oasislmf/pytools/common/data.py, the coverage_dtype variable is not used
- we have two identical coverage_type in gul and gulmc common
-  I think coverage_dtype is a better name for this variable as coverage_type sometime means something else (Building, Content, ...)

Therefore, can you:
- remove common data coverage_dtype and replace it with the definition of gul and gulmc
- replace the TIV dtype with oasis_float
- make gul and gulmc point to the new definition for coverage_dtype that will be then located in oasislmf/pytools/common/data.py

MrQQish:
``` python
# tiv field changed back to np.float64 instead of oasis_float.
# oasis_float type failing binary comparisons for output losses
# E Name: loss, dtype: float64
# E x: expected     E y: test
# E 60 102666.66    E 60 102666.67
# this is because by default oasis uses np.float32 but the tiv
# is defined as np.float64 in the oasislmf codebase thus leading
# to a precision error when reverting to defining tiv as
# oasis_float float32
```

I have switched tiv type back to float64, but in order to make the
change to oasis_float im guessing that the test results need to
updated? Potentially will require further type changes to other 
loss calculation types downstream of the coverage TIV.

MrQQish:
Any updates? This would be good to have by the next release.

PR #1740, opened 2025-07-18
support for mac M2 multiprocessing lookup

sstruzik:
### support for mac m2 in lookup multiprocessing
Apply suggested change to support mac m2 architecture in multiprocessing lookup.
- use multiprocessing.get_context("fork") to create a context object that will create the queue and process
- make target function module function instead of static method to avoid pickle error.


sambles (approved):
Attempted to test on M2 mac in AWS, but couldn't get the allocation to work.  (not tested)

Looks good from a Linux perspective

benhayes21 (approved):

PR #1745, opened 2025-07-25
feature/BI improvement FM

vinulw:
### feature/BI improvement - fmp
A new `fm` level `SiteItem` applies `BIWaitingPeriod` and `BIPOI` (period of indemnity). 
- This new level is applied before the `SiteCoverage` level. 
- Output loss from this level is calculated as `output_loss = min(input_loss - BIWaitingPeriod, BIPOI)`.
- For multi peril the terms are in accordance with the `alloc_rule_il`, by default allocation will occur proportional to the input loss.


oasislmf/utils/defaults.py
```diff
@@ -225,7 +225,9 @@ def get_default_unified_profile(path=False):
 
 def get_default_fm_aggregation_profile(path=False):
     fp = os.path.join(STATIC_DATA_FP, 'default_fm_agg_profile.json')
-    return {int(k): v for k, v in get_default_json(src_fp=fp).items()} if not path else fp
```
sstruzik, line 228: I don't understand the point of path in the original function and looking at the code path is never used as a parameter.
I think we can clean this and remove it. this way we have get_default_fm_aggregation_profile that return the default profile.

vinulw: the path variable seems to be redundant in these functions as well: 
- `get_default_accounts_profile`
- `get_default_exposure_profile`
- `get_default_fm_profile_field_values`
- `get_default_step_policies_profile`
- `get_config_profile`
- `get_default_unified_profile`

PR #1747, opened 2025-07-30
Progress bar

Ha-Ree:
See #1739


oasislmf/computation/generate/losses.py
```diff
@@ -554,7 +559,7 @@ def run(self):
                     ))
                 else:
                     self.logger.info('All {} Loss chunks generated in {}'.format(bash_params['max_process_id'], model_run_fp))
-
+                oasis_ping({'analysis_pk': bash_params['analysis_pk'], 'events_total': str(os.path.getsize("input/events.bin") / 4)})
```
vinulw, line 562: use size of `oasis_int` from `oasislmf.pytools.common.data` instead of 4 here incase we change dtype in future

sstruzik (approved):

PR #1752, opened 2025-08-06
Port/pandas 2.2.0 fix    (backport 2.3.x) 

sambles:
### Make OasisLMF compatible with pandas > 2.2   
In pandas 2.2, the ordering of merge operation is different from the one in 2.1. This leads to error in our test suite. Those error are actually red hearing because the result themselves are still correct.
This PR adapt the code to give the same result between the two versions.


sambles:
not worth it.

vinulw (approved):

PR #1764, opened 2025-09-04
Enhancement/add settings to lookup and hooks

sstruzik:
### add settings information to lookup and hooks
- make settings information available in lookup and hooks.

- use OasisDataManager in Built-in lookup.

### DIFF
--- a/oasislmf/lookup/builtin.py
+++ b/oasislmf/lookup/builtin.py
@@ -431,7 +431,7 @@ class Lookup(AbstractBasicKeyLookup, MultiprocLookupMixin):
         placeholder_keys = set(re.findall(r'%%(.+?)%%', filepath))
         for placeholder_key in placeholder_keys:
             filepath = filepath.replace(f'%%{placeholder_key}%%', self.config[placeholder_key.lower()])
-        if "keys_data_path" in placeholder_keys:
+        if "keys_data_path" in [key.lower() for key in placeholder_keys]:
             return filepath
         else:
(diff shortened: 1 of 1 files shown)


oasislmf/lookup/builtin.py
```diff
@@ -431,7 +431,10 @@ def to_abs_filepath(self, filepath):
         placeholder_keys = set(re.findall(r'%%(.+?)%%', filepath))
         for placeholder_key in placeholder_keys:
             filepath = filepath.replace(f'%%{placeholder_key}%%', self.config[placeholder_key.lower()])
-        return filepath
+        if "keys_data_path" in placeholder_keys:
```
vinulw, line 434: "KEYS_DATA_PATH" might be capitalised in placeholder keys.

oasislmf/lookup/factory.py
```diff
@@ -161,7 +161,7 @@ def create(
             lookup_config=None,
             lookup_config_json=None,
             lookup_config_fp=None,
-            complex_lookup_config_fp=None,
+            complex_lookup_config_fp=None,  # this is now use to pass the run settings if there is no complex_lookup_config_fp
```
vinulw, line 164: should the warning message in `update_derecated_args` be updated if `complex_lookup_config_fp` is expected to be used in this way?

oasislmf/lookup/builtin.py
sstruzik: good catch, fixed.

PR #1770, opened 2025-09-23
feature/evepy

vinulw:
### feature/evepy
- Implement `eve` in python as a new cli tool `evepy` 
- Maintains the same behaviour as `eve` with a round robin shuffle by default and the following command line options: 
- - `-n` : no shuffle 
- - `-r` : randomise using Fisher Yates algorithm and stream output whilst randomising 
- - `-R`: randomise using builtin shuffle 

Also support input and output file selection using the `-i` and `-o` flags respectively.


oasislmf/pytools/eve/manager.py
```diff
@@ -0,0 +1,191 @@
...
+def partition_events__no_shuffle(events, process_number, total_processes):
+    """Assign events in the order they are loaded to each process in turn. Only
+    output the event IDs allocated to the given `process_number`.
+
+    Args:
+        events (np.array): Array of ordered event IDs.
+        process_number (int): The process number to receive a partition of events.
+        total_processes (int): Total number of processes to distribute the events over.
+    """
+    # TODO - check #-events < total_processes
```
SkylordA, line 54: is this TODO meant to be left here?

PR #1797, opened 2025-11-19
give ping a port if default missing

Ha-Ree:
Ping sometimes fails if port taken: give it a free port if default unavailable

### DIFF
--- a/oasislmf/utils/socket_server.py
+++ b/oasislmf/utils/socket_server.py
@@ -13,6 +13,7 @@ class GulProgressServer:
     def __init__(self, total, host=None, port=None):
         """
         Args:
+            total (int): How many events the server has
             host (str, optional): Non-default host to use. Defaults to `OASIS_SOCKET_SERVER_IP` or SERVER_DEFAULT_IP if unset.
             port (int, optional): Non-default port to use. Defaults to `OASIS_SOCKET_SERVER_PORT` or SERVER_DEFAULT_PORT if unset.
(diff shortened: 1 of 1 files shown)


oasislmf/utils/socket_server.py
```diff
@@ -26,7 +27,14 @@ def start(self):
         self.running = True
         self.server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.server_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        self.server_socket.bind((self.host, self.port))
+
+        try:
+            self.server_socket.bind((self.host, self.port))
+        except OSError:  # Desired socket unavailable
```
vinulw: Might be worth adding a message to the logger with the new `OASIS_SOCKET_SERVER_PORT` info

oasislmf/utils/socket_server.py
```diff
@@ -6,17 +6,19 @@
 import sys
 from tqdm import tqdm
 from oasislmf.utils.defaults import SERVER_UPDATE_TIME, SERVER_DEFAULT_PORT, SERVER_DEFAULT_IP
+import logging
 
 
 class GulProgressServer:
-    def __init__(self, host=None, port=None):
+    def __init__(self, total, host=None, port=None):
```
sstruzik: Can you add info on what total is

PR #1826, opened 2025-12-22
Enhancement/platform roe

Ha-Ree:
Changing api run to send inputs for roe


oasislmf/platform_api/client.py
```diff
@@ -491,11 +495,27 @@ def upload_portfolio_file(self, portfolio_id, portfolio_file, upload_data):
                                                                  upload_data['name'])
             self.logger.info("File uploaded: {}".format(upload_data['name']))
         else:
```
vinulw: note that the pythonui uses `upload_byte` to upload files. So if the currency conversion json cannot be uploaded using this mechanism it won't be possible to upload it in the UI

oasislmf/platform_api/client.py
```diff
@@ -517,10 +537,31 @@ def upload_inputs(self, portfolio_name=None, portfolio_id=None,
...
                 self.upload_portfolio_file(portfolio_id, 'reinsurance_scope_file', ri_scope_fp)
+            if currency_conversion_fp:
+                self.upload_portfolio_file(portfolio_id, 'currency_conversion_json', currency_conversion_fp)
+            if reporting_currency:
+                self.upload_portfolio_reporting_currency(portfolio_id, 'reporting_currency', reporting_currency)
             return portfolio.json()
         except HTTPError as e:
             self.api.unrecoverable_error(e, 'upload_inputs: failed')
 
+    def get_currency_conversion(self, filepath):
```
vinulw: this might be good to put in a `util` module as it could be useful in other parts of the code and is not really using any class information

sambles (approved):

PR #1835, opened 2026-01-16
Bashgen fix - swap rl for ri when running summaypy 

sambles:
### Bashgen fix - swap rl for ri when running summaypy 
Fix for running `"rl_output": true`  in analysis settings  with `"summarypy": true`


sambles:
Merged from here https://github.com/OasisLMF/OasisLMF/pull/1851

PR #1857, opened 2026-02-03
fixup parquet key file exposure run

sstruzik:
### Summary
- Fixed `RunExposure` hardcoding `keys.csv` path, which broke parquet key file support

### Changes
- Use the keys file path returned by `GenerateKeysDeterministic().run()` instead of hardcoding `keys.csv`
- This ensures the correct format (parquet or csv) is used downstream


vinulw (approved):
👍🏾

PR #1875, opened 2026-02-13
enhancement/root `analysis_settings.json` not updated when run is complete

vinulw:
### enhancement/root `analysis_settings.json` not updated when run is complete
- make `output-dir/analysis_settings.json` match `output-dir/output/analysis_settings.json`


sstruzik:
This duplicate the analysis file in two folders. We probably should only have one, maybe with a rename.
Not sure if it worth the effort but.
Is it feasible to either?
- remove the one in output, rename the original one in run_dir (with a suffix like _input, _user or _origin)
- point the process that consume the setting after to the output one

sstruzik (approved):

PR #1901, opened 2026-03-06
Fix/chardet 7 support

sstruzik:

### DIFF
--- a/oasislmf/utils/data.py
+++ b/oasislmf/utils/data.py
@@ -56,10 +56,7 @@ import numpy as np
 import pandas as pd
 import pytz
 import chardet
-try:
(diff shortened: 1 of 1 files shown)


sstruzik:
> not sure if this really affects us, but there are some discussions going on about chardet 7 and it's licensing changes [chardet/chardet#327](https://github.com/chardet/chardet/issues/327)

If I understand correctly, the v6 version has LGPL licence and the v7 has MIT licence.  The newer version is more permissive so the original author is not "happy".
From our perspective, we are ok with both licences so it shouldn't matter even if it is reversed.
what do you think?

vinulw (approved):

PR #1920, opened 2026-03-30
enhancement/lecpy_speed

### LINKED ISSUE #1680: Improve lecpy runtime
## Issue Description

Pytool lecpy runtime is too slow, try and improve it without compromising on memory usage. A lot of the runtime is spent on the write_ept and write_ept_weighted function, especially when running full uncertainty.

SkylordA:
### enhancement/lecpy_speed

Performance improvements to the Python LEC (Loss Exceedance Curve) calculation engine.

#### Changes

**Separate `row_used` mask from `OUTLOSS_DTYPE`**
- Extracted the `row_used` boolean field out of `outloss_mean`/`outloss_sample` structured arrays into dedicated `np.bool_` memmapped files (`lec_row_used_mean.bdat`, `lec_row_used_sample.bdat`)
- `np.where` lookup is now computed once at `AggReports` initialisation and reused across all report types, avoiding repeated full-array scans

**Replace `nb.typed.Dict` tail structures with flat `ndarrays`**
- TVaR `tail` and `tail_sizes` were previously Numba typed dicts with dynamically resized array values
- Replaced with pre-allocated flat `ndarray[TAIL_valtype]` arrays indexed via pre-computed `tail_offsets`
- Eliminates dict lookups, dynamic array resizing, and Python list accumulation in `write_tvar` / `write_tvar_wheatsheaf`
- Output arrays (`EPT_dtype` / `PSEPT_dtype`) are now pre-allocated and written by position directly

**Fix `np.int32` overflow in memmap shape calculations**
- `no_of_periods` and `sample_size` are read as `np.int32` from binary headers; multiplying them to compute memmap shapes overflowed silently
- Explicit `int()` casts added before shape multiplications

**Deduplicate `row_used` update logic**
- `do_lec_output_agg_summary` previously selected target array via a branch then updated `row_used` through a shared code path
- Each branch now handles its own array independently, removing the intermediate variable and making per-array updates explicit

#### Performance improvement (using binary outputs)

image

image

closes https://github.com/OasisLMF/OasisLMF/issues/1680


oasislmf/pytools/lec/aggreports/aggreports.py
```diff
@@ -47,6 +49,8 @@ def __init__(
         self.lec_files_folder = lec_files_folder
         self.output_binary = output_binary
         self.output_parquet = output_parquet
+        self.row_used_indices_mean = np.where(row_used_mean)[0]
```
vinulw, line 52: minor: np docs suggest using `np.nonzero` if only using condition in `np.where` (https://numpy.org/doc/stable/reference/generated/numpy.where.html)

vinulw (approved):

sambles (approved):

Issue #1936, opened 2026-04-08
Timestamps on run directories are off by one hour


carlfischerjba:
Note how the model run directories in the example below all have names with a timestamp two hours behind their creation time. We would expect a one hour difference since the timestamps are UTC but I'm currently operating in British Summer Time (UTC+1).

```
$ ls -l --time=birth runs/
total 12
drwxrwxrwx 3 ... 4096 Apr  8 11:19 losses-20260408091911
drwxrwxrwx 8 ... 4096 Apr  8 11:22 losses-20260408092253
drwxrwxrwx 8 ... 4096 Apr  8 14:45 losses-20260408124541
```

This is caused by the use of non-timezone aware timestamps somewhere in the code base.

PetroczyP:
Hi — I'd like to pick this one up. I have a local repro on `TZ=Europe/Zurich` around the DST spring-forward boundary and I plan to open a PR that moves run directory naming to UTC. Happy to discuss approach first if a maintainer would prefer to weigh in on the timezone handling before I submit.

carlfischerjba:
> Hi — I'd like to pick this one up. I have a local repro on `TZ=Europe/Zurich` around the DST spring-forward boundary and I plan to open a PR that moves run directory naming to UTC. Happy to discuss approach first if a maintainer would prefer to weigh in on the timezone handling before I submit.

Hello. Already submitted a PR that should fix this (linked above) but there were some CI failures that I don't think are related to my very simple change. Feel free to comment on https://github.com/OasisLMF/OasisLMF/pull/1937 if you see a better solution.

Assume we need @sambles to take a look at the CI issues.

PetroczyP:
> > Hi — I'd like to pick this one up. I have a local repro on `TZ=Europe/Zurich` around the DST spring-forward boundary and I plan to open a PR that moves run directory naming to UTC. Happy to discuss approach first if a maintainer would prefer to weigh in on the timezone handling before I submit.
> 
> Hello. Already submitted a PR that should fix this (linked above) but there were some CI failures that I don't think are related to my very simple change. Feel free to comment on [#1937](https://github.com/OasisLMF/OasisLMF/pull/1937) if you see a better solution.
> 
> Assume we need [@sambles](https://github.com/sambles) to take a look at the CI issues.

Thanks for the heads-up!
[...]

PR #1937, opened 2026-04-08
Correct timestamps that were off by one hour

carlfischerjba:
https://github.com/OasisLMF/OasisLMF/issues/1936

### Correct timestamps in directory names
Previously, losses and input generation directories could have a timestamp that was one hour behind the actual time the model was run due to incorrect handling of timezones. These directories are now named correctly with a UTC timestamp.


carlfischerjba:
Superseded by https://github.com/OasisLMF/OasisLMF/pull/1941.

PR #1941, opened 2026-04-11
Fix #1936: correct UTC timestamps in get_utctimestamp

### LINKED ISSUE #1936: Timestamps on run directories are off by one hour

PetroczyP:
Fixes #1936. See the thread on the issue for context on #1937.

The old `get_utctimestamp` used `datetime.utcnow()` as the default — which returns a naive datetime, and `naive.astimezone(tz)` interprets naive values as local time, so the call was effectively subtracting the local offset from every run-dir timestamp on non-UTC hosts. This switches to a tz-aware default (`datetime.now(pytz.utc)`), uses a `None` sentinel to also close a mutable-default-argument issue in the old signature, and treats any naive input as UTC.

The existing `test_get_utctimestamp` asserted the function body against itself, so it passed against the buggy code. Replaced with a few cases, two of which actually fail on `main`.

### Correct UTC timestamps in run directory names
`get_utctimestamp` now produces a correct UTC wall-clock string regardless of the host's local timezone, fixing the off-by-one-hour run directory names reported in #1936.

### DIFF
--- a/oasislmf/utils/data.py
+++ b/oasislmf/utils/data.py
@@ -690,12 +690,21 @@ def get_timestamp(thedate=datetime.now(), fmt='%Y%m%d%H%M%S'):
     return thedate.strftime(fmt)
 
 
-def get_utctimestamp(thedate=datetime.utcnow(), fmt='%Y-%b-%d %H:%M:%S'):
+def get_utctimestamp(thedate=None, fmt='%Y-%b-%d %H:%M:%S'):
     """
-    Get a UTC timestamp string from a ``datetime.datetime`` object
+    Get a UTC timestamp string from a ``datetime.datetime`` object.
 
-    :param thedate: ``datetime.datetime`` object
-    :type thedate: datetime.datetime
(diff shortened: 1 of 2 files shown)


carlfischerjba:
Very thorough fix. Thanks. I'll close my PR.

SkylordA (approved):

Ha-Ree (approved):

PR #1942, opened 2026-04-14
Fix brittle PolNumber backfill in IL input preparation

mhfathom:
This changes the `PolNumber` fallback fill in OasisLMF IL input preparation to use the original dataframe index instead of relying on `output_id - 1` alignment.

The previous implementation was brittle because it assumed the merge result would stay in the same row order as the source subset. In failing cases, that assumption broke and `PolNumber` values were written back to the wrong rows or remained null.

## Problem

When `gul_inputs_df["PolNumber"]` was null, the fallback logic attempted to recover it from `accounts_df` using the policy-layer aggregation key.

The previous code:
- filtered rows with null `PolNumber`
- merged them with `accounts_df`
- used `output_id - 1` to align merged results back onto `gul_inputs_df`

That alignment strategy is unsafe because it depends on positional ordering rather than the real dataframe index.

## Change

The fallback logic now:
- isolates rows where `PolNumber` is null
- preserves the original `gul_inputs_df` row index via `reset_index()`
- performs a left merge on `default_pol_agg_key`
- restores the original row index with `set_index("index")`
- assigns recovered `PolNumber` values back using the original dataframe index

This makes the fill deterministic and robust even when merge ordering changes.

## Why this is safer

Using the real dataframe index avoids accidental row misalignment and handles single-key joins such as `["acc_id"]` cleanly.

## Scope

This change is narrowly scoped to the `PolNumber` recovery block in IL input preparation. It does not change the later `acc_idx` merge logic.

## Testing

Tested by rerunning the failing model workflow and validating that the `PolNumber` fallback no longer depends on `output_id` ordering.


sstruzik:
Pushed `1f39160b9` with two small additions:

**`.reset_index()` before `.merge()`** — the original diff called `.set_index('index')` on the merge result, but `no_polnumber_df` doesn't carry an `'index'` column (its columns are `default_pol_agg_key + ['output_id']`). Adding `no_polnumber_df.reset_index()` materialises the gul_inputs index labels as an `'index'` column so the subsequent `.set_index('index')` works. Without this, the moment any portfolio actually hits the backfill the run raises `KeyError: "None of ['index'] are in the columns"`.

**`validation/insurance_conditions/units/sc22`** — a minimal regression test that deterministically exercises the backfill path: one account (B) with two policies sharing a CondTag for sc3-style row expansion, and one account (A) with no Layer/cond terms whose rows end up with NaN PolNumber. On `main` this raises `ValueError: cannot convert NA to integer: Error while type casting for column 'acc_idx'`; with the fix it produces deterministic output. None of `sc1`–`sc20` happened to enter the `if not no_polnumber_df.empty` branch — sc22 is built specifically to enter it.

sstruzik (approved):

Ha-Ree (approved):

PR #1946, opened 2026-04-20
Add option to upload to pypi test server

sambles:
Added checkbox to push binaries to https://test.pypi.org instead of the live package repo  

- [x] revert https://github.com/OasisLMF/OasisLMF/pull/1946/commits/61cff7cb3b52d921c1777858d44c49e373025440 after test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,6 +92,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Pull latest
+      run: git pull
+
     - name: Build wheels
       uses: pypa/cibuildwheel@v2.21.3
       env:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -92,6 +92,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Pull latest
```
SkylordA, line 95: seems redundant

SkylordA (changes requested):

.github/workflows/build.yml
sambles: part of the release script it to tag the package with the targeted version,   like with https://github.com/OasisLMF/OasisLMF/pull/1946/changes/61cff7cb3b52d921c1777858d44c49e373025440. 

without out that extra pull it missed the new commit (no idea why) might be a quirk of actions

PR #1959, opened 2026-04-27
API client, added option for disabled auth

sambles:
### API client, added option for disabled auth
See: https://github.com/OasisLMF/OasisPlatform/pull/1387/  Client checks if `server_info/`  returns 200 (normally needs to be authenticated) and runs without authorization if disabled

### DIFF
--- a/oasislmf/platform_api/session.py
+++ b/oasislmf/platform_api/session.py
@@ -55,6 +55,11 @@ class APISession(Session):
         # Check connectivity & authentication
         self.health_check()
         self.retry_max = retries
+
+        if self._fetch_server_auth_type() == 'disabled':
+            self.auth_type = 'disabled'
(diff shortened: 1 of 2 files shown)


SkylordA (approved):

oasislmf/platform_api/session.py
```diff
@@ -55,6 +55,11 @@ def __init__(
         # Check connectivity & authentication
         self.health_check()
         self.retry_max = retries
+
+        if self._fetch_server_auth_type() == 'disabled':
```
SkylordA, line 59: actually possibly minor thing, should probably add `self.auth_credentials = {}` here

SkylordA (approved):
pre-approving

vinulw (approved):

PR #1960, opened 2026-04-27
Directly send port to gulmc gulpy and socket server

### LINKED ISSUE #1954: Progress bar not updating when fallback port is used
# Description
The progress bar `Gul events completed:   0%|          | 0/106423 [00:00<?, ?events/s]` remains stuck like this for the entire analysis. We're running on a machine where the default port 8888 is used by another application. 

## Expected behaviour

The underlying server/client uses a free port and the progress bar updates correctly.

## Workaround

Explicitly setting `OASIS_SOCKET_SERVER_PORT=42734` (unused port) is a workaround.

# Versions affected

`oasislmf==2.5.2`

# Diagnosis

[It looks like the Oasis socket server](https://github.com/OasisLMF/OasisLMF/blob/b2552e8a3375d5a6b003cadf389e81c7253e181b/oasislmf/utils/socket_server.py#L39) detects that the default or requested port is in use and picks a different port that is free. I believe this works correctly but the new port is not passed through to the clients running as part of the gulmc processes.

The intention was apparently to use the environment variable `OASIS_SOCKET_SERVER_PORT` to pass the new port to the clients, but my guess is that this doesn't work because the clients are running in subprocesses started via the generated `run_kernel.sh` script and do not inherit environment variables. An alternative would be to pass the server and port via a command line argument to `gulmc`. This might be clearer than the current `--socket-server` argument which sometimes takes a boolean value, sometimes a string representation of a boolean and sometimes the total number of events. I guess the environment variable could still be used but would need to be explicitly set in `run_kernel.sh` (at least).

Ha-Ree:
Avoid env vars


oasislmf/utils/ping.py
```diff
@@ -28,7 +31,9 @@ def oasis_ping(data):
             return oasis_ping_websocket(f"{os.environ['OASIS_WEBSOCKET_URL']}:{os.environ['OASIS_WEBSOCKET_PORT']}/ws/analysis-status/", msg)
         logging.error("Missing environment variables `OASIS_WEBSOCKET_URL` and `OASIS_WEBSOCKET_PORT`.")
         return False
-    target = (os.environ.get("OASIS_SOCKET_SERVER_IP", SERVER_DEFAULT_IP), int(os.environ.get("OASIS_SOCKET_SERVER_PORT", SERVER_DEFAULT_PORT)))
+    port_override = data.pop('port_override', None)
```
SkylordA, line 34: why not just `data.get` here?

oasislmf/pytools/gul/manager.py
```diff
@@ -275,7 +275,9 @@ def run(run_dir, ignore_file_type, sample_size, loss_threshold, alloc_rule, debu
 
         counter = 0
         timer = time.time()
-        ping = kwargs.get('socket_server', 'False') != 'False'
+        socket_server_val = kwargs.get('socket_server', 'False')
+        ping = socket_server_val != 'False'
+        ping_port = int(socket_server_val) if ping and str(socket_server_val).isdigit() else None
```
SkylordA, line 280: technically isdigit also accepts negative numbers and unicode chars, idk how robust the check really needs to be though, maybe just a try/except is best here. same for gulmc/manager.py

oasislmf/computation/generate/losses.py
```diff
@@ -731,6 +733,17 @@ def run(self):
         self.logger.info('Losses generated in {}'.format(model_run_fp))
         return model_run_fp
 
+    def _find_available_port(self, host, preferred_port):
```
SkylordA, line 736: Claude found this issue with potential race condition for the ports.

**Potential TOCTOU race in `_find_available_port`**

The method binds to check availability, then releases the port when the `with` block exits — before `socket-server` spawns and re-binds. Another process can grab the port in that window.

You can reproduce it with this self-contained script:

```python
import socket

def find_available_port(host, preferred_port):
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        try:
            s.bind((host, preferred_port))
            return preferred_port
        except OSError:
            s.bind((host, 0))
            return s.getsockname()[1]

HOST = '127.0.0.1'
port = find_available_port(HOST, 10006)

# Simulate another process grabbing the port in the gap between
# _find_available_port releasing it and socket-server binding to it
squatter = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
squatter.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
squatter.bind((HOST, port))

# Now simulate socket-server trying to bind — this is what fails in production
try:
    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    server.bind((HOST, port))
    print('socket-server started OK')
except OSError as e:
    print(f'socket-server failed to bind to port {port}: {e}')
finally:
    squatter.close()
```

Output:
```
socket-server failed to bind to port 10006: [Errno 98] Address already in use
```

**Suggested fix:** let `socket-server` itself bind to port 0 (OS-assigned), then print the chosen port to stdout so Python reads it back. That way the port is held continuously with no gap.

SkylordA (changes requested):

oasislmf/utils/ping.py
Ha-Ree: Could do either- just feels nicer to have it out of the message because its not really necessary info for the recieving socket

SkylordA: ah I didn't see that its removal is documented in the function info, guess this is fine then

oasislmf/pytools/gul/manager.py
Ha-Ree: The only thing that can be passed through to here in the process is either a port number or false, there is no need to validate the number is not unicode char or negative

oasislmf/computation/generate/losses.py
Ha-Ree: This is a potential problem but putting it into stdout and trying to read it back also comes with having to restructure the bash script and dealing with waits because the socket server currently gets booted in parallel with the gul commands. I think its worth looking into in the future but putting the temporary fix for the issue in because outside of malicious socket use it is very unlikely that the port is grabbed in the time between setup and use

SkylordA: ok, worth making a separate issue for it after to be solved later then.

SkylordA (approved):
just make a separate issue for the race condition comment from above

sstruzik (approved):

PR #1964, opened 2026-05-01
perf(gulmc): replace numba dicts with precomputed array-backed structures

sstruzik:
## Summary

- Replaces `numba.typed.Dict` instances throughout the ground-up loss path (`gulmc`, `getmodel`, `gulpy`) with array-backed lookups and hash-by-value maps.
- Pre-computes per-process model structures (intensity bins, areaperil/vuln keys, CDF cache keys) so worker processes load slimmer data and skip per-event dict construction.
- Introduces shared primitives in `oasislmf/pytools/common/`: `hashmap.py` (hash-by-value map) and `id_index.py` (compact id → index lookup).

## Performance impact (PiWind, gulmc, GUL+IL+RI)

Measured end-to-end with Piwind 10K `oasislmf model run` and the resource monitor report:

| Metric | main | this branch | Δ |
|---|---:|---:|---:|
| Wall time (`runner.run`) | ~310 s | ~170 s | **~45% faster** |
| gulmc CPU total | ~730 s | ~420 s | **~42% lower** |
| gulmc peak USS / instance | ~405 MB | ~310 MB | **~24% lower** |
| gulmc aggregate peak USS | ~3.0 GB | ~2.1 GB | **~30% lower** |

- Ground-up loss (`gulmc`) now runs ~45% faster end-to-end and uses ~30% less peak memory on representative workloads, by replacing numba dicts with precomputed array-backed structures.


oasislmf/pytools/gul/manager.py
```diff
@@ -93,32 +91,68 @@ def gul_get_items(input_path, ignore_file_type=set()):
...
+            (areaperil_id, vulnerability_id).
+        coverages (numpy.ndarray): coverage id to information on items.
 
     Returns:
-        item_map (Dict[ITEM_MAP_KEY_TYPE, ITEM_MAP_VALUE_TYPE]): dict storing
-          the mapping between areaperil_id, vulnerability_id to item.
+        item_map_hm (np.array[uint8]): packed hashmap table.
+        item_map_hm_keys (np.array[item_map_key_dtype]): key storage for the hashmap.
+        item_map_ja_offsets (np.array[oasis_int]): CSR offsets (N_pairs + 1).
+        item_map_ja_item_idxs (np.array[oasis_int]): flat item indices into items array.
```
SkylordA: stale comment, only 3 items returned, not item_map_ja_item_idxs

oasislmf/pytools/common/hashmap.py
```diff
@@ -0,0 +1,627 @@
...
+
+
+@overload(fnv1a)
+def fnv1a_overload_scalar(key, h=init_hash):
+    """Handles any numeric scalar: int8..int64, uint8..uint64, float32, float64, bool."""
+    if not isinstance(key, nbt.Number):
+        return None
+
+    def impl(key, h=init_hash):
+        return (h ^ np.uint64(key)) * np.uint64(1099511628211)
```
SkylordA: number here should reference FNV_PRIME

oasislmf/pytools/common/hashmap.py
```diff
@@ -0,0 +1,627 @@
...
+INFO_N = nb.int64(3)
+INFO_BYTES = INFO_N * INDEX_ITEMSIZE
+
+
+# ---------------------------------------------------------------------------
+# Bit helpers (unchanged from hashmap.py)
+# ---------------------------------------------------------------------------
+
+@nb.jit(nb.uint8(nb.uint8), cache=True)
+def extract_i_rh(lookup_val):
```
SkylordA: unused function/commented code from here to 226?

oasislmf/pytools/gul/structure.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Returns:
+        dict: mapping variable names to numpy arrays / scalars.
+    """
+    model_storage = get_storage_from_config_path(
+        os.path.join(run_dir, 'model_storage.json'),
+        os.path.join(run_dir, 'static'),
+    )
+    input_path = os.path.join(run_dir, 'input')
+    ignore_file_type = set(ignore_file_type)
```
SkylordA, line 73: this is not passed to downstream functions like read_coverages which do take ignore_file_type as an argument

tests/pytools/common/test_hashmap.py
SkylordA, line 1: I think more tests should be added, for example find_key is not explicitely tested at all, not sure about any other "modes" for functions as there's quite a few.

SkylordA (changes requested):

oasislmf/pytools/common/hashmap.py
```diff
@@ -0,0 +1,707 @@
...
+            # new: just inserted, info[HM_INFO_N_VALID] post-increment == this key's agg_id
+            res[i_item] = np.uint64(info[HM_INFO_N_VALID])
+        else:
+            # existing: copy agg_id from the first occurrence
+            res[i_item] = res[index_table[result]]
+
+    return res
+
+
+def factorize(df):
```
vinulw, line 828: not sure if this is a leftover interface but it is not directly tested anywhere, might be worth adding some tests particularly testing the different dtype cases.

tests/pytools/common/test_hashmap.py
```diff
@@ -0,0 +1,189 @@
...
+
+    The float64 case exercises bit-cast hashing AND bit-compare equality
+    for scalar floats: +0.0 and -0.0 are bit-distinct and must therefore
+    hash and compare as distinct keys."""
+    cases = [
+        (np.int32, np.array([5, 7, 100, 7, 5, 0], dtype=np.int32)),
+        (np.uint64, np.array([1, 2**40, 1, 0, 2**40], dtype=np.uint64)),
+        # +0.0 / -0.0 bit-distinct ⇒ distinct slots
+        (np.float64, np.array([0.0, -0.0, 1.5, 0.0, np.inf, -0.0], dtype=np.float64)),
+    ]
```
vinulw: minor: could use pytest fixtures to run each case independently?

oasislmf/pytools/common/id_index.py
```diff
@@ -0,0 +1,451 @@
...
+        arr[2] = min_id
+        arr[3] = max_id
+        data = arr[HEADER:]
+        for i in range(n):
+            data[np.int64(keys[i] - min_id)] = dtype.type(i)
+        return arr
+
+    # ── SORTED MODE (keys already sorted, store directly) ──
+    avg_error = _avg_err(keys)
+    if avg_error < 500.0:
```
vinulw: not sure I understand why this measure is independent of the length of `keys`. Generating keys with the non uniform generation method in `test_id_index.py` generated different modes depending on the number of elements in the key array: 

```python
In [40]: keys = np.unique(((rng.pareto(1.5, size=20000) + 1) * 100).astype(np.uint32))[:5000]

In [41]: _avg_err(keys)
Out[41]: np.float64(655.7834343144006)

In [42]: keys = np.unique(((rng.pareto(1.5, size=20000) + 1) * 100).astype(np.uint32))[:1000]

In [43]: _avg_err(keys)
Out[43]: np.float64(98.30368279970217)
```

oasislmf/pytools/common/id_index.py
```diff
@@ -0,0 +1,451 @@
...
+        return arr
+
+    min_id = keys[0]
+    max_id = keys[-1]
+
+    id_range = np.int64(max_id) - np.int64(min_id) + 1
+    density = n / id_range
+
+    # ── FLAT MODE ──
+    if density >= 0.25:
```
vinulw: minor: might be worth having the density cutoff (0.25) as a globabl constant

tests/pytools/common/test_hashmap.py
vinulw: from what I can tell the `float32` case is not directly tested in the `overload`ed functions

oasislmf/pytools/common/id_index.py
sstruzik: you are correct, I'm now dividing by the number of key. I've benchmarked it to find that 0.02 was a good value

tests/pytools/common/test_hashmap.py
sstruzik: done

oasislmf/pytools/common/hashmap.py
sstruzik: added some test for this too

tests/pytools/common/test_hashmap.py
sstruzik: I've added some test for more dtype, plus I made sure float and string are supported for both scalar and record version.

oasislmf/pytools/gul/structure.py
sstruzik: done

oasislmf/pytools/common/hashmap.py
sstruzik: done

oasislmf/pytools/common/hashmap.py
sstruzik: removed

oasislmf/pytools/gul/manager.py
sstruzik: fixed

vinulw (approved):

SkylordA (approved):

PR #1971, opened 2026-05-11
Fix IL merge failure when layers sharing a CondTag mix %-TIV and flat terms

sstruzik:
## Summary

Fixes a `pandas.errors.MergeError: Merge keys are not unique in right dataset; not a many-to-one merge` crash in IL input preparation when multiple policies/layers on the same account share a `CondTag` and at least one declares a `%-of-TIV` (or `BI`) financial term while another declares a flat or non-TIV-dependent term.

## Background

The cond FM levels (5/6/7) aggregate cross-layer using `agg_key = ['acc_id', 'CondTag']`. Per-layer cond terms are normally preserved by a *second* merge later in `get_il_input_items` whose join intersection includes `layer_id`. The first projection-merge at `il_inputs.py:923-924` deliberately drops `layer_id` so layers collapse — but the projection still carries `need_tiv`, which is computed per-row from each layer's own `lim_type` / `ded_type`. When layers disagree (e.g. one with `lim_type=2` → `need_tiv=True`, another flat → `need_tiv=False`) the right side of the merge has duplicate join keys with conflicting `need_tiv`, and `validate='many_to_one'` raises.

This pattern is the same one that the `validation/insurance_conditions/units/sc3` test exercises (4 policies sharing one CondTag with different `CondLimit6All` values), which works because all four happen to declare flat terms and therefore agree on `need_tiv=False`. The crash is specifically about disagreement, not about layering itself.

## Fix

`need_tiv` only drives a *per-agg_id* operation downstream (the `agg_tiv` aggregation for `%-TIV` term conversion). Layers in the same agg group share that `agg_tiv`; if any of them needs it, computing it is required, and layers that don't reference it (their `lim_type`/`ded_type` aren't 2 or 3) simply don't read it. So the correct coalesce across the join key is a boolean OR.

The fix groups `level_df` by the actual pandas merge join keys (the intersection of right-side columns and `gul_inputs_df.columns`) and OR-aggregates `need_tiv` via `groupby(...).transform('any')` before the projection-dedup. After coalescing, `drop_duplicates()` collapses to one row per join key and the merge passes.

This also fixes a related symptom where `get_cond_info` inserts zero-term pass-through `extra_accounts` rows for layers that don't reference a given CondTag — those have `need_tiv=False` and used to conflict with the real cond row's `need_tiv=True` whenever the cond carried a `%-TIV` term.

A new diagnostic helper, `_check_unique_merge_keys`, replaces pandas' opaque `MergeError` with an `OasisException` that names the FM level, lists the offending aggregation keys translated back to `PortNumber`/`AccNumber`, and shows the conflicting term values from `level_df`. The helper triggers only when a genuine OED inconsistency remains after the coalesce (same `(AccNumber, CondTag)` declaring genuinely different financial terms on multiple rows, which OED forbids).

## New test case

`validation/insurance_conditions/units/sc21` — minimal reproducer derived from a real-world failing portfolio: two policies on one account sharing `CondTag=1`, one with `CondLimit6All=100% TIV`, the other with `CondMinDed6All=200000`. Pre-fix this crashed; post-fix it produces deterministic per-layer cond profiles equivalent to the sc3 pattern.


oasislmf/preparation/il_inputs.py
```diff
@@ -463,6 +463,87 @@ def associate_items_peril_to_policy_peril(item_perils, policy_df, fm_peril_col,
...
+    Raises:
+        OasisException: When two or more rows share the same join keys but
+            disagree on a non-join right-side column. The message lists the
+            conflicting groups (translated to PortNumber/AccNumber when
+            possible) and the column values that disagree.
+    """
+    right_cols = list(agg_id_merge_col) + list(agg_id_merge_col_extra)
+    gul_cols = set(gul_inputs_columns)
+    # Replicate pandas' implicit `on=intersection(...)` behaviour.
+    join_keys = [c for c in right_cols if c in gul_cols]
```
Ha-Ree, line 541: Isn't this the same as on l939? Could just have it as an argument instead of computing twice

vinulw (approved):

Ha-Ree (approved):

PR #1992, opened 2026-05-22
Speed up summarypy read_buffer

sstruzik:
Two changes to `summarypy`'s `read_buffer` hot path:

1. **Parse pair payloads via a packed structured-dtype view instead of per-field `mv_read`.** After consuming the item header, the next bytes are a sequence of `(sidx, loss)` pairs ending in `(0, 0)`. Slice and view that whole payload once as a `loss_pair_dtype = [('sidx', oasis_int), ('loss', oasis_float)]` array, then iterate by field. Avoids two `byte_mv[c:c+n].view(dtype)[0]` calls per pair. The structured dtype carries the packed offsets, so this works for any `(oasis_int, oasis_float)` width combination — including `i4`/`f8`.

2. **Sort `present_summary_id` once per yielded event, not on every buffer exhaustion.** The previous code sorted at both the event-boundary return and at the function-exit return; the latter fired on every mid-event buffer-refill cycle and was wasted work because the next call could append more entries. Moved the sort into a new `SummaryReader.finalize_event()` method, called by `EventReader.read_streams` exactly once per yielded event via a default-noop `finalize_event()` hook on the base class.

3. add summarypy_low_memory option in the cli to control the creation of the summary idx file
4. Use binary format instead of csv when writing the summary index file (improve writing and reading speed).

Output is byte-identical to the previous implementation; `tests/pytools/summarypy/` and `tests/pytools/join-summary-info/` pass.

##performance
Change 1 give a nice improvement when tested on Piwind
| version | wall (warm cache) | CPU |
|---|---|---|
| mv_read (pre-change) | 10.4s | ~10s |
| slice+view (this PR) | 2.3s | ~2.3s |
| summarycalc (C++ reference) | 5.5s | 4.4s |

Change 2 become relevant when using multiple summary_set_id on big portfolio. Bespoke test set went from unable to be completed in a hour to completed in 8s


oasislmf/pytools/summary/manager.py
```diff
@@ -173,34 +176,36 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id,
...
-            sidx, cursor = mv_read(byte_mv, cursor, oasis_int, oasis_int_size)
-            if sidx:
-                loss, cursor = mv_read(byte_mv, cursor, oasis_float, oasis_float_size)
+            sidx_loss_view = byte_mv[cursor:cursor + n_sidx_loss * loss_pair_size].view(loss_pair_dtype)
+            terminated = False
+            for k in range(n_sidx_loss):
+                sidx = sidx_loss_view[k]['sidx']
+                if not sidx:
+                    ##### do item exit ####
+                    ##########
```
Ha-Ree, line 186: Line of hashes?? Was something initially inside or is this supposed to be at the end of the if?

oasislmf/pytools/summary/manager.py
```diff
@@ -230,10 +235,7 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id,
                 elif has_affected_risk is not None:
                     loss_summary[loss_index[summary_set_index], NUMBER_OF_AFFECTED_RISK_IDX] += new_risk
             ##########
-    for summary_set_index in range(summary_sets_id.shape[0]):  # reorder summary_id for each summary set
-        summary_set_start = summary_set_index_to_loss_ptr[summary_set_index]
-        summary_set_end = summary_set_index_to_present_loss_ptr_end[summary_set_index]
-        present_summary_id[summary_set_start: summary_set_end] = np.sort(present_summary_id[summary_set_start: summary_set_end])
+    # Buffer exhausted; caller may re-enter to accumulate more sidx_loss for this event,
```
SkylordA: just a comma at the end of the sentence here

SkylordA (approved):

Ha-Ree (approved):

oasislmf/pytools/summary/manager.py
sstruzik: It's from the template in oasislmf/pytools/common/event_stream.py,
All jitted read buffer function are designed the same way with specialization to do inside the ## part.
I prefer to keep it as a guide for future dev, so we can see directly where the custom code go. For me it is clearer

PR #1996, opened 2026-05-27
Relocate socket sevrer trap

### LINKED ISSUE #2004: Relocating socket server trap for consistency
From carl: socket server trap being outside of main trap can cause issues

Ha-Ree:
Move socket server trap into the same place as others to make ctrl c behaviour more consistent


oasislmf/execution/bash.py
```diff
@@ -213,6 +213,7 @@
    trap - QUIT HUP INT KILL TERM ERR EXIT
 
    kill -9 $pid0 2> /dev/null
+   [ -n "${spid:-}" ] && { kill -TERM "$spid" 2>/dev/null; kill -9 "$spid" 2>/dev/null; } || true
```
sambles: I'm not sure if both `kill -TERM "$spid" 2>/dev/null`   `SIGTERM`    and     `kill -9 "$spid" 2>/dev/null;` (`SIGKILL`)  need to be called one after the other. 

Its minor, but probably cleaner to have just one or the other.   If the socket-server background process has clean-up to do then `TERM`  otherwise a `kill -9`

sambles (changes requested):

oasislmf/execution/bash.py
SkylordA: From Claude:
Minor pre-existing issue worth fixing here while we're touching $spid cleanup:

`kill -0 "$spid" 2>/dev/null && kill -9 "$spid"`
If the socket-server exits naturally (e.g. all clients disconnect when the pipeline finishes) in the gap between `kill -0` succeeding and `kill -9` running, `kill -9` returns non-zero on the dead PID. Because this is the last command in a `&&` chain under `set -euET`, the ERR trap fires and the run is reported as `Kernel execution error` even though everything completed successfully.

Fix is just adding `2>/dev/null` to match the pattern already used in the exit_handler:

`kill -0 "$spid" 2>/dev/null && kill -9 "$spid" 2>/dev/null`

SkylordA (changes requested):

Ha-Ree:
Tests are failing on main unrelated to changes

SkylordA (approved):

sambles (approved):
looks good

PR #2005, opened 2026-06-03
gulmc/gulpy: Philox Latin Hypercube RNG (new default random_generator) + deterministic-hazard RNG skip

sstruzik:
## Summary

Improvements to gulmc/gulpy random-number generation, reviewable as separate commits.

> [!WARNING]
> **Default model output changes.** This PR makes the Philox Latin Hypercube generator the new default (`random_generator=2`). Default-configured runs now produce different (statistically equivalent, unbiased) loss samples than previous Mersenne-Twister-based versions. Select `--gul-random-generator 1` (LH on MT) or `0` (MT19937) to reproduce earlier results.

1. **Skip hazard RNG for deterministic hazards** (`a3c4688`) — when an areaperil's hazard intensity is a single bin (deterministic), no hazard random row is generated, and effective-damageability runs skip hazard RNG entirely. Value-preserving (each row is seeded independently, so dropping unused rows is bit-identical). All gulmc tests pass unchanged.

2. **`random_generator` option 2: Latin Hypercube on Philox4x32-7** (`978c04f`, `c564085`) — a counter-based, "reseed-free" generator. Philox is a keyed pseudo-random function, so a per-group stream is obtained by re-keying rather than re-initialising MT19937's 624-word state. That removes the per-row `np.random.seed()` digestion which dominates RNG cost at gulmc's sample sizes (~90% of per-row time at small sample counts), while preserving determinism per `(group_id, event_id)`.
   - Round function validated against the official Random123 known-answer vectors; LH structure (one sample per stratum) + determinism guarantees (repeatable, order-/batch-independent, `skip_seeds`) are unit-tested; `random_generator=2` is included in the gulmc and gulpy reference matrices.

3. **Make Philox the default** (`b5fb40b`) — default switched `1 → 2` across the gulmc/gulpy CLIs, the generate-losses computation params, the bash command builders and the JIT warmup. ids `0` (Mersenne-Twister) and `1` (Latin Hypercube on MT) remain selectable. Bash-script reference fixtures regenerated accordingly.

### Measured impact & validation
- PiWind `model run` (S=10): ~**20% lower gulmc CPU** vs LH(MT); the per-event RNG step is up to ~30× faster at small sample sizes (`generate_rndm` in isolation is ~6–20× faster for LH).
- PiWind rng1-vs-rng2 comparison: analytical losses **identical**; sampled losses and EP curves agree within Monte-Carlo noise and converge as samples grow (**no bias**) — see the validation comment below.
- Quality: Philox4x32-7 passes the discriminating statistical tests (incl. birthday-spacings) on par with MT19937.

### Usage
Philox is now used by default. To pin a generator: `oasislmf model run --gul-random-generator {0|1|2}` (or `gul_random_generator` in the config).

* **gulmc/gulpy: the default random generator is now Latin Hypercube on Philox4x32-7 (`random_generator=2`)**, a faster counter-based RNG. **Default model outputs change** versus previous versions (statistically equivalent and unbiased). Select `--gul-random-generator 1` (Latin Hypercube on Mersenne-Twister) or `0` (Mersenne-Twister) to reproduce earlier results.
* gulmc/gulpy: add `random_generator` option `2` (Latin Hypercube on Philox4x32-7).
* gulmc: skip hazard random-number generation for deterministic (single-intensity-bin) areaperils — faster, with unchanged results.


sstruzik:
### Validation: PiWind `model run`, rng=2 (Philox LH) vs rng=1 (MT LH)

Ran `oasislmf model run` on OasisPiWind with `--gul-random-generator 1` and `2`, comparing GUL/IL/RI ORD outputs at two sample sizes.

**Analytical losses (RNG-independent) — identical.** AAL and per-event mean loss match to float32 precision (reldiff `0.0000%`) across GUL/IL/RI at both sample sizes — confirming rng=2 feeds the same CDFs/deterministic pipeline and only perturbs the sampled component.

**Sampled losses — agree within Monte-Carlo noise and converge with sample count (no bias):**

| metric | S=10 | S=1000 |
|---|---|---|
| GUL AAL (sampled) reldiff | 0.13% | 0.0005% |
| IL/RI AAL (sampled) reldiff | 0.18% | 0.017% |
| GUL EP curve mean / max reldiff | 0.53% / 8.3% | 0.032% / 0.45% |

Largest S=10 deviation was the extreme tail (5000-yr AEP `8.3% → 0.29%` at S=1000). Differences flip sign across return periods (unbiased noise, not a fixed offset). MeanDR (analytical) EP curves are identical at every return period.

**Convergence efficiency:** on GUL (where the analytical mean is the true mean) both generators show the same per-event sampling-error magnitude and the same convergence rate — neither needs more samples than the other, as expected for two valid Latin-Hypercube generators sharing the same stratification.

**Conclusion:** rng=2 is statistically sound — identical on deterministic outputs, sampled outputs converge to the same result as rng=1 with no bias.

sstruzik:
need https://github.com/OasisLMF/OasisPiWind/pull/217 for change in CI output

oasislmf/pytools/gul/random.py
```diff
@@ -243,3 +262,126 @@ def random_LatinHypercube(seeds, n, skip_seeds=0):
...
+    See the module comment above `random_LatinHypercube_Philox7` for the algorithm.
+
+    Args:
+        seeds (array[int]): per-row seeds (a hash of group_id/event_id).
+        n (int): number of samples to generate for each seed.
+        skip_seeds (int): number of leading rows to skip (left as zeros); correlation
+          arrays pass 1.
+
+    Returns:
+        rndms (array[float64]): 2-d array of shape (len(seeds), n) of LH samples in [0, 1).
```
SkylordA: technically `(0, 1]` no?

SkylordA (approved):

oasislmf/pytools/gulmc/README.md
```diff
@@ -212,7 +212,7 @@ Positive indices `1..sample_size` are the Monte Carlo samples.
 ## Usage
 
 ```bash
-gulmc --random-generator=1 \
+gulmc --random-generator=2 \
```
vinulw, line 215: minor: should add info about `random_generator=2` to Key flags docs below as well

oasislmf/pytools/gul/random.py
```diff
@@ -243,3 +262,126 @@ def random_LatinHypercube(seeds, n, skip_seeds=0):
...
+                t = perms[idx]
+                perms[idx] = perms[jj]
+                perms[jj] = t
+                idx -= 1
+            if idx >= 1:
+                jj = int(np.float64(w3) * PHILOX_INV32 * np.float64(idx + 1))
+                t = perms[idx]
+                perms[idx] = perms[jj]
+                perms[jj] = t
+                idx -= 1
```
vinulw: minor: feel like this iteration over wi can be captured in a for loop, also might be worth sing similar head / tail logic as present in the jitter here to avoid some `idx >= 1` comparisons?

oasislmf/pytools/gul/random.py
```diff
@@ -243,3 +262,126 @@ def random_LatinHypercube(seeds, n, skip_seeds=0):
...
+    Nseeds = len(seeds)
+    rndms = np.zeros((Nseeds, n), dtype=np.float64)
+    perms = np.empty(n, dtype=np.float64)
+    inv_n = np.float64(1.0) / np.float64(n)
+    nfull = n - (n & 3)
+    zero = np.uint32(0)
+    for i in range(skip_seeds, Nseeds):
+        s = np.uint64(seeds[i])
+        k0 = np.uint32(s & PHILOX_U32_MASK)
+        k1 = np.uint32(s >> PHILOX_SHIFT32)
```
vinulw, line 331: from Claude: note `generate_hash` mod by 31 bits so `k1` is likely to always be 0. Worth checking just to make sure seed `k1` is not redundant.

vinulw (approved):
assuming `k0` provides enough randomness

oasislmf/pytools/gul/random.py
sstruzik: yes, done

oasislmf/pytools/gulmc/README.md
sstruzik: done

oasislmf/pytools/gul/random.py
sstruzik: yes done

oasislmf/pytools/gul/random.py
sstruzik: I considered widening to a 64-bit seed (using both key words), but it only benefits option 2 — options 0/1 are capped at 32 bits by np.random.seed — and it would force a second full regeneration of every rng2 
fixture + the PiWind expected set. So keeping k1 as-is for now.

PR #2006, opened 2026-06-03
fix(#1845): map OED InuringPriority to RI output levels

### LINKED ISSUE #1845: RI Inuring Priority doesn't map to RI outputs
**Issue**

- The RI insuring priority input in the analysis setting file is intended to allow outputs at any inuring priority
- As it stands, the outputs are actually generated at a combination of inuring priority and risk level
- So, if I have 2 RI contracts at different risk levels which don't overlap and both apply at inuring priority = 1, I actually end up with RI_1 and RI_2 levels being applied

**Example**

- RI Contract 1 applies to Account A1111 at Risk Level LOC with inuring priority 1
- RI Contract 2 applies to Account A1112 at Risk Level ACC with inuring priority 1 as well
- Analysis settings specifies IP 1 as an output, but the results only show the application of RI Contract 1

There is also no maintained mapping of the `RI_x` output levels to original Inuring Priority and Risk Level, so it's impossible for the user to know what the outputs mean

**Intended behaviour**
The outputs `RI_x` should map directly to the inuring priorities and no intermediate mapping should be required. All contracts at IP1, for example, should be applied together. I assume this design decision was taken to allow for overlapping contracts, but these are invalid and should be rejected rather than handled.

[analysis_settings.json](https://github.com/user-attachments/files/24879760/analysis_settings.json)

[SourceLocOEDPiWind10.csv](https://github.com/user-attachments/files/24879764/SourceLocOEDPiWind10.csv)
[SourceAccOEDPiWind.csv](https://github.com/user-attachments/files/24879765/SourceAccOEDPiWind.csv)
[SourceReinsInfoOEDPiWind.csv](https://github.com/user-attachments/files/24879762/SourceReinsInfoOEDPiWind.csv)
[SourceReinsScopeOEDPiWind.csv](https://github.com/user-attachments/files/24879763/SourceReinsScopeOEDPiWind.csv)

benhayes21:
## Summary

Fixes #1845.

Previously, `ri_inuring_priorities` in `analysis_settings` was expected to contain **RI layer indices** (1, 2, 3… counting every risk-level/priority combination), but users naturally supply **OED InuringPriority values** (1, 2, 3… from `ri_info.csv`). These two numbering schemes diverge whenever a single OED InuringPriority spans more than one risk level — e.g. InuringPriority 1 with both LOC and ACC contracts creates RI_1 and RI_2, so its "output level" is RI layer 2, not 1.

### Changes

- **`oasislmf/computation/generate/files.py`** — During RI input generation, after writing `ri_layers.json`, computes and writes `ri_inuring_priority_output_levels.json` to the run's `input/` directory. This file maps each OED InuringPriority (integer key) to the index of the last RI layer belonging to that priority (the output level).

- **`oasislmf/preparation/summaries.py`** — Adds `get_ri_inuring_priority_output_levels()` helper. Updates `generate_summaryxref_files()` to:
  - Validate `ri_inuring_priorities` from analysis settings as OED values (not layer indices).
  - Convert each requested OED priority to the correct RI output level before writing the `fmsummaryxref` files to the appropriate `RI_*` directories.
  - Copy `ri_inuring_priority_output_levels.json` into the run's `output/` directory so it is available alongside the results.

- **`oasislmf/execution/bash.py`** — Adds `import json`. In `bash_params()`, when `model_run_dir` is set and the mapping file exists, converts `analysis_settings['ri_inuring_priorities']` from OED values to RI layer indices in-place, so all downstream `get_ri_inuring_priorities()` call sites automatically produce the correct output levels in the generated bash script.

- **`oasislmf/computation/generate/losses.py`** — Both `bash_params()` call sites now pass `model_run_dir=model_run_fp` to enable the conversion above.

## Test plan

- [ ] Run existing bash script generation tests: `pytest tests/model_execution/test_bash.py` (207 pass)
- [ ] Run RI fmpy tests: `pytest tests/fm/test_fmpy.py -k reinsurance` (6 pass)
- [ ] End-to-end test with a multi-priority RI programme where one InuringPriority spans multiple risk levels — verify `ri_inuring_priority_output_levels.json` appears in both `input/` and `output/`, and that intermediate RI outputs are written to the correct RI layer directories

### DIFF
--- a/oasislmf/preparation/summaries.py
+++ b/oasislmf/preparation/summaries.py
@@ -14,6 +14,7 @@ __all__ = [
 
 import logging
 import pathlib
+import shutil
 
 import io
 import json
@@ -788,10 +789,10 @@ def generate_summaryxref_files(
         # Copy the inuring-priority-to-output-level mapping into the output directory so it is
         # available alongside the results for downstream consumers.
         pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
-        with io.open(os.path.join(model_run_fp, 'input', 'ri_inuring_priority_output_levels.json'), encoding='utf-8') as src_f:
-            mapping_data = src_f.read()
-        with io.open(os.path.join(output_dir, 'ri_inuring_priority_output_levels.json'), 'w', encoding='utf-8') as dst_f:
-            dst_f.write(mapping_data)
+        shutil.copyfile(
+            os.path.join(model_run_fp, 'input', 'ri_inuring_priority_output_levels.json'),
+            os.path.join(output_dir, 'ri_inuring_priority_output_levels.json'),
+        )
 
 
 def get_exposure_summary_field(df, exposure_summary, field_name, field_value, status):
--- a/oasislmf/preparation/summaries.py
+++ b/oasislmf/preparation/summaries.py
@@ -1,6 +1,7 @@
 __all__ = [
     'get_summary_mapping',
     'generate_summaryxref_files',
+    'get_ri_inuring_priority_output_levels',
     'merge_oed_to_mapping',
     'write_exposure_summary',
     'get_exposure_summary',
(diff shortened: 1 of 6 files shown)


oasislmf/preparation/summaries.py
```diff
@@ -718,36 +738,60 @@ def generate_summaryxref_files(
...
-                write_df_to_parquet_file(ri_summary_desc[desc_key], os.path.join(model_run_fp, 'output'), desc_key)
+                write_df_to_parquet_file(ri_summary_desc[desc_key], output_dir, desc_key)
             else:
-                write_df_to_csv_file(ri_summary_desc[desc_key], os.path.join(model_run_fp, 'output'), desc_key)
+                write_df_to_csv_file(ri_summary_desc[desc_key], output_dir, desc_key)
+
+        # Copy the inuring-priority-to-output-level mapping into the output directory so it is
+        # available alongside the results for downstream consumers.
+        pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
+        with io.open(os.path.join(model_run_fp, 'input', 'ri_inuring_priority_output_levels.json'), encoding='utf-8') as src_f:
```
vinulw: minor: do we need to read the file again? we can just output `inuring_priority_to_output_level` or better use `shutil.copyfile`.

tests/preparation/test_ri_inuring_priorities.py
```diff
@@ -0,0 +1,410 @@
...
+        with tempfile.TemporaryDirectory() as d:
+            with self.assertRaises(Exception):
+                get_ri_inuring_priority_output_levels(d)
+
+
+# ---------------------------------------------------------------------------
+# Tests: mapping computation (logic mirrored from files.py)
+# ---------------------------------------------------------------------------
+
+class TestInuringPriorityMappingComputation(TestCase):
```
vinulw, line 78: if we want to test this logic I feel like we should call the implemented logic from `files.py`, perhaps we can abstract it in files and call this method directly. Otherwise this test is just testing its own logic.

vinulw (approved):

sstruzik (approved):

PR #2007, opened 2026-06-03
Clear cache

### LINKED ISSUE #1990: Numba JIT stale cache causes incorrect results when `OASIS_FLOAT=f8` or `OASIS_AREAPERIL_INT=u8` with `cache=True`
**Component:** `oasislmf/pytools/` — any JIT-compiled function using `cache=True` with variable oasis type signatures  
**Severity:** Medium — silent correctness issue; no exception raised, results are silently wrong

### Description

Numba JIT caches compiled functions keyed on input types at first invocation. When oasis type environment variables (`OASIS_FLOAT`, `OASIS_AREAPERIL_INT`, etc.) are changed between runs without clearing the Numba cache, the previously compiled version is loaded from disk and executed with the wrong type specialisation.

This was observed with `mv_write_delimiter` and likely affects any JIT function that:
1. Has a signature that depends on variable-length oasis types resolved at import time from env vars
2. Has `cache=True` set

The function executes without raising an error but produces incorrect numerical output.

This issue has been ***hard to explicitly reproduce***, but running a script that forcefully removes the cache under the site-package and we have in our deployment removed all instances of cache=True and while this has marginally affected runtimes for small portfolios for larger runs it is insubstantial.

 
### Suggested fix

For any JIT function whose signature depends on oasis environment variables, **remove `cache=True`**. If caching is required for performance, use eager compilation with an explicit signature to pin the type

Ultimately, it would be preferred if the types are set as part of the model config jsons at runtime rather than being cooked into the environment. This would hopefully make the run more aware of type changes for cache clearing and for debugging, as well as making it easier for client running models locally on CLI with multiple types to avoid cross contamintation, where type errors are hard to diagnose and lead to fatal and obscure errors.

Ha-Ree:
Functionality to remove cache to fix issue about cache holding on to env vars without removing the cache from those functions

### DIFF
--- a/oasislmf/cli/clearcache.py
+++ b/oasislmf/cli/clearcache.py
@@ -1,0 +1,83 @@
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+from .command import OasisBaseCommand
+
+# Deletion runs in a subprocess with NUMBA_DISABLE_JIT=1 because Numba
+# re-creates cache files instantly if JIT is active in the same process
+
+_DELETE_SCRIPT = """\
+import sys
+from pathlib import Path
+
+roots = sys.argv[1:]
+files = [f for r in roots for ext in ('*.nbi', '*.nbc') for f in Path(r).rglob(ext)]
+removed = failed = 0
+for f in files:
+    try:
+        f.unlink()
+        removed += 1
+    except OSError as e:
+        print(f'WARNING: {e}', flush=True)
+        failed += 1
+print(removed, failed, flush=True)
+"""
+
+_CACHE_GLOBS = ('*.nbi', '*.nbc')
+
+
+def _find_cache_files(roots):
+    return [f for r in roots for ext in _CACHE_GLOBS for f in r.rglob(ext)]
+
+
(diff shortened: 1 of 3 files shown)


oasislmf/cli/clearcache.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Deletion runs in a subprocess with NUMBA_DISABLE_JIT=1 because Numba
+# re-creates cache files instantly if JIT is active in the same process
+
+_DELETE_SCRIPT = """\
+import sys
+from pathlib import Path
+
+roots = sys.argv[1:]
+files = [f for r in roots for ext in ('*.nbi', '*.nbc') for f in Path(r).rglob(ext)]
```
SkylordA: suggestion: should just take the `_CACHE_GLOBS` as a sys.argv here rather than hardcode again

SkylordA (approved):

Ha-Ree:
Tests are failing on main unrelated to changes

sstruzik (approved):

PR #2012, opened 2026-06-09
fix(hashmap): hash all record fields, not just the last

### LINKED ISSUE #2013: Model run fails with "rehashed too many times" (GULMC aggregate vulnerability weights / large gulpy item maps)
## Symptom

A model run aborts with:

```
Exception: rehashed too many times
```

This happens while building internal GUL/GULMC structures, before any losses are produced. You will most likely see it with **GULMC models that use aggregate vulnerabilities + vulnerability weights**, and it can also affect **gulpy** on large analyses (see "Who is affected" below).

## Error / traceback

GULMC (aggregate vulnerability weights):

```
- oasislmf.computation.base - INFO - Creating GULMC shared structures: .../runs
  ...
  File ".../oasislmf/pytools/gulmc/structure.py", line 219, in build_structures
    process_vulnerability_weights(...)
  File ".../oasislmf/pytools/common/hashmap.py", line 776, in rehash
    raise Exception("rehashed too many times")
Exception: rehashed too many times
```

The same `rehashed too many times` from `hashmap.py` can also surface via the gulpy item map (`oasislmf/pytools/gul/manager.py`).

## Who is affected

The failure happens when many keys differ only by `areaperil_id` for a shared `vulnerability_id`:

- **GULMC** with aggregate vulnerability definitions **and** a weights file (`weights.csv` / `weights.bin`) — e.g. postcode-resolution models. **Confirmed reproducible.**
- **gulpy** on large analyses where more than ~16 distinct `areaperil_id`s share the same `vulnerability_id` in the item map.

Models without these characteristics are not affected.

## How to reproduce (confirmed)

Using the `PiWindPostcode` model from [OasisModels](https://github.com/OasisLMF/OasisModels), which ships aggregate vulnerability weights:

```bash
cd OasisModels/PiWindPostcode/tests/test_2
oasislmf model run -C oasislmf.json -r ./runs
```

`test_2` runs with `gulmc: true` at postcode resolution (1248 areaperils), which exercises the aggregate-vulnerability-weights path. The run fails during `Creating GULMC shared structures`.

## Status

Tracked and fixed in #2012. With that fix applied, `PiWindPostcode/test_2` runs to completion (exit 0).

sstruzik:
## Summary

Fixes a bug in the structured-record path of the Robin Hood hashmap (`oasislmf/pytools/common/hashmap.py`) where the per-field FNV-1a accumulation **dropped all record fields except the last**, turning distinct keys into true 64-bit hash collisions.

### Root cause

`fnv1a_overload_record` builds the hash impl by string codegen and interpolated the FNV prime into the source as a **bare decimal literal**:

```python
src.append(f"    h = (h ^ np.uint64(record['{fname}'])) * {FNV_PRIME}")
# generates:  h = (h ^ np.uint64(record['areaperil_id'])) * 1099511628211
```

numba types `uint64 * <int literal>` as **int64** (not uint64). `init_hash` (~1.47e19) does not fit in signed int64, so the FNV multiply overflows and collapses to a constant (INT64_MIN), erasing the contribution of every field except the last one. Records differing only in an earlier field therefore hashed identically.

The scalar overload and the unicode branch were already correct because they multiply by the `FNV_PRIME` **name** (dtype uint64), keeping the arithmetic in unsigned 64-bit.

### Impact

`process_vulnerability_weights` (GULMC aggregate vulnerability weights) keys a hashmap on `(areaperil_id, vuln_idx)` records. The bug collapsed ~1248 distinct `areaperil_id`s into a single bucket as genuine 64-bit collisions, far exceeding the 16-slot Robin Hood probe limit. `rehash` cannot separate true collisions (a larger table doesn't change identical hashes), so the build aborted with `Exception: rehashed too many times`.

### Fix

Multiply by the uint64 `FNV_PRIME` name in the float and int branches of the record codegen (3 lines), matching the already-correct scalar/unicode paths.

### Verification

- Record hashes are now distinct per leading field and match an independent pure-Python FNV-1a + fmix64 reference implementation.
- `tests/pytools/common/test_hashmap.py`: **10/10 pass**.
- The **OasisModels `PiWindPostcode/test_2` model run now passes end-to-end** (`oasislmf model run`, exit 0). It previously failed with `rehashed too many times` while building the GULMC shared structures.

Fixed structured-record hashing in the GULMC hashmap so all record fields are mixed into the hash (previously only the last field contributed), which caused `rehashed too many times` failures for aggregate-vulnerability models such as PiWindPostcode.

### DIFF
--- a/oasislmf/pytools/common/hashmap.py
+++ b/oasislmf/pytools/common/hashmap.py
@@ -335,11 +335,11 @@ def fnv1a_overload_record(record, h=init_hash):
             if ftype.bitwidth == 64:
                 src.append("    _buf = np.empty(1, dtype=np.float64)")
                 src.append(f"    _buf[0] = record['{fname}']")
-                src.append(f"    h = (h ^ _buf.view(np.uint64)[0]) * {FNV_PRIME}")
+                src.append("    h = (h ^ _buf.view(np.uint64)[0]) * FNV_PRIME")
             else:  # float32 → zero-extend to 64 bits after bit-cast to uint32
                 src.append("    _buf = np.empty(1, dtype=np.float32)")
                 src.append(f"    _buf[0] = record['{fname}']")
-                src.append(f"    h = (h ^ np.uint64(_buf.view(np.uint32)[0])) * {FNV_PRIME}")
+                src.append("    h = (h ^ np.uint64(_buf.view(np.uint32)[0])) * FNV_PRIME")
         elif isinstance(ftype, nbt.UnicodeCharSeq):
             # Fixed-width unicode (e.g. 'U3') field. str(field) trims trailing
             # NUL padding via numba's UnicodeCharSeq.__len__ (numpy's fixed-
@@ -351,7 +351,7 @@ def fnv1a_overload_record(record, h=init_hash):
         else:
             # int/uint/bool: np.uint64() is already a zero-extension, equivalent to bitcast
             src.append(
-                f"    h = (h ^ np.uint64(record['{fname}'])) * {FNV_PRIME}"
+                f"    h = (h ^ np.uint64(record['{fname}'])) * FNV_PRIME"
             )
     # Final Murmur3 fmix64 — three rounds of shift / mult / shift. See _fmix64().
     src.append("    h ^= h >> np.uint64(33)")


SkylordA (approved):

sambles (approved):
Thanks for the fix

PR #2014, opened 2026-06-11
feature/aal_lec_idx_files

### LINKED ISSUE #1995: lecpy/aalpy: implement per-summary-id processing using binary .idx files to reduce disk usage
## Issue Description

For large per-location runs, lecpy and aalpy can produce enormous intermediate `.bdat` files (e.g. 767G for `lec_outloss_sample.bdat` with ~10k locations, 1k samples, 10k periods). This is because lecpy pre-allocates a dense array covering all possible `(period, sidx, summary_id)` combinations upfront regardless of how sparse the actual loss data is.

Two improvements are proposed:

**1. Per-summary-id processing using binary `.idx` files**
As of #1992, summarypy can generate packed binary `.idx` side-files alongside the `.bin` summary files when run with `-m`/`--low-memory`. If these are present, lecpy/aalpy should process one `summary_id` at a time using the `.idx` offsets, bounding peak disk usage to `O(periods × sidxs)` regardless of location count and eliminating the large `.bdat` files entirely.
Behaviour when `.idx` files are absent should fall back to the current dense memmap path unchanged.

**2. Disk space check before pre-allocating `.bdat` files**
When falling back to the dense memmap path (no `.idx` files present), lecpy/aalpy should estimate the required disk space upfront and raise a clear error if there is insufficient space, rather than failing mid-run or silently filling the disk.

SkylordA:
### feature/aal_lec_idx_files

Implements per-summary-id processing in `lecpy` and `aalpy` using binary `.idx` side-files, eliminating the large intermediate `.bdat` files that caused excessive disk usage on large per-location runs.

#### Changes
- `lec/manager.py`, `aal/manager.py`: when `.idx` files are present (generated by `summarypy -m`), process one `summary_id` at a time using `.idx` offsets, bounding peak disk to `O(periods × sidxs)` regardless of location count
- Falls back to the existing dense memmap path when `.idx` files are absent
- `lec/manager.py`: adds a best-effort disk space pre-check on the fallback path — raises a clear error before allocation rather than failing mid-run or silently filling the disk. When multiple lecpy instances run concurrently (e.g. gul/il/ri in parallel), the check can pass for all instances simultaneously even if their combined allocation exceeds available space; the idx path avoids this entirely
- `lec/aggreports/aggreports.py`: refactor to support the new per-summary processing path
- Tests added for both `aalpy` and `lecpy` covering the idx and fallback path

#### Benchmarks

**LEC results:**
image

**AAL results:**
image

closes https://github.com/OasisLMF/OasisLMF/issues/1995

### DIFF
--- a/oasislmf/pytools/lec/manager.py
+++ b/oasislmf/pytools/lec/manager.py
@@ -560,7 +560,7 @@ def run(
             return  # skip sequential path below
 
         # ── Sequential path (no .idx files) ──────────────────────────────────────
-        
+
         # Check required disk space for work bdat files
         num_sidxs = int(sample_size) + 2
         no_of_periods = int(file_data["no_of_periods"])
--- a/oasislmf/pytools/lec/manager.py
+++ b/oasislmf/pytools/lec/manager.py
@@ -16,7 +16,7 @@ from oasislmf.pytools.common.input_files import PERIODS_FILE, read_occurrence, r
 from oasislmf.pytools.lec.data import (AEP, AEPTVAR, AGG_FULL_UNCERTAINTY, AGG_SAMPLE_MEAN, AGG_WHEATSHEAF, AGG_WHEATSHEAF_MEAN,
                                        OCC_FULL_UNCERTAINTY, OCC_SAMPLE_MEAN, OCC_WHEATSHEAF, OCC_WHEATSHEAF_MEAN, OEP, OEPTVAR,
                                        OUTLOSS_DTYPE, EPT_dtype, EPT_fmt, EPT_headers, PSEPT_dtype, PSEPT_fmt, PSEPT_headers)
-from oasislmf.pytools.lec.aggreports import AggReports
+from oasislmf.pytools.lec.aggreports import AggReports, LecConfig, make_output_fn, output_for_summary_idx
 from oasislmf.pytools.lec.utils import get_outloss_mean_idx, get_outloss_sample_idx
 from oasislmf.pytools.utils import redirect_logging
 
@@ -222,7 +222,6 @@ def process_input_file(
     row_used_mean,
     outloss_sample,
     row_used_sample,
-    summary_ids,
     occ_map,
     use_return_period,
     num_sidxs,
@@ -235,7 +234,6 @@ def process_input_file(
         row_used_mean (ndarray[bool]): bool mask for outloss_mean
         outloss_sample (ndarray[OUTLOSS_DTYPE]): ndarray indexed by summary_id, sidx, period_no containing aggregate and max losses
         row_used_sample (ndarray[bool]): bool mask for outloss_sample
-        summary_ids (ndarray[bool]): bool array marking which summary_ids are used
         occ_map (nb.typed.Dict): numpy map of event_id, period_no, occ_date_id from the occurrence file
         use_return_period (bool): Use Return Period file.
         num_sidxs (int): Number of sidxs to consider for outloss_sample
@@ -244,7 +242,6 @@ def process_input_file(
     # Set cursor to end of stream header (stream_type, sample_size, summary_set_id)
     cursor = oasis_int_size * 3
 
-    # Read all samples
     valid_buff = len(fin)
     while cursor < valid_buff:
         event_id, cursor = mv_read(fin, cursor, oasis_int, oasis_int_size)
@@ -261,7 +258,6 @@ def process_input_file(
             continue
 
         filtered_occ_map = occ_map[event_id]
-        summary_ids[summary_id - 1] = True
 
         while cursor < valid_buff:
             sidx, cursor = mv_read(fin, cursor, oasis_int, oasis_int_size)
@@ -292,7 +288,6 @@ def run_lec(
     row_used_mean,
     outloss_sample,
     row_used_sample,
-    summary_ids,
     occ_map,
     use_return_period,
     num_sidxs,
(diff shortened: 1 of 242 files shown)


tests/pytools/aalpy/test_aalpy.py
```diff
@@ -160,3 +185,127 @@ def test_alct_output_parquet():
...
+        # Add an empty partition: header-only .bin + 0-byte .idx
+        stream_hdr = np.frombuffer(stream_info_to_bytes(SUMMARY_STREAM_ID, 1), dtype=np.int32)[0]
+        np.array([stream_hdr, 100, 1], dtype=np.int32).tofile(work_dir / "summarypy9.bin")
+        (work_dir / "summarypy9.idx").touch()
+
+        out_dir = tmp_dir / "out"
+        out_dir.mkdir()
+
+        main(run_dir=tmp_dir, subfolder="gul", ext="csv", aal=out_dir / "py_aalgul.csv")
+        assert (out_dir / "py_aalgul.csv").exists(), "AAL output should be created when populated bins are present"
```
vinulw: minor: might be worth still checking the output is correct here to make sure empty partition does not change result

tests/pytools/lecpy/test_lecpy.py
```diff
@@ -1,15 +1,42 @@
...
+import pytest
+from oasislmf.pytools.common.data import summary_stream_index_dtype
+from oasislmf.pytools.common.event_stream import SUMMARY_STREAM_ID, stream_info_to_bytes
 from oasislmf.pytools.lec.data import EPT_dtype, PSEPT_dtype
 from oasislmf.pytools.lec.manager import main
 
 TESTS_ASSETS_DIR = Path(__file__).parent.parent.parent.joinpath("assets").joinpath("test_lecpy")
 
 
+def make_idx_from_bin(bin_path: Path, idx_path: Path) -> None:
```
vinulw: minor: might be worth abstracting this as its used in both `test_aalpy` and `test_lecpy`.

tests/pytools/lecpy/test_lecpy.py
```diff
@@ -212,3 +237,126 @@ def test_lec_output_parquet():
...
+        with patch("oasislmf.pytools.lec.manager.shutil.disk_usage", return_value=fake_usage):
+            with pytest.raises(RuntimeError, match="Insufficient disk space"):
+                main(
+                    run_dir=tmp_dir, subfolder="gul", use_return_period=False,
+                    ept=out_dir / "py_ept.csv", psept=out_dir / "py_psept.csv", ext="csv",
+                    **_ALL_OUTPUT_FLAGS,
+                )
+
+
+def test_lec_idx_partial_coverage_falls_back_to_sequential():
```
vinulw, line 314: minor: would be good to have this kind of test for aal as well

oasislmf/pytools/lec/manager.py
```diff
@@ -145,14 +147,82 @@ def do_lec_output_agg_summary(
...
+    for file_index, idx in enumerate(idx_handles):
+        if idx is None or len(idx) == 0:
+            continue
+        chunk = np.empty(len(idx), dtype=_MERGED_IDX_DTYPE)
+        chunk["summary_id"] = idx["summary_id"]
+        chunk["file_index"] = file_index
+        chunk["offset"] = idx["offset"]
+        parts.append(chunk)
+    if not parts:
+        return np.empty(0, dtype=_MERGED_IDX_DTYPE)
```
vinulw, line 213: dimension inconsistent between the two returns. looks to me like this branch is 1 dim smaller than than the other one due to `np.concatenate`. Doesn't cause error right now but think its worth fixing just in case for future.

oasislmf/pytools/lec/manager.py
```diff
@@ -248,14 +313,37 @@ def run_lec(
...
             row_used_sample,
-            summary_ids,
             occ_map,
             use_return_period,
             num_sidxs,
             max_summary_id,
         )
 
 
+def _open_output_files(outmap, stack, output_binary, output_parquet, noheader):
```
vinulw, line 323: minor: for loops could be outer scope to avoid redudant for loop and `if not outmap` logic in each branch

oasislmf/pytools/lec/manager.py
SkylordA: I'm pretty sure the dimension is the same
```python3
import numpy as np

dtype = np.dtype([("summary_id", np.int32), ("file_index", np.int32), ("offset", np.int64)])

empty_path  = np.empty(0, dtype=dtype)
concat_path = np.concatenate([np.empty(3, dtype=dtype), np.empty(2, dtype=dtype)])

print(empty_path.ndim,  empty_path.shape)   # 1 (0,)
print(concat_path.ndim, concat_path.shape)  # 1 (5,)
print(empty_path.ndim == concat_path.ndim)  # True

```

oasislmf/pytools/lec/manager.py
SkylordA: True, but I prefer the readability of this way around, plus avoids having to check the output type on each loop.

oasislmf/pytools/lec/manager.py
vinulw: my bad, I didn't realise what `np.empty` was doing.

vinulw (approved):
👍🏾

sstruzik (approved):

PR #2023, opened 2026-06-16
fix(lookup): assume EPSG:4326 when geometry file has no CRS metadata

### LINKED ISSUE #2022: fix(lookup): CRS missing on geometry file causes sjoin failure
## Problem

When using the `rtree` lookup with a geometry file (e.g. a parquet generated with an older version of geopandas), the file may not have CRS metadata embedded. Newer versions of geopandas enforce CRS matching on `gpd.sjoin()`, so joining a location GeoDataFrame with `EPSG:4326` against a geometry GeoDataFrame with `CRS: None` raises:

```
Use `to_crs()` to reproject one of the input geometries to match the CRS of the other.

Left CRS: EPSG:4326
Right CRS: None

  gdf_loc = gpd.sjoin(gdf_loc, gdf_geometry, 'left')
```

## Cause

`gdf_geometry` is read from file in `builtin.py` without any fallback CRS assignment. Older parquet files written by earlier geopandas versions often lack embedded CRS metadata.

## Fix

After reading `gdf_geometry`, check if `crs is None` and call `set_crs("EPSG:4326")` to assume the geometry file shares the same coordinate reference system as the location points.

benhayes21:
## Summary

- Geometry files (e.g. parquet) generated with older geopandas versions may lack embedded CRS metadata, resulting in `CRS: None`
- Newer geopandas enforces CRS matching on `gpd.sjoin()`, crashing with a `Use to_crs() to reproject` error
- After reading the geometry file, fall back to `set_crs("EPSG:4326")` if no CRS is present

Fixes #2022

## Test plan

- [ ] Run rtree lookup with a parquet geometry file that has no embedded CRS — confirm no crash
- [ ] Run rtree lookup with a parquet geometry file that has EPSG:4326 embedded — confirm no change in behaviour
- [ ] Run rtree lookup with a shapefile (has CRS) — confirm no change in behaviour

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 OasisLMF Changelog
 ==================
 
+`2.5.5`_
+ ---------
+* [#2023](https://github.com/OasisLMF/OasisLMF/pull/2023) - fix(lookup): assume EPSG:4326 and warn when geometry file has no CRS metadata
+
 `2.5.4`_
  ---------
 * [#1940](https://github.com/OasisLMF/OasisLMF/pull/1940) - enhancement/profile check script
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
vinulw, line 1: don't think we need to edit the changelog manually, not sure if this will conflict with the changelog update script

vinulw (approved):
assuming changelog is fine

CHANGELOG.rst
benhayes21: reverted, thanks

sstruzik (approved):

PR #2028, opened 2026-06-18
Added max sample size flag to gulmc testing

sambles:
### Added max sample size flag to gulmc testing
With JIT disabled unit tests can take 40-50mins, a lot of time can be saved by limiting the sample sizes used in gulmc. 
Added `--gulmc-max-sample-size=<int>`  flag to set this option


SkylordA (approved):

sstruzik (approved):

sambles:
already done in https://github.com/OasisLMF/OasisLMF/pull/2005/  (without flag, but same result)

PR #2029, opened 2026-06-22
fix(il): NaN acc_idx crash when an FM level has terms but none apply to keys perils (#1989)

### LINKED ISSUE #1989: `ValueError: cannot convert NA to integer` for `acc_idx` in `get_summary_mapping` when running FM

sstruzik:
## Description

Fixes #1989, where `oasislmf model generate-oasis-files` crashes with:

```
ValueError: cannot convert NA to integer: Error while type casting for column 'acc_idx'
```

### Root cause

During IL input construction (`get_il_input_items`), every item starts at `layer_id = 1`. The final step joins items back to their account on `['PolNumber', 'acc_id', 'layer_id']` to recover `acc_idx` (used later by `get_summary_mapping`).

The per-level loop has a "reset non layered agg_id" step that sets unlayered items to `layer_id = 0` *in anticipation* of the level merge restoring their real `layer_id`. The problem: this reset ran **before** the `if level_df.empty: continue` guard.

When a level has term *definitions* (so it passes the earlier `if not terms_maps` skip) but, after filtering, **no term actually applies** — e.g. the only at-risk coverage's keys peril does not match the policy's `PolPerilsCovered` — `level_df` comes out empty and the loop `continue`s. The restoring merge is skipped, so the item is **stranded at `layer_id = 0`**.

The account row has `layer_id = 1`, so the final `acc_idx` left-merge on `layer_id` misses → `acc_idx` is `NaN` → `get_summary_mapping` fails casting `NaN` to `uint32`.

This is distinct from the `PolNumber` backfill issue fixed in #1942 (same symptom, different column), which is why it still reproduced on 2.5.4.

### Fix

Move the `layer_id` reset to **after** the empty-`level_df` check, so a level that is skipped never disturbs `layer_id`. Between the old and new positions `layer_id` is only written, never read, so behaviour for non-empty levels is unchanged.

### Testing

- New regression test `test_files__level_with_no_matching_terms__acc_idx_populated__issue_1989` (with the minimal reproducer exposure + keys as fixtures). Fails on the previous code, passes with the fix.
- `tests/fm/test_fmpy.py`: 21/21 pass (unchanged from baseline — no regression).
- `tests/computation/test_generate_files.py`: all pass.

Fix `ValueError: cannot convert NA to integer` for `acc_idx` when an FM level has policy terms defined but none apply to the perils returned by the keys lookup.


benhayes21:
this should close issue #1989 too

Ha-Ree (approved):

SkylordA (approved):

PR #2044, opened 2026-06-26
fix: IL outputs are non-deterministic when accounts CSV rows are reordered

### LINKED ISSUE #2040: Ordering changes values in oasislmf exposure run
[ils_different_when_reordered.zip](https://github.com/user-attachments/files/29374426/ils_different_when_reordered.zip)

2.4.11- large difference 
```
$ cat */output2.csv
PortNumber,loss_gul,loss_il
1,143641680400.0,59063.68
PortNumber,loss_gul,loss_il
1,143641680400.0,11065974.66
PortNumber,loss_gul,loss_il
1,143641680400.0,23143651.7
```
Large difference also observed in personal runs of 2.5.4 but not by client report?

Needs to be checked and fixed

Ha-Ree:
**Fix: IL outputs are non-deterministic / incorrect when account CSV rows are reordered (issue #2040)**

Reordering the rows of account.csv (same data, different order) produced different insured losses; GUL losses were unaffected. The bug exists in release/2.4.13 and also persists in 2.5.4 — so 2.5.4 is not a correct reference.

**Root cause**
The "drop premature layering" dedup in get_il_input_items collapsed policy layers that carried genuinely different financial terms (e.g. layer-specific CondLimits). The dedup keyed on [gul_input_id, agg_id, profile_id, layered_id], where layered_id is only set to the real layer_id once an agg is already "layered". At the condition level no agg is yet layered, so layered_id became 0 for every row and distinct layers were merged into one — dropping layers and applying the wrong LayerParticipation. Because the surviving row depended on accounts.csv order, reordering changed the result, and once otherwise stabilised the collapsed output was simply wrong (some locations dropped to zero IL).

This collapse is only valid when all of a (gul_input, agg)'s layers share the same condition profile (the case for ordinary accounts and the existing FM acceptance fixtures).

**Fix (confined to oasislmf/preparation/il_inputs.py)**
Premature-layering dedup — keep a row per layer when a (gul_input_id, agg_id) has more than one distinct profile_id across its layers (i.e. layer-specific terms), and collapse only when every layer shares the same profile:

```
layer_specific = non_layered_inputs_df.groupby(['gul_input_id', 'agg_id'])['profile_id'].transform('nunique') > 1
layered_id     = layer_id.where(layer_specific | agg_id.isin(layered_inputs_df['agg_id']), 0)
# dedup on [gul_input_id, agg_id, profile_id, layered_id]
```
This preserves every distinct policy layer (and its participation) for layer-specific accounts, giving correct, order-independent losses, while remaining a byte-for-byte no-op for ordinary accounts so the FM structure stays canonical.

PolNumber backfill — use reset_index()/set_index('index') so the assignment aligns even when gul_inputs_df has a non-contiguous index from CondTag row-expansion.

acc_idx — add a layer-aware fallback: a preserved policy layer can carry a non-canonical PolNumber, leaving acc_idx unmatched by the term-based merge; fill any remaining gaps from the (acc_id, layer_id[, CondTag]) relationship that determines the source account row (acc_idx drives summary grouping only, not loss values).

Note: earlier iterations also added determinism sorts (sort before cumcount in prepare_account_df; sort level_df by layer_id before get_profile_ids; re-assign the post-concat sort_values). These were dropped — they are unnecessary (the dedup fix alone makes the output order-independent) and they changed the canonical agg_id/layer_id/row ordering of the FM files, breaking the byte-comparison FM acceptance tests (test_insurance_account, test_insurance_conditions).

**Verification**
Synthetic 4-layer scenario: hand-calculated correct IL = 737,501; fix produces 737,500.66 across original/reversed/shuffled orderings (the old collapse gave 166,942, with locations zeroed out).
validation/insurance validated expected total IL (19,729,855,529) reproduced; FM structure stays canonical.
Real issue dataset: order-independent at loss_il = 61,266,006 for all orderings.
tests/fm/test_fm.py 9/9 pass; tests/fm/test_fmpy.py insurance_account/insurance_conditions comparisons pass (the fix is a no-op for those fixtures).
Adds tests/preparation/test_il_ordering.py with regression tests covering FM-file identity across account row orderings, that every genuine policy layer is preserved, and an end-to-end IL loss check with an absolute-magnitude guard (so a future regression that is order-independent but numerically wrong is still caught)


oasislmf/preparation/il_inputs.py
sambles: Claude seems to think that this line needed to be changed, it was in fix attempt 1, did something break? 
I think it tracks, unless `sort_values` works in place? 


`gul_inputs_df = gul_inputs_df.sort_values(by=['gul_input_id', 'layer_id'])`

sambles (commented):
it was also pretty keen on added a sort to  `prepare_account_df` after the line


https://github.com/OasisLMF/OasisLMF/blob/8e0c6dbf2dbae8a4a6f4e7313698e300b206ba13/oasislmf/utils/data.py#L829

`id_df = id_df.sort_values(layers_cols + ['PolNumber', 'LayerNumber'])`

but not sure if thats needed or not

oasislmf/preparation/il_inputs.py
Ha-Ree: I'll put both in and run the tests here but I seem to remember some of the initial sorts being unnecessary and breaking some things

benhayes21:
Hi @Ha-Ree — while digging into #2040 to understand exactly why the dedup was collapsing rows, I found a couple of things that seem worth flagging for review.

**1. `PolNumber` (or any per-policy identity) was never part of the dedup key, before or after this fix.**

The key is `['gul_input_id', 'agg_id', 'profile_id', 'layered_id']`. `agg_id` is shared across all policies on the account at these levels, so the *only* column that could ever preserve per-policy identity was `layered_id` (derived from `layer_id`, itself a stand-in for `(PolNumber, LayerNumber)`). Pre-fix, `layered_id` was unconditionally zeroed for every row not already flagged "layered" by an earlier level — so at any level below Policy Layer, the effective dedup key was just `(gul_input_id, agg_id, profile_id)`, with no way to tell two different policies apart if they happened to resolve to the same `profile_id`.

**2. This looks like a general defect across every level below Policy Layer, not just `cond all`.**

Per `oasislmf/utils/fm.py`, the level order is:
```
site item → site coverage → site pd → site all →
cond coverage → cond pd → cond all →
policy coverage → policy pd → policy all →
policy layer → account all
```
The affected branch runs for `not is_policy_layer_level and level_id not in cross_layer_level` — i.e. all 10 levels before Policy Layer / Account All. `cond all` is just the level with data in the reproduction; `policy coverage` / `policy pd` / `policy all` (ordinary `PolLimit*` / `PolDed6All` terms — no `CondNumber` involved at all) would hit the identical collapse given a similar "some policies share a resolved term, one differs" pattern.

The `layer_specific` fix is computed generically per-level, so it should already cover these too — that part looks right to me.

**3. Possible test coverage gap**

`tests/preparation/test_il_ordering.py`'s account fixture header is:
```
PortNumber,AccNumber,AccCurrency,PolNumber,PolPerilsCovered,
CondPeril,CondNumber,CondLimit6All,CondDed6All,...,LayerParticipation,...
```
There's no `PolLimit*` / `PolDed6All` field, so the suite can only exercise `cond all` — it doesn't prove the fix holds at `policy coverage` / `policy pd` / `policy all`, where the same class of bug is structurally possible even with zero Conditions involved.

Would it be worth adding a second fixture that varies `PolLimit1Building` / `PolDed6All` directly across policies (no `CondNumber` at all) to close that gap and confirm the fix's full claimed scope, rather than just the level it happened to be discovered on?

benhayes21:
Follow-up to my comment above on test coverage — one refinement worth separating out.

I'd rank the coverage gaps differently by priority:

**Higher priority: `policy coverage` / `policy pd` / `policy all`**

These are a completely different OED field family (`Pol*`, not `Cond*`) with **zero** existing test coverage at any of the three sub-levels — not even the `all` case. `test_il_ordering.py`'s fixture has no `PolLimit*` / `PolDed6All` variation at all, so nothing currently proves the fix holds there.

**Lower priority, but still worth adding: `cond coverage` / `cond pd`**

These share the `Cond*` field family with `cond all`, which *is* tested. They run through the identical shared dedup block (`layer_specific` → `layered_id` → `drop_duplicates`), which is blind to what `agg_id` semantically represents — it only needs `gul_input_id` + `agg_id` + `profile_id` to compare correctly, regardless of granularity. So `cond all` passing gives reasonable indirect confidence for the group as a whole.

The residual risk is narrower than "does the fix work here at all": `cond coverage` and `cond pd` build `agg_id` at a finer granularity than `cond all` (coverage level splits further by `coverage_type_id`, pd by peril), so the open question is really whether that finer `agg_id` construction ever interacts with the `layer_specific` groupby in a way the coarser `all` case doesn't exercise — a different, more specific failure mode than #2040 itself.

Given the existing fixture already parameterizes by `CondTag`, adding a `cond coverage` / `cond pd` variant (e.g. varying `CondLimit1Building` / `CondLimit2Other` per coverage type instead of the blanket `CondLimit6All`) should be a small, low-cost addition — I'd just prioritize the `policy` group fixture first, since that one has no coverage at all right now.

benhayes21 (approved):

PR #2045, opened 2026-06-26
2.4/fix failing  FM acceptance tests

sambles:
### fix failing  FM acceptance tests

### Tests fixed
- `FmAcceptanceTests::test_insurance_conditions` (`model_perils_covered=['WTC']`)
- `FmAcceptanceTests::test_insurance_conditions_2_subperils` (`model_perils_covered=['WW2']`)

### Root cause

Commit `b87bed9ce` ("fix: in special condition remove necessity for dummy layer with multiple cond peril") directly spliced two extra test scenarios — **sc21** and **sc22** — into the combined `validation/insurance_conditions/account.csv` and `location.csv` files, without creating corresponding unit directories under `units/` or entries in `units.txt`. This broke the test in two ways:

**1. sc21 — `MergeError` in the condition-level merge (`il_inputs.py`)**

sc21 has two policies (PolNumber 1010 and 1234) that share the same `CondTag=1`. One policy has a non-zero TIV-proportional limit (`CondLimitType6All=2`, value≠0 → `need_tiv=True`) and the other has a zeroed limit (value=0 → `lim_type` zeroed by `prepare_ded_and_limit` → `need_tiv=False`). After `drop_duplicates()` on the right-side dataframe for the `many_to_one` merge, both rows survive (they differ only in `need_tiv`), but share the same actual join key `(acc_id, CondTag, FMTermGroupID, peril_id)` — violating the `many_to_one` constraint.

**2. sc22 — `ValueError: cannot convert NA to integer` on `acc_idx`**

sc22 has policies that go through the condition level but have no matching row at the policy layer level, leaving `PolNumber=NaN` in `gul_inputs_df`. At the final `acc_idx` merge, neither the primary merge (keyed on `CondTag + PolNumber + acc_id + layer_id`) nor the fallback merge (keyed on `PolNumber + acc_id + layer_id`) can match a `NaN` PolNumber, so `acc_idx` stays `NaN`. Writing the summary map then fails when pandas tries to cast the nullable `Int64` column to `uint32`.

#### 1. `oasislmf/preparation/il_inputs.py` (committed as "Fix from Claude for FM test issue")

Added a guard before the condition-level `many_to_one` merge. When the right-side dataframe has duplicate join keys (i.e. the actual join key — columns present in both dataframes — is not unique), the non-join columns (e.g. `need_tiv`) are aggregated with `max` so that `need_tiv=True` wins:

```python
agg_id_merge_col = list(set(agg_id_merge_col).intersection(level_df.columns))
_right_df = level_df[agg_id_merge_col + agg_id_merge_col_extra].drop_duplicates()
# Policies can share a condition (same CondTag) but differ in term types,
# causing non-join columns (e.g. need_tiv) to conflict on the same join key.
# Resolve by taking max over the actual join key so need_tiv=True wins.
_join_key = [c for c in _right_df.columns if c in gul_inputs_df.columns]
_non_join = [c for c in _right_df.columns if c not in gul_inputs_df.columns]
if _non_join and _right_df[_join_key].duplicated().any():
    _right_df = (_right_df.groupby(_join_key, as_index=False, dropna=False)
                 .agg({c: 'max' for c in _non_join}))
gul_inputs_df = gul_inputs_df.merge(_right_df, how='left', validate='many_to_one')

This is a defensive fix for a valid real-world scenario (multiple policies sharing a condition with different term types). It is a no-op when no duplicate join keys exist.

2. validation/insurance_conditions/account.csv and location.csv

Regenerated both combined files from units.txt using combine.py:

python validation/combine.py -dir validation/insurance_conditions --suffix ""

This removes the sc21 and sc22 rows that were manually spliced in by b87bed9ce. The units.txt file lists fm12, fm41, sc1–sc14, sc16–sc20 (no sc21/sc22), which is the authoritative source of truth for what should be in the combined files.

Why the tests are now passing

With sc21/sc22 removed from the combined input files, neither of the two broken data scenarios is present. The il_inputs.py fix remains in place as a safeguard for the general case of policies sharing a condition with conflicting need_tiv values.
```


sambles:
should be fixed, merging into release branch for review and testing

PR #2051, opened 2026-07-02
enhance/supported_oed_versions

SkylordA:
### enhance/supported_oed_versions
Delegate `supported_oed_versions` conversion to `OedExposure`

Removes the `to_version` call from `GenerateKeys.run()` and forwards `supported_oed_versions` (from `model_settings.data_settings`) and `disable_oed_version_update` through `get_exposure_data_config()` on all computation step entry points. Conversion now happens inside `OedExposure.__init__()` (ODS_Tools #286), covering all construction paths uniformly.

Updated 3 tests to assert on config forwarding rather than `to_version` mock calls.

closes https://github.com/OasisLMF/ODS_Tools/issues/230


sstruzik (approved):

vinulw (approved):
assuming https://github.com/OasisLMF/ODS_Tools/pull/286 merged and is compatible

PR #2052, opened 2026-07-07
enhance/pytools converters dtypes

### LINKED ISSUE #2036: update pytools converters to use dtypes driven by environment variabls
## Background

PR #2027 fixed a specific instance of this problem: the carry-state scalars in
`footprint_tobin` were hardcoded as `np.uint32(0)`, causing type mismatches when
`OASIS_AREAPERIL_TYPE=u8`. The fix was to use `dtype['areaperil_id'].type(0)` so
the scalar always matches the configured type.

The same class of problem exists more broadly across the converter utilities: many
files use hardcoded NumPy dtypes (`np.int32`, `np.uint32`, `np.float32`, etc.)
instead of the environment-variable-driven types defined in
`oasislmf/pytools/common/data.py`:

```python
oasis_int      = np.dtype(os.environ.get('OASIS_INT', 'i4'))
oasis_float    = np.dtype(os.environ.get('OASIS_FLOAT', 'f4'))
areaperil_int  = np.dtype(os.environ.get('OASIS_AREAPERIL_TYPE', 'u4'))

### PR DESCRIPTION
**Pull request by vinulw**
### Summary
The pytools binary/CSV converters and event-stream writers previously hard-coded `int32`/`float32` for stream fields (`event_id`, `item_id`, `sidx`, `loss`, etc.). This branch propagates the configurable dtypes (`oasis_int`, `oasis_float`, `areaperil_int`, driven by the `OASIS_INT` / `OASIS_FLOAT` / `OASIS_AREAPERIL_TYPE` environment variables) through the entire conversion path, so round-trips stay correct when non-default types are configured.

### Main changes

**Core type handling** (`common/data.py`, `common/event_stream.py`)
- Add `def_to_type_and_size()` helper to resolve a field definition to its `(np.dtype, itemsize)`.
- Derive `loss_pair_dtype`, `item_header_dtype`, and `summary_stream_index_dtype` from the configurable field definitions instead of fixed types.
- Split the numba stream writers (`mv_write_item_header`, `mv_write_sidx_loss`) into public wrappers + `*_cached` implementations that take type/size info as arguments, so numba caching stays valid when dtypes change.
- Fix an incorrect `section_id` header name (was mislabelled `scale2`).

**Converters** (`csvtobin` & `bintocsv` )
- Replace hard-coded `int32`/`float32` reads/writes and chunk-size math with dtype-derived sizes and the shared write helpers.
- Fix GUL/FM cache-write and delimiter handling, `fm` `csvtobin`, the `bintocsv` `fm.py` read buffer, and abstract the vulnerability header size.

**getmodel & fm** (`getmodel/common.py`, `footprint.py`, `manager.py`, `vulnerability.py`, `fm/compare.py`, `footprintconv.py`)
- Use numpy dtypes for `FootprintHeader`, fix a circular import, correct footprint/vulnerability type usage, and fix the loss type in `fm/compare.py`.

**Tests**
- New `test_converters_env.py`: runs conversions in a subprocess under adjusted dtype environments and verifies both output equivalence and the serialized dtypes.
- Add `envdtype` test assets (items, vulnerability, weights, fm_profile, gul/ils, coverages, damagebin) and refactor the existing converter tests to run as class methods.

### CODE REVIEW

File: oasislmf/pytools/fm/compare.py [hunk 1]
```diff
@@ -34,12 +40,9 @@ def stream_to_dict_array(stream_obj):
...
         else:
+            if stream_obj.readinto(event_agg_mv) < event_agg_dtype.itemsize:
+                break
             event_id, agg_id = event_agg[0]
-            # if event_id_last != event_id:
-            #     if event_id_last:
-            #         break
-            #     else:
-            #         event_id_last = event_id
             cur_array = np.zeros(len_sample + EXTRA_SIDX_COUNT, dtype=oasis_float)
```
SkylordA: missing import, removed by changes

oasislmf/pytools/common/event_stream.py
SkylordA, line 152: why not add same caching pattern here?

tests/pytools/converters/test_converters.py
SkylordA, line 1: do a flake8 (and probably pep8) check of these test files, I don't think they are tested by CI by default, but there are some dead imports now

oasislmf/pytools/converters/bintocsv/utils/gul.py
```diff
@@ -77,8 +74,9 @@ def _update_idxs():
             for k in range(n_pairs):
                 sidx = sidx_loss_view[k]["sidx"]
                 if sidx == 0:  # sidx == 0, end of record (loss field is the trailing 0)
+                    #                    breakpoint()
```
SkylordA: left over breakpoint

oasislmf/pytools/converters/bintocsv/utils/gul.py
```diff
@@ -45,23 +48,17 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id, data, idxs, stat
     last_event_id = event_id
     idx = idxs[0]
 
-    def _reset_state():
```
SkylordA, line 48: technically this change is fine, however I liked the use of these functions to clearly show what was happening as it's the same pattern used in elt/plt too.

oasislmf/pytools/fm/compare.py
SkylordA: also i assume since this never produced an error, this path was probably untested.

SkylordA (changes requested):

oasislmf/pytools/converters/bintocsv/utils/gul.py
vinulw: sure, i've put back in both `_reset_state` and `_update_idxs` which are both present in `elt` and `plt`

oasislmf/pytools/converters/bintocsv/utils/fm.py
```diff
@@ -45,23 +48,19 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id, data, idxs, stat
...
-            if valid_buff - cursor >= 2 * oasis_int_size:
-                event_id_new, cursor = mv_read(byte_mv, cursor, oasis_int, oasis_int_size)
+            if valid_buff - cursor >= event_id_dtype_size + item_id_dtype_size:
+                event_id_new, cursor = mv_read(byte_mv, cursor, event_id_dtype,
+                                               event_id_dtype_size)
                 if last_event_id != 0 and event_id_new != last_event_id:
                     # New event, return to process the previous event
-                    _update_idxs()
-                    return cursor - oasis_int_size, last_event_id, item_id, 1
+                    idx = idxs[0]
```
Ha-Ree: Is this the intended direction? This won't change anything and it seems like you've done the opposite in all the other calls?

SkylordA (approved):

Ha-Ree (approved):
Assuming tests pass again

oasislmf/pytools/converters/bintocsv/utils/fm.py
vinulw: rolled back the change to the same pattern from `elt` and `plt`

PR #2057, opened 2026-07-10
Extent api client CLI options

sambles:
<!--start_release_notes-->
## New subcommands

### `oasislmf api post`
Upload files to the platform.
- `--portfolio-id` / `--portfolio-name` + `--oed-location-csv` / `--oed-accounts-csv` / `--oed-info-csv` / `--oed-scope-csv` / `--currency-conversion-json` / `--reporting-currency` — uploads to a portfolio; omitting `--portfolio-id` creates a new one.
- `--analyses-id` + `--analyses-settings-json` — uploads analysis settings (requires `--analyses-id`).
- `--model-id` + `--model-settings-json` — uploads model settings (requires `--model-id`).
- Errors if nothing is given to upload.

### `oasislmf api validate`
Validate a portfolio's OED exposure files.
- `--portfolio-id` (required).
- Default behaviour: triggers a new validation run and **polls** until it completes, then returns the detailed result.
- `--get-status` — skip triggering/polling, just return the current status snapshot.
- `--poll-interval` (default `5`s).

### `oasislmf api exposure-run`
Runs `oasislmf exposure run` server-side against a portfolio (mirrors `oasislmf exposure run -h` flags: `-a/--kernel-alloc-rule-il`, `-A/--kernel-alloc-rule-ri`, `--model-perils-covered`, `-l/--loss-factor`, `--supported-oed-coverage-types`, `--extra-summary-cols`, `--fmpy-low-memory`, `--fmpy-sort-output`, `--check-oed`, `--do-disaggregation`).
- `--portfolio-id` (required), `-o/--output-dir` for the downloaded result.
- Default: triggers a run, **polls** to completion, downloads the result file.
- `--fetch` — skip triggering, just download the last result.
- `--poll-interval` (default `5`s).

### `oasislmf api exposure-transform`
Converts a portfolio's exposure data between OED and AIR.
- `--portfolio-id`, `--file-type` (`location`/`accounts`/`ri_info`/`ri_scope`), `--mapping-file`, `--transform-file` (all required).
- Triggers the transform and **polls** until complete, then returns the updated portfolio object.
- `--poll-interval` (default `5`s).

### `oasislmf api combine`
Combines the ORD output of multiple `RUN_COMPLETED` analyses.
- `--analysis-ids` (required, list), `--combine-settings-json` (required), `--combine-name` (default `combine-analysis`).
- Fire-and-forget — returns the immediate API response, no polling.

### `oasislmf api cancel`
Cancels an in-flight analysis (input generation or run, whichever is active).
- `--analysis-id` (required).

### `oasislmf api sub-tasks`
Lists the sub-tasks of an analysis run.
- `--analysis-id` (required).


### `oasislmf api reconnect`
Reconnect to an in-progress  analysis and resume polling for status
- `--analysis-id` (required).

### `oasislmf api plot` 
Plot a Gantt chart of an analysis's sub-tasks, with a status summary 

<img width="1950" height="1906" alt="analysis-4-gantt" src="https://github.com/user-attachments/assets/1859af29-98e1-48e6-91ed-1e45af8080b6" />

## Changes to existing subcommands

### `oasislmf api list`
- Added `-t/--subtask <id...>` — print task-status details fing `-m/--models`, `-p/--portfolios`, `-a/--analyses`).
                                                                                                                                                                                                                                                                                                      
### `oasislmf api get`
- Added `--subtask-output-log`, `--subtask-error-log`, `--subtask-retry-log <id...>` — download task-status logs, following the same `<id...>` list pattern as the existing model/portfolio/analyses download fl


<!--end_release_notes-->


Ha-Ree:
Tested all the new commands except for `combine` (@vinulw should know more about that) and overall looks good but maybe would be worth changing something to ensure v2 only endpoints (transform, exposure run, combine) either only run on v2 or throw an early error

oasislmf/computation/run/platform.py
```diff
@@ -505,3 +591,485 @@ def run(self):
...
+    Portfolio files are uploaded to a single portfolio per invocation - give
+    `--portfolio-id` to update an existing portfolio, or omit it (optionally
+    with `--portfolio-name`) to create a new one.
+    """
+    step_params = PlatformBase.step_params + [
+        # Portfolio
+        {'name': 'portfolio_id', 'type': int, 'help': 'API `id` of an existing portfolio to update. Omit to create a new portfolio'},
+        {'name': 'portfolio_name', 'help': 'Name for a newly created portfolio (ignored if --portfolio-id is given)'},
+        {'name': 'oed_location_csv', 'flag': '-x', 'is_path': True, 'pre_exist': True, 'help': 'Source location CSV file path'},
+        {'name': 'oed_accounts_csv', 'flag': '-y', 'is_path': True, 'pre_exist': True, 'help': 'Source accounts CSV file path'},
```
vinulw, line 642: minor: could use clearer flags for `location` and `accounts` files e.g. `l` and `a`

oasislmf/computation/run/platform.py
```diff
@@ -505,3 +591,485 @@ def run(self):
...
+            self.logger.info('Exposure run: Starting (id={})'.format(self.portfolio_id))
+            self.poll_portfolio_field(self.portfolio_id, 'exposure_status', self.poll_interval, 'Exposure run')
+            rsp = self.server.portfolios.exposure_run.get(self.portfolio_id)
+
+        content_type = rsp.headers.get('content-type', '').partition(';')[0].strip()
+        ext = guess_extension(content_type) or '.bin'
+        filename = os.path.join(self.output_dir, 'portfolio_{}_exposure_run{}'.format(self.portfolio_id, ext))
+        with io.open(filename, 'wb') as f:
+            f.write(rsp.content)
+        self.logger.info('Exposure run result saved to: {}'.format(filename))
```
vinulw, line 782: not sure if we want the exposure run to always download the results after, we don't do this for other run commands by default. Might be better to have this as an option

oasislmf/computation/run/platform.py
```diff
@@ -505,3 +591,485 @@ def run(self):
...
+            summaries = settings.get(f'{perspective}_summaries') or []
+            if settings.get(f'{perspective}_output') and summaries:
+                rows.append((f'{label} summaries', self._describe_summaries(summaries)))
+            else:
+                rows.append((f'{label} summaries', 'Disabled'))
+
+        return rows
+
+    def _plot(self, sub_tasks, analysis, settings, output_file, plt, mdates, Patch):
+        sub_tasks = sorted(sub_tasks, key=lambda t: t["id"])
```
vinulw, line 988: minor: might want to handle edge case where no subtasks e.g. an analysis in the `NEW` state which generates and empty figure

vinulw:
> Tested all the new commands except for `combine` (@vinulw should know more about that) and overall looks good but maybe would be worth changing something to ensure v2 only endpoints (transform, exposure run, combine) either only run on v2 or throw an early error

verified that the `oasislmf api combine` command works

vinulw (commented):
Main thing that might need changing is allowing the option to download the `exposure-run` command instead of always downloading.

oasislmf/computation/run/platform.py
Ha-Ree: Best to leave as x and y imo as this is also standard for oasislmf model run?

sambles: yeah this, its following the existing flags from model run. I think its ok for this PR, and changing it (everywhere) could be pushed into another clean up issue

Ha-Ree (changes requested):
see above comments

Ha-Ree (approved):

vinulw (approved):
🚀

PR #2060, opened 2026-07-10
fix(pre-analysis): preserve original exposure file format for snapshot saves

### LINKED ISSUE #2059: ExposurePreAnalysis silently downgrades parquet exposure sources to slow csv snapshots
## Summary

`oasislmf/computation/hooks/pre_analysis.py`'s `ExposurePreAnalysis.run()` calls `exposure_data.save(...)` twice (the "raw" original snapshot and the post-adjustment snapshot) without an explicit `compression=` argument. For a freshly loaded exposure source, this silently defaults to `csv` regardless of the original input file's format - even when the source was parquet.

This makes the pre-analysis snapshot writes drastically slower than necessary for large portfolios. On a ~23.4M-row location parquet file (744MB), the raw snapshot write took well over an hour as csv (~350-450KB/s) versus **under two minutes** as parquet (~27MB/s) once fixed.

## Root cause

- `ods_tools.oed.source.OedSource.from_filepath()` builds the initial `sources['orig']` dict as `{'source_type': 'filepath', 'filepath': filepath, 'read_param': read_param}` - it never records an `'extension'` key.
- `ods_tools.oed.exposure.Exposure.save()` determines the output compression with:
  ```python
  if compression is None:
      if oed_source.sources[oed_source.cur_version_name]['source_type'] == 'filepath':
          compression = oed_source.sources[oed_source.cur_version_name].get('extension')
      if compression is None:
          compression = 'csv'
  ```
  Since the freshly loaded `'orig'` version has no `'extension'` key, this always falls through to `'csv'` for the very first save of any newly loaded source - independent of the real file's format, which is only ever consulted at *read* time via `Path(filepath).suffix` (in `OedSource.load_dataframe`), never at *save* time.
- `pre_analysis.py`'s two `exposure_data.save(...)` calls (raw snapshot and adjusted snapshot) don't pass `compression=`, so both hit this default.

## Reproduction

1. Provide a `.parquet` location file as `oed_location_csv` with a reasonably large row count (a few hundred thousand rows is enough to see the difference clearly).
2. Run `oasislmf model generate-oasis-files` (or any command that triggers the `ExposurePreAnalysis` hook).
3. Observe `raw_location.csv` (not `.parquet`) being written in the run's input directory, at csv write speed, regardless of the original parquet input.

## Suggested fix

See PR (linked below) - derive the compression to use from the original location source's file extension instead of hardcoding a new default, so existing csv-based workflows are unaffected and parquet-based workflows keep their original format through the pre-analysis snapshots.

benhayes21:
## Summary
- Fixes #2059 - `ExposurePreAnalysis.run()`'s raw/adjusted exposure snapshot saves silently defaulted to csv for any freshly loaded source, even when the original input was parquet, because `Exposure.save()`'s compression fallback reads a `'extension'` key that's never set on a source's initial `'orig'` version (see issue for the full root-cause trace through `ods_tools`).
- Added `get_source_compression()`, which derives the save format from the original location source's file extension, and passes it explicitly to both `exposure_data.save()` calls in `ExposurePreAnalysis.run()`.
- Falls back to `None` (i.e. the previous csv default) when the format can't be determined, so existing csv-based workflows are unaffected.

## Impact
On a ~23.4M-row location parquet source (744MB), the raw pre-analysis snapshot write dropped from an estimated 1+ hour (csv, ~350-450KB/s) to under 2 minutes (parquet, ~27MB/s) with this change - end to end pre-analysis + keys lookup verified on that portfolio after the fix.

## Test plan
- [x] Added `test_exposure_pre_analysis_preserves_parquet_source_format` (parquet source in, asserts `raw_location.parquet`/`location.parquet` are written and no `.csv` snapshot files exist)
- [x] Existing `tests/model_preparation/test_exposure_pre_analysis.py` suite still passes unchanged (csv-based tests are unaffected since their source extension is correctly detected as csv)
- [x] `flake8` clean on the modified file
- [x] Verified against a real ~23.4M-row parquet portfolio on a test model - pre-analysis and keys lookup now complete promptly instead of stalling on the csv write

### DIFF
--- a/oasislmf/computation/hooks/pre_analysis.py
+++ b/oasislmf/computation/hooks/pre_analysis.py
@@ -4,13 +4,48 @@ __all__ = [
 
 import json
 import pathlib
-from ods_tools.oed import UnknownColumnSaveOption
+from pathlib import Path
+from ods_tools.oed import UnknownColumnSaveOption, PANDAS_COMPRESSION_MAP
 
 from ..base import ComputationStep
 from ...utils.data import get_exposure_data, prepare_oed_exposure, analysis_settings_loader, model_settings_loader
 from ...utils.inputs import str2bool
 from ...utils.path import get_custom_module
 from ...utils.exceptions import OasisException
+
+
+def get_source_compression(exposure_data):
+    """
+    Derive the compression/format to use when persisting the pre-analysis
+    exposure snapshots, based on the original location source file's extension.
+
+    Exposure.save() defaults to csv whenever no explicit compression is given
+    and the current source version has no recorded 'extension' (which is
+    always true for a freshly loaded source - see ods_tools
+    OedSource.from_filepath). Without this, the raw/adjusted exposure
+    snapshots below are silently written as csv even when the original input
+    was e.g. parquet, which can be drastically slower for large portfolios.
+
+    Args:
+        exposure_data (OedExposure): the loaded exposure data
+
+    Returns:
+        str or None: a key of ods_tools.oed.common.PANDAS_COMPRESSION_MAP
+                      matching the original location file's extension, or
+                      None if it can't be determined (Exposure.save() then
+                      falls back to its default of csv, unchanged from
+                      current behaviour).
+    """
+    if not exposure_data.location:
+        return None
+    source = exposure_data.location.current_source
+    if source.get('source_type') != 'filepath':
+        return None
+    suffix = Path(source['filepath']).suffix.lstrip('.')
+    for compression, mapped_suffix in PANDAS_COMPRESSION_MAP.items():
+        if mapped_suffix.lstrip('.') == suffix:
+            return compression
+    return None
 
 
 class ExposurePreAnalysis(ComputationStep):
@@ -98,7 +133,8 @@ class ExposurePreAnalysis(ComputationStep):
 
         ids_option = {'loc_id': UnknownColumnSaveOption.DELETE,
                       'loc_idx': UnknownColumnSaveOption.DELETE}
-        exposure_data.save(path=input_dir, version_name='raw', save_config=True, unknown_columns=ids_option)
+        source_compression = get_source_compression(exposure_data)
+        exposure_data.save(path=input_dir, version_name='raw', compression=source_compression, save_config=True, unknown_columns=ids_option)
         kwargs['exposure_data'] = exposure_data
         kwargs['input_dir'] = input_dir
         kwargs['model_data_dir'] = self.model_data_dir
@@ -126,7 +162,7 @@ class ExposurePreAnalysis(ComputationStep):
         print(_class(**kwargs))
         _class_return = _class(**kwargs).run()
 
-        exposure_data.save(path=input_dir, version_name='', save_config=True, unknown_columns=ids_option)
+        exposure_data.save(path=input_dir, version_name='', compression=source_compression, save_config=True, unknown_columns=ids_option)
         # regenerate ids
         exposure_data.location.dataframe = exposure_data.location.dataframe.drop(columns=['loc_id', 'loc_idx'])
         prepare_oed_exposure(exposure_data)
--- a/tests/model_preparation/test_exposure_pre_analysis.py
+++ b/tests/model_preparation/test_exposure_pre_analysis.py
@@ -1,6 +1,7 @@
 import os
 from tempfile import TemporaryDirectory
 
+import pandas as pd
 import pytest
 
 from oasislmf.manager import OasisManager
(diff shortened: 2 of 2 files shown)


Ha-Ree:
Could this use --intermediate-csv option Stephane added a few months ago?

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -13,6 +14,40 @@
 from ...utils.exceptions import OasisException
 
 
+def get_source_compression(exposure_data):
```
SkylordA, line 17: `get_source_compression()` derives compression from the location file only, but this value gets passed into `exposure_data.save()` and applied to every OED source (`location`, `account`, `ri_info`, `ri_scope`). A submission with `location.parquet` + `account.csv` will have its account file silently force-converted to `parquet`. Compression needs to be derived per-source, not once from `location`.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -13,6 +14,40 @@
...
+                      current behaviour).
+    """
+    if not exposure_data.location:
+        return None
+    source = exposure_data.location.current_source
+    if source.get('source_type') != 'filepath':
+        return None
+    suffix = Path(source['filepath']).suffix.lstrip('.')
+    for compression, mapped_suffix in PANDAS_COMPRESSION_MAP.items():
+        if mapped_suffix.lstrip('.') == suffix:
```
SkylordA, line 46: case sensitivity of suffix, should lowercase it first

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -4,7 +4,8 @@
 
 import json
 import pathlib
-from ods_tools.oed import UnknownColumnSaveOption
+from pathlib import Path
```
SkylordA, line 7: `pathlib.Path` is already used elsewhere in code, so kind of a double import

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -13,6 +14,40 @@
...
+                      None if it can't be determined (Exposure.save() then
+                      falls back to its default of csv, unchanged from
+                      current behaviour).
+    """
+    if not exposure_data.location:
+        return None
+    source = exposure_data.location.current_source
+    if source.get('source_type') != 'filepath':
+        return None
+    suffix = Path(source['filepath']).suffix.lstrip('.')
```
SkylordA, line 44: This only preserves format for the pre-analysis snapshot. The actual gap is in `ods_tools`: `OedSource.from_filepath` never records the source's original extension, so `OedExposure.save(compression=None)` always defaults to csv for any freshly-loaded source.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -98,7 +133,8 @@ def run(self):
 
         ids_option = {'loc_id': UnknownColumnSaveOption.DELETE,
                       'loc_idx': UnknownColumnSaveOption.DELETE}
-        exposure_data.save(path=input_dir, version_name='raw', save_config=True, unknown_columns=ids_option)
+        source_compression = get_source_compression(exposure_data)
```
SkylordA, line 136: `source_compression` is derived before `exposure_pre_analysis_module.run()` executes and reused unchanged for the post-hook save. If a hook ever replaces `exposure_data.location`'s underlying source rather than just editing the dataframe, the stale value would apply. Don't think it's likely but still worth mentioning.

benhayes21:
@Ha-Ree Good question, but I think you're referring to a different mechanism. `intermediary_csv` (on `GenerateFiles`/`RunExposure`/`GenerateLosses`) controls whether the **GUL/IL intermediate input files** produced during the genbash/file-generation stage are csv vs parquet - it's a performance/size tradeoff for that stage, defaulting to parquet since Stephane's commit.

This PR fixes a different set of files: the **pre-analysis exposure snapshots** (`raw_location`/`location`, `raw_account`/`account`, etc.) saved by `ExposurePreAnalysis.run()` before/after the pre-analysis hook runs. There's no existing flag for that - the bug was that these snapshots always silently defaulted to csv regardless of the original input format. So `intermediary_csv` doesn't cover this case; it's a sibling feature for a different pipeline stage.

benhayes21:
@sambles heads up for the release: this PR and #2157 both touch `ExposurePreAnalysis.run()` in `oasislmf/computation/hooks/pre_analysis.py` (this one adds `get_source_compression()`/`save_exposure_data()` around the two `exposure_data.save()` calls; #2157 restructures the code between them to add multiprocessing). #2157 was branched before this PR existed, so it doesn't yet account for these changes.

Whichever of the two you merge first, the other will need a rebase to reconcile the overlapping section of `run()` - they shouldn't be merged in parallel without one rebasing onto the other first. Flagging so you can sequence them for the release rather than merging both independently.

benhayes21:
@SkylordA all 5 review comments should now be addressed - flagging so you can take a look and resolve the threads if you agree:

1. **Per-source compression leaking across sources** (the main correctness bug) - fixed via `save_exposure_data()`, which groups sources by their own derived compression and saves each group in a separate `exposure_data.save()` call (temporarily hiding the other sources), so one source's format is never force-applied to another.
2. **Case sensitivity of the suffix match** - fixed, `.lower()` is now applied before matching against `PANDAS_COMPRESSION_MAP`.
3. **Duplicate `pathlib`/`Path` import** - fixed, only `import pathlib` remains and `pathlib.Path` is used consistently.
4. **Root cause is in `ods_tools`** - agreed, and filed separately: [ODS_Tools#308](https://github.com/OasisLMF/ODS_Tools/issues/308) / [ODS_Tools#309](https://github.com/OasisLMF/ODS_Tools/pull/309) (you're tagged as reviewer there). This PR's fix remains necessary independently of that one, since it also encodes the csv→parquet upgrade decision below.
5. **Stale `source_compression` reused across the hook run** - resolved as a side effect of the `save_exposure_data()` refactor: compression is now recomputed fresh from each source's current state at every save call, rather than derived once upfront and reused.

Separately (not from your review, but worth flagging): a csv source is now upgraded to parquet by default rather than just preserved as csv, since parquet is much faster for large portfolios - any other recognized format is still preserved as-is per source. See the updated PR description for the full rundown and test plan.

SkylordA (approved):

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -13,6 +14,84 @@
...
+    for oed_name, oed_source in original_sources.items():
+        if oed_source:
+            sources_by_compression[get_source_compression(oed_source)].append(oed_name)
+
+    try:
+        for compression, oed_names in sources_by_compression.items():
+            for oed_name, oed_source in original_sources.items():
+                setattr(exposure_data, oed_name, oed_source if oed_name in oed_names else None)
+            exposure_data.save(path=path, version_name=version_name, compression=compression,
+                               save_config=False, unknown_columns=unknown_columns)
```
Ha-Ree: save_config=False makes filepath in ods_tools absolute instead of relative: can just set to true because it will be overwritten at the end anyway?

Ha-Ree (changes requested):

PR #2063, opened 2026-07-14
feat(gulmc): coverage dependency (source-driven conditional vulnerability)

### LINKED ISSUE #1248: Conditional Damage Functions
Proposal to include conditional damage functions in oasislmf

Example spreadsheets attached

Conditional Damage Functions (conditional_demo.xlsx)

- Want to tie contents/BI damage to building damage
- Proposed solution is outlined in dummy script (conditionalDamage.py)
- Workaround exists that fits within the oasis framework can tie the damage ratios back to the original hazard intensity. However, the probability distributions are not properly preserved.
- Correlation – full vulnerability correlation can miss “valid combinations”, while uncorrelated vulnerability can draw “invalid combinations” partial correlation may be a good compromise
- Portfolio level – effect is likely negligible. However, may be significant on location-level (particularly for underwriting). This is also an important feature by users, particularly for US models

Multi-peril sampling (sampling_example.xlsx)

- On admin unit resolution, the joint probabilities of combinations of perils
- For certain perils (particularly flood), this is very important to be preserved. In testing, losses were affected on the portfolio level when the correlation between perils was lost. 
- Preferred approach is to create new bins representing combinations of intensity bins (for different perils)
- Issue: a very large number of new bins will be created, and all of these intensity bins need to be defined in the vulnerability table – in a test case, vulnerability size increased significantly (60 MB -> 210 GB). Having runtime issues
- Ideal solution: add a step between the hazard and damage sampling, to map the new bin that was drawn (representing combinations of bins for different perils) back to the peril-specific intensity bin; this would allow the original smaller vulnerability to be used for damage sampling
- Other proposed approach is described in the spreadsheet as “Proposed 1”
- This seems to work (in earlier testing), but would require that the footprint never be resorted. This seems like it could be broken very easily.
[conditionalDamage.py.txt](https://github.com/OasisLMF/OasisLMF/files/11102568/conditionalDamage.py.txt)
[sampling_example.xlsx](https://github.com/OasisLMF/OasisLMF/files/11102569/sampling_example.xlsx)
[conditional_demo.xlsx](https://github.com/OasisLMF/OasisLMF/files/11102570/conditional_demo.xlsx)

sstruzik:
Addresses #1248 (Conditional Damage Functions — the coverage-dependency part; multi-peril sampling in that issue is out of scope here).

## Coverage dependency in gulmc

Opt-in, data-driven **coverage dependency**: a dependent coverage (e.g. Contents) is driven by a
**source** coverage (e.g. Building) at the same location and peril. The source's per-sample sampled
damage bin indexes the dependent's **conditional (damage-transition) vulnerability** — a
`P(dependent damage bin | source damage bin)` matrix — instead of an independent draw, so the
dependent's severity is tied to how badly the source was damaged in that same sample.

### How it works
- **Config** lives in `model_settings.json` as `coverage_dependency_settings`: a list of
  `{source_coverage_type, dependent_coverage_type}` pairs.
- The link is resolved **per item** during file generation into a `source_item_id` column on the
  **correlations** file: each dependent item names the source item at its own location, building
  and peril. gulmc derives the coverage-level dependency forest from those item links.
- `gulmc` computes each dependent after its source via a DFS over that forest, driving it from the
  source item's stored damage bin (full Monte Carlo) or the source's effective-damage marginal
  (effective damageability — marginal-only by design).
- Conditional vulnerabilities are a separate model input (`conditional_vulnerability.{csv,bin}`),
  sized `num_damage_bins x num_damage_bins`, independent of the footprint's intensity resolution,
  and registered with the standard converters as the `conditionalvulnerability` file type.
- An uninsured (zero-TIV) source is retained purely to drive its dependent, and reports **zero
  loss** whatever the damage type. Retention is tested on (location, peril, areaperil) so nothing
  is retained that cannot actually drive a dependent, and it follows a chain — in a configured
  `building -> contents -> BI` with only the BI insured, both the contents and the building are kept.

### Why the link is per item, not per coverage
A coverage can hold several items at one areaperil (two perils geocoded to the same cell), and the
source's and dependent's vulnerability ids come from different id spaces — conditional ids are not
in `vulnerability.bin` at all — so their sort orders need not agree. Naming the source item removes
any need for the engine to infer the pairing from item ordering, which could otherwise cross the
perils silently. `source_item_id` **replaces** the `source_coverage_id` an earlier revision of this
PR added, so the correlations record is 24 bytes either way — but note it is 24 vs 20 bytes against
`main`, since the column itself is new in this PR.

### Mixed conditional and hazard-indexed vulnerability on one coverage type
Supported, and verified end to end. An item that finds no source item — the location has no source
coverage, or the key server placed the configured pair at different areaperils (logged at INFO) —
is computed independently. File generation cannot decide whether that is acceptable, because only
the model's static data says which vulnerability ids are conditional, so gulmc resolves it per item:
an unpaired item with a hazard-indexed vulnerability runs independently; an unpaired item with a
conditional vulnerability is refused. A model may therefore supply a conditional vulnerability where
the cells align and a hazard-indexed one where they do not, and both locations run.

### Guarantees
- **No change when unconfigured**: the dependency machinery is short-circuited by a
  `do_coverage_dependency` flag and an all-zero forest. The byte-for-byte golden files for
  `test_model_1`..`7` are unaffected.
- Fail-loud validation: a dependent must use a conditional vulnerability, an unpaired item must not,
  dependents can't be aggregate, source links can't be non-existent/self-referential/cyclic, and the
  vulnerability data may not declare more damage bins than the `damage_bin_dict`.
- `conditional_vulnerability` conversion validates ids ascending, damage bins strictly increasing
  within a source damage bin (so a repeated triple cannot be silently reduced to its last row), and
  each defined column summing to 1 within 1e-6 — the same checks `vulnerability.csv` gets, sharing
  the same implementation.
- **Inputs the pipeline already tolerated still work.** A keys file may hold several rows for one
  `(loc_id, peril_id, coverage_type_id)`: they share an `item_id` and are deduplicated at the end of
  `get_gul_input_items`. Link resolution runs against that same first-occurrence view, so
  configuring the feature does not turn an accepted input into a crash. An item whose configured
  source coverage is absent at its location is likewise left unpaired rather than raising.
- 3274 tests pass (6 skipped) across the whole `tests/` tree.
- **`tests/assets/test_model_8`**: a reference model run by `test_gulmc` across every sample size,
  back-allocation rule, `ignore_correlation` setting, random generator and effective-damageability
  mode (144 golden files). A two-deep dependency chain rooted on an uninsured coverage, alongside an
  independent one, with damage group ids deliberately mixed so both the coupled and uncoupled
  sampling cases are covered. See its `README.md`.

### Note for modellers: keeping the authored conditional probabilities
A dependent draws its damage sample from the same per-item random stream as any other item, and its
source used that stream to position itself within its own damage bin. Where the two draws are
coupled, the conditional probabilities the engine realises are **not** the ones written in
`conditional_vulnerability`. Two settings control that, and both are the modeller's choice: sharing
a `group_id` (the default `damage_group_id_cols` do not include coverage type, so all coverages at a
location share a stream) and a non-zero `damage_correlation_value` between the source and dependent.
Measured worst deviation from a file authored at 0.500, over 20 000 samples: 0.019 with distinct
group ids and no damage correlation, 0.142 at `damage_correlation_value = 0.3`, 0.343 at 0.7, and
0.461 with a shared group id. This is documented on the explanation page.

- **New feature — gulmc coverage dependency.** A coverage can be configured (via
  `model_settings.coverage_dependency_settings`) to depend on a source coverage at the same location
  and peril; the dependent's damage is driven by the source's sampled damage bin through a new
  `conditional_vulnerability` damage-transition matrix. Opt-in; unconfigured runs are unaffected.
  The correlations file gains a `source_item_id` column in place of `source_coverage_id` (same
  record size), and `conditional_vulnerability` is supported by `csvtobin` / `bintocsv` as the
  `conditionalvulnerability` file type.
- **Fix — return-period protection under a dynamic footprint** is now applied to the whole item
  rather than per sample. Previously, for an areaperil with more than one footprint hazard record, a
  protected item reported non-zero mean, standard deviation and maximum loss alongside all-zero
  samples; it now reports zero for those and for its TIV. Because the TIV split is applied per row
  across a coverage's items, an *unprotected* item sharing a coverage with a protected one may now
  report a **higher** mean and maximum loss under back-allocation rules 1/2/3, since the protected
  item no longer takes a share of the split. Per-sample losses are unchanged. Results change for
  existing dynamic-footprint models that use `ReturnPeriod`; single-hazard-record areaperils are
  unaffected.
- **Fix — logging after a gulmc or gulpy run.** `redirect_logging` left every `oasislmf.*` logger
  pinned at the run's log level, silently suppressing INFO and DEBUG from those modules for the rest
  of the process (for example in `oasislmf model run`, which continues with output steps after the
  engine). The logger's level and propagation are now both restored.


sstruzik:
Docs note (for whoever rebases this after the docs restructure lands):

This branch carries its own coverage-dependency docs at `docs/source/explanation/coverage-dependency.rst`, in the **current (pre-restructure)** docs layout. The docs-migration PR (#2065) deliberately does **not** document this feature — its copy was removed there so the page ships with the code here (avoids documenting code that is absent from `main`).

When this PR is rebased/merged onto the restructured `main` (after #2065), align this page to the new Diátaxis layout:
- it already lives under `docs/source/explanation/`, so it is mostly a format/wiring touch-up, not a rewrite;
- convert/keep it consistent with the new `explanation/` pages (MyST `.md` alongside the others) and add it back to `docs/source/explanation/index.md` toctree;
- re-add the "Enable coverage dependency" how-to section to `docs/source/how-to/ground-up-losses.md` (removed from #2065 along with the page).

No action needed until the rebase.

oasislmf/pytools/getmodel/manager.py
SkylordA, line 1: Bug: `get_vulns()` (in `getmodel/manager.py`, changed by this PR) gained an `allow_missing_vuln_ids` param so conditional-only vulnerability ids can be absent from `vulnerability.bin`. The PR updates the call site in the new gulmc path (`gulmc/structure.py:585`, also changed) to pass it, but there's a second, pre-existing production call site of the same function — `getmodel/structure.py:101`, backing the classic, non-gulmc `gul` engine (still the default unless gulmc is explicitly selected) — that this PR never touches, so it still calls with the old default (`None`). Any model with `coverage_dependency_settings` configured will crash this path with `Exception("Vulnerability_ids {...} are missing from ...")` (`getmodel/manager.py:552/612`), since `gul_inputs.py` (changed by this PR) now unconditionally writes dependent items pointing at conditional-only ids regardless of which engine will consume them. Thread the same kwarg through this call site too, or reject the config combination explicitly at settings-validation time.

oasislmf/preparation/correlations.py
```diff
@@ -41,3 +43,48 @@ def map_data(data: Optional[dict], logger) -> Optional[pd.DataFrame]:
...
+        except (KeyError, TypeError, ValueError) as e:
+            raise OasisException(f"Invalid coverage_dependency_settings entry {entry}: {e}")
+        if source_cov_type == dependent_cov_type:
+            raise OasisException(
+                f"Invalid coverage_dependency_settings entry {entry}: a coverage type cannot depend on itself.")
+        if dependent_cov_type in seen_dependents:
+            raise OasisException(
+                f"Invalid coverage_dependency_settings: coverage type {dependent_cov_type} is listed as a dependent "
+                "more than once; each dependent coverage type must have exactly one source.")
+        seen_dependents.add(dependent_cov_type)
```
SkylordA: Bug: only rejects a coverage type depending on itself (81) and a dependent listed twice (84-87). A 2-cycle across types — `[{source: B, dependent: A}, {source: A, dependent: B}]` — passes both checks (each entry has a distinct single source). It's still caught, but only later in `gulmc/structure.py`'s `_validate_acyclic_coverage_dependency`, with a `coverage_id`-based error that doesn't point back at the settings. Add a cycle check over the (source, dependent) pairs here to fail at config time.

oasislmf/pytools/converters/bintocsv/utils/conditional_vulnerability.py
```diff
@@ -0,0 +1,33 @@
...
+
+from oasislmf.pytools.common.data import (
+    resolve_file, write_ndarray_to_fmt_csv, vulnerability_bin_header_type,
+)
+from oasislmf.pytools.converters.data import TOOL_INFO
+
+DEFAULT_BUFFER_SIZE = 1000000
+
+
+def conditional_vulnerability_tocsv(stack, file_in, file_out, file_type, noheader):
```
SkylordA: Reuse: near-duplicate of the no-idx branch of `vulnerability_tocsv` (`bintocsv/utils/vulnerability.py:146-157`) — same header-skip/memmap/stdin/chunked-write pattern, with an unexplained different batch size (`1_000_000` vs `1<<13`) now drifting independently. Conditional vulnerability never has an idx/zip variant, so delete this file and call `vulnerability_tocsv(stack, file_in, file_out, file_type, noheader, idx_file_in=None, zip_files=False)` directly from the dispatch entry. (The encode side, `csvtobin/utils/vulnerability.py`, did this correctly by generalizing `_validate_chunk` with a `fields` param instead of duplicating — same fix, applied the other direction.)

oasislmf/pytools/gulmc/structure.py
```diff
@@ -56,15 +60,344 @@
...
+    damage_bin = recs['damage_bin']
+    if source_bin.min() < 1 or source_bin.max() > num_damage_bins \
+            or damage_bin.min() < 1 or damage_bin.max() > num_damage_bins:
+        raise OasisException(
+            f"conditional_vulnerability bins must be in 1..{num_damage_bins}: source damage bin "
+            f"range [{int(source_bin.min())}, {int(source_bin.max())}], damage bin "
+            f"range [{int(damage_bin.min())}, {int(damage_bin.max())}]."
+        )
+
+    cond_vuln_ids = np.unique(recs['vulnerability_id'])
```
SkylordA, line 270: Efficiency: `id_to_idx = {int(v): i for ...}` + `for r in recs: conditional_vuln_array[id_to_idx[int(r['vulnerability_id'])], ...] = r['probability']` is a plain-Python, per-row dict-lookup loop over the whole conditional-vulnerability file at model load — no `@njit`, no vectorization. `cond_vuln_ids` is already sorted (`np.unique`), and the file's own `build_vuln_idx_to_cond_idx` two functions below (line 397) already does `np.searchsorted(cond_vuln_ids, ...)` for the equivalent lookup — reuse that here: `idx = np.searchsorted(cond_vuln_ids, recs['vulnerability_id'])` then one fancy-index assignment.

oasislmf/pytools/common/data.py
SkylordA, line 1: `correlations_dtype` grew 20→24 bytes/record (`source_item_id` added) with no format-version check; `read_correlations` (`common/input_files.py:143-148`) will hard-fail or silently misread a pre-PR 20-byte `correlations.bin` if one is ever loaded against this code. Low risk since the file is regenerated per-run rather than persisted across upgrades — worth a one-line confirmation that no workflow reuses a cached run dir across the version bump.

oasislmf/preparation/gul_inputs.py
```diff
@@ -406,6 +444,71 @@ def get_gul_input_items(
...
+        if source_items.empty or not dep_mask.any():
+            continue
+        # left merge preserves the order of the dependent rows, so the result aligns
+        # positionally with gul_inputs_df.loc[dep_mask]
+        merged = gul_inputs_df.loc[dep_mask, ['loc_id', 'building_id', 'peril_id', 'areaperil_id']].merge(
+            source_items, on=['loc_id', 'building_id', 'peril_id'], how='left')
+        # A source in another cell cannot drive this dependent, so leave it unpaired. Not an error
+        # here: only the model's static data says whether the item's vulnerability is a conditional
+        # one, so gulmc decides (validate_coverage_dependency).
+        mismatch = merged['_src_item_id'].notna() & (merged['_src_areaperil_id'] != merged['areaperil_id'])
```
SkylordA, line 529: A dependent whose configured source resolves to a different `areaperil_id` (key-server mismatch) is silently downgraded to independent with only `logger.info(...)` — no warning, no aggregate count anywhere else. Intentional per the inline comment, but a config mismatch that changes computed losses is arguably worth a WARNING or a surfaced count rather than routine INFO noise.

oasislmf/preparation/gul_inputs.py
```diff
@@ -370,8 +379,37 @@ def get_gul_input_items(
...
         gul_inputs_df.loc[mask, 'tiv'] = gul_inputs_df.loc[mask, tiv_col['tiv_col']]
 
-    # Filter out rows with zero TIV
-    gul_inputs_df = gul_inputs_df[gul_inputs_df['tiv'] > 0]
+    # Keep a zero-TIV coverage that is a dependency source for a kept dependent, so an uninsured
+    # building can still drive its contents. Tested on (loc_id, peril_id, areaperil_id) — the triple
+    # a dependent pairs its source on — so a source no dependent can pair with is still dropped.
+    # Resolved to a fixed point, not one pass per configured pair: in a chain
+    # building -> contents -> BI with only BI insured, keeping contents also requires the building.
+    keep_zero_tiv_source = pd.Series(False, index=gul_inputs_df.index)
```
SkylordA: Re-scans the full dataframe once per `(pass, pair)` for `len(coverage_dependency_settings)` passes to reach a fixed point, when the settings list is small and static — a one-time topological sort (deepest-dependent-first) would need exactly one pass per pair instead of up to `len(pairs)²`. Not currently expensive (list is ~2-4 entries); flagging as avoidable recomputation, not a bug.

oasislmf/preparation/gul_inputs.py
SkylordA, line 1: The zero-TIV retention, per-item `source_item_id` linking, and coverage-level consistency check (~100 lines of self-contained logic) are inline in this already-large function, untestable without the full item-generation pipeline. This same PR extracts equivalent-complexity logic cleanly elsewhere (`gulmc/structure.py`'s `build_coverage_dependency_forest`, `get_conditional_vulns`, etc.) — worth the same treatment here, lower priority since it's pre-existing function size the PR only adds to.

oasislmf/preparation/gul_inputs.py
sstruzik: The areaperil_id and the vulnerability_id are set by the key server.
dependent and independent vulnerability_id are then define in the model data in 2 different files.
If there is an accidental mismatched in the areaperil_id, the item is set to be independent but the vulnerability stay the same. And later in the run gulmc raise an error because a dependent vuln is used without dependent item.
That is why I downgrade the logging level, there won't be a case where this "setting to independent" will be silently missed.
I've updated the doc to reflect this more clearly

oasislmf/pytools/common/data.py
sstruzik: `correlations.csv`/`.bin` are analysis inputs, not model static data — `write_gul_input_files` writes them into the run's input directory alongside `items.bin` and `coverages.bin` (`gul_inputs.py:627-628`) every time Oasis files are generated. A 20-byte record can therefore only meet a 24-byte reader by pointing a newer engine at a run directory generated by an older `oasislmf`, which isn't a supported flow — the input set moves as a unit, and this PR doesn't change that.

If we do want to support mixed vintages, the marker belongs on the input set as a whole rather than on this one dtype. Happy to raise that separately; I don't think it needs to block here.

oasislmf/preparation/gul_inputs.py
sstruzik: Measured it before deciding. The early `break` plus the `range(len(pairs))` bound means the shipped single-pair configuration already does exactly **1** scan — the `len(pairs)²` worst case (9) needs a 3-deep chain across 4 coverage types.

I did build your topological version and checked it over every valid forest on 4 coverage types × every zero/positive TIV mask × every input ordering (7,872 cases): identical retention sets, worst case 9 → 3 scans. So the suggestion is correct. But it needs a `source_of` map and a depth walk, taking the block from 22 to ~29 lines, and it trades the fixed point's order-independence for 6 boolean ops — so I've left it.

I took the free half in c1d8b529f: `kept` is recomputed per pair and retention accumulates into one series instead of two. Same line count, one fewer accumulator, −3.5% scans (worst case unchanged).

oasislmf/preparation/gul_inputs.py
sstruzik: Took two of the three in a5a2271a6 — `resolve_zero_tiv_dependency_sources` and `validate_single_source_per_coverage` are now module-level and directly callable. `get_gul_input_items` goes 443 → 400 lines.

The criterion I used is whether a block's precondition can be stated as **data** rather than as **position**. Both of these qualify: retention needs only `tiv`, `coverage_type_id`, `loc_id`, `peril_id`, `areaperil_id`; the check needs only `source_item_id`, `item_id`, `coverage_id`. Each signature is honest about what it requires.

The per-item linking block stays inline, because it fails that test: it needs `coverage_id` already assigned *and* must resolve against the same first-occurrence view that `drop_duplicates(subset='item_id')` applies later in the same function. Its docstring would have to encode a position in the pipeline, which hides the coupling rather than removing it — and that coupling is exactly where the duplicate-keys bug in cd9e9b0e6 came from, so I'd rather not churn it here.

No behaviour change, and the existing tests still drive `get_gul_input_items` end to end.

oasislmf/pytools/getmodel/manager.py
sstruzik: Good catch — fixed in 9188826b8, taking your second option (reject the combination) rather than threading the kwarg.

`getmodel/structure.py` now reads the conditional vulnerability ids and refuses up front if any item uses one, naming them and pointing at gulmc, instead of failing later with a bare "Vulnerability_ids ... are missing". Threading `allow_missing_vuln_ids` through would have let the classic engine load a damage-transition matrix it has no source damage bin to index, so failing loudly is the right outcome here.

oasislmf/preparation/correlations.py
sstruzik: Correct — a 2-cycle passed both checks. Cycle detection added in 9188826b8 and folded into the parse loop in 009df66ae, so `get_coverage_dependency_settings` now walks each new edge to its root and rejects anything that closes a cycle, naming the coverage types involved.

a700a1287 adds doctests covering it, including a multi-type loop.

oasislmf/pytools/converters/bintocsv/utils/conditional_vulnerability.py
sstruzik: Agreed — deleted in 9188826b8. The dispatch entry now calls `vulnerability_tocsv` directly with `idx_file_in=None, zip_files=False`, so there's one implementation and no batch size drifting independently.

oasislmf/pytools/gulmc/structure.py
sstruzik: Fixed in 9188826b8 — the per-row loop is gone, replaced by a single vectorised scatter:

```python
conditional_vuln_array[np.searchsorted(cond_vuln_ids, recs['vulnerability_id']),
                       recs['damage_bin'] - 1,
                       recs['source_damage_bin'] - 1] = recs['probability']
```

oasislmf/pytools/common/data.py
```diff
@@ -128,6 +132,7 @@ def def_to_type_and_size(def_str):
 share2 = ("share2", oasis_float, "%f")
 share3 = ("share3", oasis_float, "%f")
 sidx = ("sidx", 'i4', "%d")
+source_item_id = ("source_item_id", 'u4', "%u")
```
vinulw: minor: type mismatch between `item_id` and `source_item_id`. For consistency they should be the same

oasislmf/preparation/gul_inputs.py
```diff
@@ -120,6 +123,88 @@ def process_group_id_cols(group_id_cols, exposure_df_columns, has_correlation_gr
 
 
 @oasis_log
```
vinulw: oasis_log should be applied to `get_gul_input_items` not this function

oasislmf/pytools/common/data.py
sstruzik: Agreed — fixed in cda266949. `source_item_id` is now `'i4'`/`"%d"`, matching the `item_id` it holds.

The correlations record stays 24 bytes so there's no layout or compatibility change, and it removes a signed/unsigned mix that `np.searchsorted` in `build_coverage_dependency_forest` was silently upcasting. Test fixtures mirroring the items layout are updated with it; `coverage_id` is left as `'u4'`, which is its production type.

oasislmf/preparation/gul_inputs.py
sstruzik: You're right, and this was a regression I introduced — thanks for spotting it. Fixed in cda266949.

a5a2271a6 extracted the two helpers by inserting them above `def get_gul_input_items(`, which sits *below* its `@oasis_log` decorator. So the decorator ended up on `resolve_zero_tiv_dependency_sources` and `get_gul_input_items` lost it — dropping the entry/timing log line for the main file-generation function and adding a spurious one for the helper. The test suite couldn't catch it, since `oasis_log` only logs and doesn't change results.

SkylordA (approved):

Ha-Ree (approved):

PR #2065, opened 2026-07-17
docs: Diátaxis restructure + ktools→pytools kernel + executable notebooks

sstruzik:
Restructures the OasisLMF documentation into a Diátaxis set co-located with the code, drains the ktools kernel docs in and rewrites them against `pytools`, and adds executable tutorials.

Part of the cross-repo Oasis documentation restructure (Option B): each repository owns the docs for what it owns, and `GenerateDocs` (#54) aggregates them into the published site.

## Why

Two overlapping documentation systems had never been reconciled. The high-value conceptual content lived in a *separate repository* from the code it described, so it drifted; the in-repo system auto-generated an unstructured 519-page API dump with no narrative; and both had, at different times, deployed to the same GitHub Pages target. This PR consolidates on one in-repo, Diátaxis-structured Sphinx project.

## What this adds

- **Diátaxis structure** — tutorials, how-to, reference, explanation — with audience landing pages, and the shared Oasis branding used across all six component sites.
- **Conceptual content migrated in from GenerateDocs and co-located with the code it describes**: the financial module (plus the previously-orphaned `fm_architecture.md`), modelling and sampling methodology, correlation, disaggregation, vulnerability adjustments, absolute damage, post-loss amplification, pre-analysis adjustments, currency conversion, keys/lookup, geocoding, model data formats and the MDK guides.
- **The ktools kernel docs drained in and rewritten for pytools** — `eve`→`evepy`, `getmodel`→`modelpy`, `gulcalc`→`gulpy`/`gulmc`, `fmcalc`→`fmpy`, `summarycalc`→`summarypy`, the output tools and the converters — with the CLI text taken from the real `--help` rather than the retired C++ binaries. ktools is being decommissioned and owns no docs.
- **Autoapi scoped per subsystem** instead of a whole-package dump: `pytools/fm`, `gulmc`, `common`, the output modules (`elt`, `lec`, `plt`, `aal`, `summary`, `pla`) and `oasislmf/lookup`. The build is seconds, not minutes.
- **Executable tutorials** (myst-nb): explore model data, and analyse ORD results from an ELT to an EP curve. They run at build time, so their output cannot silently rot, and each publishes a downloadable `.ipynb`.
- **Every page builds warning-free** — including 13 product docstrings reformatted to valid RST, verified AST-identical to their originals with docstrings blanked, so no behaviour changed.

### Kernel command examples that could not work (@Ha-Ree)

Every finding was verified against the real CLI, and several were worse than reported:

- **`summarypy` has no stdout path.** Its `-<id> <path>` arguments go straight to `open(path, 'wb')`, so `-2 -` created a file literally named `-` and the downstream step read nothing. Every `summarypy | eltpy/pltpy` pipeline is now two steps, and the prose says why.
- **`katpy -m` is MELT, not MPLT** (`-M` is MPLT), so the example concatenated the wrong tables.
- **`bintocsv summarycalc` does not exist** — the mapping table and examples now say so explicitly, and point at `eltpy`/`pltpy`.
- **`csvtobin gul` requires `-S`; `csvtobin footprint` requires `-x` and `-m`** — the examples exited 2 as written.
- **`RandomNumbers.md` documented flags that do not exist at all.** `gulmc` has no `-R`, no `-r`, no `-s`, no random-number-file input and no seed option; `-S00` was a typo and `-i -` raises `FileNotFoundError`. `ridx` appears nowhere in pytools either, so the methodology built on a shared indexed buffer was ktools-only. Rewritten from `pytools/gul/random.py`: seeds derived per `(group_id, event_id)` so runs repeat without a seed flag, hazard sampling with its own constants keyed off `hazard_group_id` (never previously mentioned), `--random-generator {0,1,2}`, and `-d {1,2}` to inspect the numbers drawn. The three legacy modes survive as a marked historical note.
- **`--gulmc` and `--do-disaggregation` are already `default: True`** — both sections are reframed as opt-out.
- **Allocation rule 3 added**, with the multiplicative behaviour distinguished from rules 1-2.
- **`Specification.md` contradicted itself.** The prose was corrected but the stream table still declared a cdf input for `gulmc`. Checking the neighbours showed the same defect twice more, all from one cause — the table had no vocabulary for the event stream. `evepy`'s `stream_events()` writes only 4-byte ids with **no header**, so it carries no stream_id; `modelpy` reads that on stdin yet the table said `none`; and `evepy` had no row at all. All three fixed, a `gulpy` row added so the cdf stream has a consumer, and stream 1 relabelled `gul` — it is the legacy stream that is deprecated, not gulmc.

### Content that documented things which no longer exist (@SkylordA)

- **`mdk-builtin-lookup.rst` removed.** All 348 lines described a superseded three-class API (`OasisPerilLookup`, `OasisVulnerabilityLookup`, `OasisLookup` via `OasisLookupFactory`) with IPython transcripts and a class diagram, plus a workflow built on `oasislmf model generate-peril-areas-rtree-file-index` — a command that is gone with **no successor**: `build_rtree` reads a user-supplied geometry file and joins locations at runtime with geopandas, so there is no index artefact. There was nothing to repoint; `explanation/keys-service.rst` documents the current design, and the how-to index points there. `DEFAULT_RTREE_INDEX_PROPS` — dead configuration for that removed artefact — was dropped too, after confirming zero references in this package or any sibling repository.
- **`reference/outputs/results.rst` restructured.** It led with ~550 lines of deprecated formats under a heading that collided with `reference/kernel/OutputComponents.md`, and the ORD section introduced itself as an afterthought. ORD now leads and states it is the current standard, organised by table (SELT; MELT/QELT; SPLT/MPLT/QPLT; EPT/PSEPT; ALT/PALT) rather than by retired binary, with a lead-in mapping each `ord_output` flag to its table and the pytools component that emits it. Legacy formats moved to the end behind an explanatory note; all 63 ktools command invocations removed, since `kernel/OutputComponents.md` documents the tools. 1434 → 1157 lines. The 15 field-specification tables are **kept**: they give name, type, byte width and example — the encoding Oasis writes — which the generated ORD reference does not cover.
- **The stale autoapi note** claimed two subsystems when `conf.py` had grown to ten.
- **`financial-module.rst`** — ten lines still said `fmcalc`/`summarycalc`; flags checked against `fm/cli.py` before rewriting.

### Structure and links (@vinulw)

- **How-to guides had no sidebar dropdown** because the other three sections point at a `*/index` page and How-to listed eight leaves, with no `how-to/index` existing. Added — and the three pages that sat at the docs root (`installation`, `building-and-running-models`, `logging-configuration`) moved under `how-to/`, which made four references *shorter* and only two longer.
- **`event_stream.py` was missing from the reference** — `pytools/common` is now in `autoapi_dirs`. That took the build from 0 to 17 warnings, so its docstrings were swept to valid RST in the same commit.
- **Google-style `Returns:` sections fixed** in `gulmc/aggregate.py` and `pytools/common/input_files.py`: napoleon treats everything before the first colon as the return *type*, so named returns turned a name into a type and emitted a cross-reference.
- Missing geocoding link, `DOCS_STRATEGY.md` made a real link, and five mislabelled code blocks in `keys-service.rst` — fixing which exposed a **genuine bug**: two snippets had `pd.concat([loc_missing, locations[~mask])`, never closing the list.
- **The planning documents were reconsidered.** Neither `DOCS_STRATEGY.md` nor `MIGRATION_PLAN.md` exists on `main`, so this PR was *introducing* 807 lines of planning material. The durable half — architecture, content model, tooling, authoring traps — is now a published page, `explanation/documentation-strategy`; the transient half is dropped, with its content in history at `7b9643178`. Writing it up meant checking the claims, and several "decisions" had never been implemented: there is no `linkcheck` job, `interrogate` and mermaid are not configured, and there is no render check. Those are now listed as *agreed but not yet implemented* rather than published as guidance.

### Publishing

- **`build-and-deploy-docs.yml` removed.** It deployed this repository's own Pages site on every push to `main` — live today at `oasislmf.github.io/OasisLMF/`, serving the old autoapi dump. Left in place, merging this PR would publish the same docs twice on one domain at URLs differing only by capitalisation, with the in-repo copy usually fresher and its cross-component links pointing at paths that do not exist. `build-docs.yml` is kept as the PR-time build, with its `contents: write` and Pages-artifact upload reduced to what a build check needs.
- **Dead external links repointed** (found by the link check in GenerateDocs#55): the deleted `main-platform1` branch, `libspatialindex.github.io/index.html`, and the geopandas course, which moved to Read the Docs.
- **The MDK workflow diagram reinstated.** It was the only image in GenerateDocs' clear-out that is still accurate and had no copy anywhere. Checked before reinstating: it names no ktools binaries, only MDK commands that all still exist. It sits in the MDK how-to beside the module diagram, with a caption stating what the prose never did — that `oasislmf model run` is the three `generate-*` commands in sequence.

Build is **0 warnings** standalone and in the aggregated build, with both notebooks executing.

Restructure the OasisLMF documentation (Diátaxis, co-located with the code), drain and rewrite the ktools kernel reference for pytools, and add executable tutorials.


docs/source/reference/kernel/Workflows.md
```diff
@@ -0,0 +1,101 @@
...
+
+### 3. Loss exceedance curves (EPT)
+
+`lecpy` (like `aalpy`) is not a stream stage — it reads all of a summary set's binaries
+from `work/`, since EP curves are not valid on an event subset. Write the summary
+binaries over multiple partitions, then run `lecpy` once:
+
+```bash
+evepy 1 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 - > work/summary2/p1.bin
+evepy 2 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 - > work/summary2/p2.bin
```
Ha-Ree: `summarypy` has no stdout path. Its positional `summary_sets_output` args are `-<id> <path>` pairs and each path is opened with `open(path, 'wb')` (`summary/manager.py:394`), so `-2 -` creates a file literally named `-` in the cwd and the `>` redirect captures nothing — `work/summary2/p1.bin` is left empty and the `lecpy`/`aalpy` steps below read zero-byte inputs.

```suggestion
evepy 1 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 work/summary2/p1.bin
evepy 2 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 work/summary2/p2.bin
```

Same at lines 55-56.

Ha-Ree (commented):
Reviewed the docs build and every kernel command example against the current pytools CLIs.

**The infrastructure side is sound.** I built the docs locally (`sphinx-build -b html source build`, myst-nb executing the notebooks): the build succeeds with **zero warnings**, the notebooks execute cleanly, the `conf.py` autoapi scoping and `reference/api/...` toctree entries resolve, and every `{doc}`/`:ref:`/relative link target resolves. The Python diff under `oasislmf/lookup/` and `oasislmf/pytools/**` is docstring-only — I checked every hunk, no executable code is touched. New `requirements.in` entries are picked up by the docs workflows' pip-compile, and `matplotlib` for the notebooks already arrives via `optional-package.in`.

**The findings are all documented commands that don't run.** I verified each by invoking the CLI or reading the argparse/manager source. Two patterns account for most of them:

- **`-` used as a stdout placeholder.** `summarypy` has no stdout path — each `-<id> <path>` pair is opened with `open(path, 'wb')`, so `-2 -` silently creates a file named `-` and the pipe or redirect downstream gets nothing. Six sites.
- **Required arguments omitted**, so the example exits 2: `csvtobin gul` (`-S`), `csvtobin footprint` (`-x`, `-m`), plus `bintocsv summarycalc`, which isn't a valid sub-command at all.

Then four content errors: the `eve | modelpy | gulmc` pipeline (wrong binary, wrong engine, `-i -`), the `gulmc` stdin description in `Specification.md`, `--gulmc`/`--do-disaggregation` presented as opt-in when both default to `True`, `katpy -m` vs `-M`, and allocation rule `3` missing from the `-a` table.

Worth flagging as a pattern: the kernel examples read as translated from their ktools equivalents rather than run against pytools. A CI smoke test that executes the fenced `bash` blocks (or at least `--help`-checks the flag sets) would catch this class cheaply and keep it caught.

Verified correct along the way, for what it's worth: `-C/--config`, `EPType`/`EPCalc` values, `gul_random_generator` default `2`, all `ord_output` flag names, `gul_S1_ept.csv` naming, the `lecpy`/`aalpy`/`eltpy`/`pltpy`/`evepy`/`fmpy` flag lists, and the `analysis_settings` example.

docs/source/reference/kernel/Workflows.md
```diff
@@ -0,0 +1,101 @@
...
+## Single-output workflows
+
+### 1. Insured-loss event loss table (ELT)
+
+Run ground-up → FM → summary (portfolio summary set 2) → ELT, per partition, then
+concatenate:
+
+```bash
+evepy 1 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 - | eltpy -s elt_p1.csv
+evepy 2 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 - | eltpy -s elt_p2.csv
```
Ha-Ree: Same root cause as the `-2 -` usage below, but here it also breaks the pipe: `summarypy` writes the summary set to a file called `-` and emits nothing on stdout, so `eltpy` gets an empty stream. `eltpy` does read stdin by default (`-i/--files_in` defaults to `["-"]`), but `summarypy` can't feed it, so this has to be two steps:

```suggestion
evepy 1 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 summary_p1.bin
evepy 2 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 summary_p2.bin
eltpy -i summary_p1.bin -s elt_p1.csv
eltpy -i summary_p2.bin -s elt_p2.csv
```

Same at lines 34-35 (`pltpy`). The `tee >(...)` variant at 68-69 needs a bigger rethink for the same reason — neither branch can stream out of `summarypy`.

docs/source/reference/kernel/OutputComponents.md
```diff
@@ -0,0 +1,163 @@
...
+**MELT** (moment ELT — mean and standard deviation by event/summary) and **QELT**
+(quantile ELT).
+
+**Parameters** — `--run_dir`; `-i, --files_in`; `-s SELT`, `-m MELT`, `-q QELT` (output
+files for each table); `-E {csv,bin,parquet}`; `-H` (no header).
+
+**Usage**
+
+```bash
+summarypy -t gul -1 - < gul.bin | eltpy -s gul_selt.csv -m gul_melt.csv
```
Ha-Ree: `summarypy` can't write to stdout, so this pipe hands `eltpy` an empty stream — `-1 -` just creates a file named `-`. `summarypy` does have `-i/--files-in`, so:

```suggestion
summarypy -t gul -i gul.bin -1 summary_gul.bin
eltpy -i summary_gul.bin -s gul_selt.csv -m gul_melt.csv
```

Same at line 48 for `pltpy`.

docs/source/reference/kernel/OutputComponents.md
```diff
@@ -0,0 +1,163 @@
...
+
+**Parameters** — `-o, --out` (output file); `-f, --file_type {csv,parquet,bin}`;
+`-i, --files_in` (explicit files) or `-d, --dir_in` (a directory); one flag per ORD
+table type (`-s` SELT, `-m` MELT, …) to select what to concatenate.
+
+**Usage**
+
+```bash
+katpy -s -d work/kat/gul_S1_elt_sample -o gul_S1_selt.csv          # concat SELT partitions (csv)
+katpy -m -f parquet -i mplt_P1.parquet mplt_P2.parquet -o MPLT.parquet
```
Ha-Ree: `-m/--melt` is MELT; MPLT is `-M/--mplt` (`kat/cli.py:36,39`). As written this concatenates the MPLT parquet files as if they were MELT.

```suggestion
katpy -M -f parquet -i mplt_P1.parquet mplt_P2.parquet -o MPLT.parquet
```

docs/source/reference/kernel/StreamConversionComponents.md
```diff
@@ -0,0 +1,91 @@
...
+**Usage**
+
+```bash
+evepy 1 1 | gulmc -S100 -a1 | fmpy -a2 | bintocsv fm -o il.csv
+bintocsv fm -i il.bin -o il.csv
+```
+
+**Output fields** — `event_id`, `output_id`, `sidx`, `loss`.
+
+## bintocsv summarycalc
```
Ha-Ree: There is no `summarycalc` sub-command — `bintocsv --help` lists `{aal, aalmeanonly, aggregatevulnerability, alct, amplifications, cdf, complex_items, correlations, coverages, damagebin, ept, eve, fm, fm_policytc, fm_profile, fm_profile_step, fm_programme, fm_summary_xref, fm_xref, footprint, gul, gul_summary_xref, items, lossfactors, melt, mplt, occurrence, periods, psept, qelt, qplt, quantile, random, returnperiods, selt, splt, vulnerability, weights}`, and `bintocsv summarycalc` exits 2 with "invalid choice".

This affects the mapping table row at line 12 (`summarycalctocsv` → `bintocsv summarycalc`) and both examples at 68-69. What's the intended replacement for `summarycalctocsv`? If there isn't a direct one, the table row is probably better as "no direct equivalent — use `eltpy`/`pltpy` for ORD output" than as a command that doesn't exist.

docs/source/reference/kernel/StreamConversionComponents.md
```diff
@@ -0,0 +1,91 @@
...
+
+Converts a ground-up loss CSV back into the binary loss stream (e.g. to pipe into
+`fmpy`). The CSV has `event_id`, `item_id`, `sidx`, `loss`; for each event/item a
+`sidx = -1` (mean) and `sidx = -2` (std dev) record must be present (even if zero),
+while zero-loss positive samples may be omitted.
+
+**Usage**
+
+```bash
+csvtobin gul -i gul.csv -o gul.bin        # sample size / stream-type options as needed
```
Ha-Ree: `-S/--max_sample_index` is required (`usage: csvtobin gul [-h] [-t STREAM_TYPE] -S MAX_SAMPLE_INDEX ...`), so this exits 2 as written.

```suggestion
csvtobin gul -i gul.csv -o gul.bin -S 100        # -S/--max_sample_index is required
```

docs/source/reference/kernel/ValidationComponents.md
```diff
@@ -0,0 +1,73 @@
...
+- **`csvtobin damagebin`** — the first `bin_index` is 1; bin indices are contiguous; each
+  interpolation (mean) damage lies within its bin range. *(Warning if the lower limit of
+  the first bin is not 0.)*
+- **`csvtobin footprint`** — the probabilities for each `(event_id, areaperil_id)` group
+  sum to 1; rows are sorted by `event_id` then `areaperil_id` ascending.
+- **`csvtobin vulnerability`** — `vulnerability_id`s are ascending; the probabilities for
+  each `(vulnerability_id, intensity_bin_id)` group sum to 1; damage bins are contiguous.
+
+```bash
+csvtobin footprint -i footprint.csv -o footprint.bin        # validates by default
```
Ha-Ree: `csvtobin footprint` requires both `-x/--idx_file_out` and `-m/--max_intensity_bin_idx` (`usage: csvtobin footprint [-h] -x IDX_FILE_OUT [-z] -m MAX_INTENSITY_BIN_IDX ...`), so this exits 2 before validating anything.

```suggestion
csvtobin footprint -i footprint.csv -o footprint.bin -x footprint.idx -m 50        # validates by default
```

(`-N/--no_validation` is the flag that turns the validation off, which is worth naming here given the section title.)

docs/source/how-to/ground-up-losses.md
```diff
@@ -0,0 +1,109 @@
...
+  oasislmf model run --gulmc --gulmc-vuln-cache-size 500 -C oasislmf.json
+  ```
+
+## Run gulmc directly in a kernel pipeline
+
+For low-level runs, `gulmc` reads an event stream and writes a GUL stream, like
+the other kernel components (see {doc}`../reference/kernel/CoreComponents`):
+
+```bash
+eve 1 1 | modelpy | gulmc -S 100 -a 0 --random-generator 2 -i - -o gulmc.bin
```
Ha-Ree: Three problems in one line:

1. `eve` is the ktools binary — the pytools one is `evepy`, and this page is otherwise all-pytools.
2. `gulmc` must not be fed `modelpy` CDFs. It reads the event stream and loads footprint/vulnerability itself; piping CDFs into it is the `gulpy` pipeline (see `CoreComponents.md:96-99`, which has this right).
3. `-i -` reaches `open('-', 'rb')` → `FileNotFoundError`. `gulmc` reads stdin when `-i` is omitted.

```suggestion
evepy 1 1 | gulmc -S 100 -a 0 --random-generator 2 -o gulmc.bin
```

docs/source/how-to/ground-up-losses.md
```diff
@@ -0,0 +1,109 @@
...
+these options see the {doc}`../explanation/index` pages; for the full option list
+see {doc}`../options_config_file` and {doc}`../generated_options`.
+
+These options can be passed as CLI flags to `oasislmf model run` **or** set in the
+run configuration / analysis settings JSON (same names, with underscores). See
+{doc}`../building-and-running-models` for the base run command.
+
+## Select gulmc as the ground-up engine
+
+Turn on the full Monte-Carlo Python engine:
```
Ha-Ree: `gulmc` is already the default — `{'name': 'gulmc', 'default': True, ...}` (`computation/generate/losses.py:426`, and again at :645). So `--gulmc` isn't turning anything on; the flag users actually need is `--gulmc False` to opt *out* and fall back to `gulpy`. `CoreComponents.md:18` describes it correctly as the default engine, so these two pages disagree.

Same for `--do-disaggregation` at line 62: `default: True` in `computation/generate/files.py:112`.

Suggest reframing both sections as "on by default, here's how to turn it off".

docs/source/reference/kernel/Specification.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+### modelpy 
+
+modelpy is the component which generates a stream of effective damageability cdfs for a given set of event_ids and the impacted exposed items on the basis of their areaperil_ids (location) and vulnerability_ids (damage function). 
+
+
+### gulmc 
+
+gulmc is the component which calculates ground up loss. It takes the modelpy output as standard input and based on the sampling parameters specified, performs Monte Carlo sampling and numerical integration. The output is a stream of ground up loss samples in Oasis kernel format with random samples identified by positive sample indexes (sidx 1 and greater), and special meaning samples assigned to negative sample indexes.
```
Ha-Ree: This describes `gulpy`, not `gulmc`. `gulmc` doesn't take a CDF stream — it reads the event stream and samples from footprint/vulnerability directly, which is the whole point of the full Monte-Carlo engine. Contradicts `CoreComponents.md:17,96-99`.

docs/source/reference/kernel/CoreComponents.md
```diff
@@ -0,0 +1,225 @@
...
+| -5 | maximum loss |
+
+**Allocation rule** (`-a`) — how item losses are adjusted when a coverage is hit by
+multiple perils (total loss to a coverage cannot exceed its TIV):
+
+| `-a` | description |
+|------|-------------|
+| 0 | pass losses through unadjusted (single-peril models) |
+| 1 | sum losses, cap to TIV, back-allocate to items in proportion to unadjusted losses |
+| 2 | keep the maximum sub-peril loss, others zero; back-allocate equally on ties |
```
Ha-Ree, line 153: Allocation rule `3` is missing. Both engines accept it — `gul/manager.py:228` and `gulmc/manager.py:110` validate against `[0, 1, 2, 3]` — and it's distinct from rule 1: rules 1-2 cap via `split_tiv_classic`, rule 3 uses `split_tiv_multiplicative`, i.e. total loss is `tiv * (1 - (1-A)(1-B)...)` over sub-peril damage ratios, back-allocated proportionally (`gul/core.py:130-150`).

```suggestion
| 0 | pass losses through unadjusted (single-peril models) |
| 1 | sum losses, cap to TIV, back-allocate to items in proportion to unadjusted losses |
| 2 | keep the maximum sub-peril loss, others zero; back-allocate equally on ties |
| 3 | combine sub-peril losses multiplicatively, `tiv * (1 - (1-A)(1-B)...)`, back-allocated in proportion to losses |
```

docs/source/reference/outputs/results.rst
```diff
@@ -0,0 +1,1433 @@
...
+|
+
+----
+
+.. _ORD_output_components_results:
+
+ORD Output Components
+---------------------
+
+As well as the set of legacy outputs described in OutputComponents.md, ktools also supports Open Results Data "ORD" output
```
SkylordA: OutputComponents is in this file at the top no?

docs/source/reference/outputs/results.rst
```diff
@@ -0,0 +1,1433 @@
+Results
+=======
+
+.. note::
```
SkylordA, line 4: This doesn't make it clear enough and is easy to miss that *ORD Outputs* is what should be used, especially since just *Output Components* is the first header you see.

docs/source/reference/kernel/RandomNumbers.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Methodology
+
+The method of assigning random numbers in gulmc uses an random number index (ridx), an integer which is used as a position reference into a list of random numbers.  S random numbers corresponding to the runtime number of samples are drawn from the list starting at the ridx position.
+
+In pytools the random-number generator is selected with `--random-generator`: `0`
+Mersenne-Twister, `1` Latin Hypercube, `2` Latin Hypercube on Philox4x32-7 (the default).
+The seeding and correlation methodology described here still applies; historically ktools
+offered three ways to source random numbers (below), now selected via `--random-generator`
+rather than the legacy `-R`/`-r`/`-s` flags.
```
SkylordA: examples below use these flags

docs/source/explanation/financial-module.rst
```diff
@@ -0,0 +1,662 @@
...
+works is to be fed data in order to execute calculations, so for the insurance calculations it needs to know the
+structure, parameters and calculation rules to be used. This data must be provided in the files used by the Oasis
+Financial Module:
+
+* **fm_programme**: defines how coverages are grouped into accounts and programmes
+* **fm_profile**: defines the layers and terms
+* **fm_policytc**: defines the relationship of the contract layers
+* **fm_xref**: specifies the summary level of the output losses
+
+This section explains the design of the Financial Module which has been implemented in the **fmcalc** component.
```
SkylordA: fmpy?

docs/source/reference/index.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Python API
+
+The API reference is generated directly from the source docstrings with
+[`sphinx-autoapi`](https://sphinx-autoapi.readthedocs.io/) (AST-based, so it
+never imports the package).
+
+```{note}
+The API reference is **scoped per-subsystem** — currently the Financial Module
```
SkylordA: stale note

docs/source/explanation/coverage-dependency.md
SkylordA, line 1: claude says things in this file aren't used anywhere?
"""
Documents an entire "coverage dependency" mechanism — `coverage_dependency_settings`, `source_coverage_type`, `dependent_coverage_type`, `conditional_vulnerability.csv/.bin`, functions `build_coverage_dependency_forest`/`get_conditional_vulns`/`validate_coverage_dependency` — none of which exist anywhere in `oasislmf/pytools/gulmc/` or the rest of the repo (verified: zero grep hits repo-wide). This looks like documentation for a planned/abandoned feature that was never built.
"""

docs/source/how-to/mdk-builtin-lookup.rst
SkylordA, line 1: similar issues, claude says things don't exist

"""
References classes `OasisPerilLookup`, `OasisVulnerabilityLookup`, `OasisLookup` under `oasismf.model_preparation.lookup` (note also the typo "oasismf") — confirmed none of these three classes exist anywhere in the repo (`grep -rn "class OasisPerilLookup|class OasisVulnerabilityLookup|class OasisLookup\b"` → 0 hits). Also references CLI command `oasislmf model generate-peril-areas-rtree-file-index` (lines 15, 77) — not in `oasislmf/cli/model.py`'s `sub_commands` dict, doesn't exist. One nuance: `OasisLookupFactory` itself (also referenced in this file) does still resolve, via a deprecated import shim in `oasislmf/__init__.py:41-79` that aliases `oasislmf.model_preparation.*` → `oasislmf.preparation.*` — so that one specific class isn't part of the bug, only `OasisPerilLookup/OasisVulnerabilityLookup/OasisLookup` and the CLI command are.
"""

SkylordA (changes requested):

DOCS_STRATEGY.md
vinulw, line 1: Think that the use of this document and `MIGRATION_PLAN.md` have significant overlap as both of them include Claude planning details for the migration as well as documentation info. I think it would be better if all the migration details were moved to `MIGRATION_PLAN.md` and this document was entirely in place as a reference for future documentation. This would mean prioritising things like section 4, 5, 6 and maybe 10 in this doc.

docs/source/explanation/coverage-dependency.md
sstruzik: Yes, it's actually a documentation that correspond to PR #2063, that I used to test. I've removed it from this PR.

docs/source/explanation/pre-analysis-adjustments.rst
```diff
@@ -0,0 +1,58 @@
...
+Oasis currently offers two toy models that demonstrate the possible options for pre-analysis adjustment: 
+:doc:`Disaggregation </explanation/disaggregation>` via `PiWind Postcode 
+<https://github.com/OasisLMF/OasisModels/tree/main/PiWindPostcode>`_, and Geocoding via 
+`PiWind Pre Analysis <https://github.com/OasisLMF/OasisModels/tree/main/PiWindPreAnalysis>`_.
+
+For more information on these model:
+
+* :doc:`Disaggregation </explanation/disaggregation>`
+
+* Geocoding
```
vinulw: missing link to geocoding doc

docs/source/explanation/keys-service.rst
vinulw, line 1: some of these code blocks do not have their types correctly defined so the syntax highlighting is missin

docs/source/index.rst
vinulw, line 1: not sure why but the sidebar has everything in drop down menus except the How-To Guides. Think that this should also be in a dropdown for consistency

docs/source/index.rst
```diff
@@ -1,39 +1,84 @@
...
+=====================================================
+
+This repository provides a Python toolkit (the Model Development Kit) for
+building, running and testing Oasis LMF models end-to-end. It includes an API
+client, a class framework for working with models and exposures, a keys/lookup
+factory, and a command line interface (run ``oasislmf --help``).
+
+The documentation is organised using the `Diátaxis <https://diataxis.fr/>`_
+framework — **Tutorials**, **How-to guides**, **Reference** and **Explanation** —
+so each page has one clear job. See ``DOCS_STRATEGY.md`` in the repository root
```
vinulw: should have a link for this

oasislmf/pytools/common/event_stream.py
vinulw, line 1: file not included in the reference section

oasislmf/pytools/gulmc/aggregate.py
vinulw, line 1: formatting looks weird for these returns

docs/source/reference/kernel/Workflows.md
sstruzik: Fixed in b00c62e7a. `summarypy` no longer appears upstream of a pipe anywhere in this file — each summary set is written to a file and read back with `-i`:

```bash
evepy 1 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 summary_p1.bin
evepy 2 2 | gulmc -S100 -a1 | fmpy -a2 | summarypy -t il -2 summary_p2.bin
eltpy -i summary_p1.bin -s elt_p1.csv
eltpy -i summary_p2.bin -s elt_p2.csv
```

Thanks for catching this — a `-2 -` would have silently created a file called `-` and left the downstream steps reading nothing, which is exactly the kind of example that looks right and teaches the wrong thing.

docs/source/reference/kernel/Workflows.md
sstruzik: Fixed in b00c62e7a, including the `tee` variant you flagged as needing a bigger rethink. Every pipeline in the file now ends at a summary binary, and the prose says why rather than leaving it implicit (lines 69-70):

> `summarypy` writes summary binaries (it has no stdout stage), so each summary is written to a file and read back by `eltpy`

The two-perspective example keeps the `tee` for the GUL/IL split, but both branches now terminate in files:

```bash
evepy 1 2 | gulmc -S100 -a1 | tee >(summarypy -t gul -2 gul_summary_p1.bin) \
          | fmpy -a2 | summarypy -t il -2 il_summary_p1.bin
eltpy -i gul_summary_p1.bin -s gul_elt_p1.csv
```

docs/source/reference/kernel/OutputComponents.md
sstruzik: Fixed in b00c62e7a, using the `-i` form you suggested:

```bash
summarypy -t gul -i gul.bin -1 summary_gul.bin
```

Same change applied to the `pltpy` case further down (now `summarypy -t il -i il.bin -1 summary_il.bin`).

docs/source/reference/kernel/OutputComponents.md
sstruzik: Fixed in b00c62e7a — `-M` for MPLT, exactly as suggested:

```bash
katpy -M -f parquet -i mplt_P1.parquet mplt_P2.parquet -o MPLT.parquet
```

docs/source/reference/kernel/StreamConversionComponents.md
sstruzik: Fixed in b00c62e7a. To answer the question you raised — there is no replacement sub-command, because the raw summary stream has no CSV converter; ORD output comes from `eltpy`/`pltpy` instead. Both places now say so.

The mapping table row (line 12):

> | `summarycalctocsv` | *no direct equivalent — use `eltpy`/`pltpy` for ORD output from the summary stream* |

and the examples are replaced by an explicit note (line 62):

> There is no `bintocsv summarycalc` sub-command — the raw `summarypy` summary stream has no direct CSV converter. Produce ORD tables from it with `eltpy`/`pltpy` (see `OutputComponents`) …

docs/source/reference/kernel/StreamConversionComponents.md
sstruzik: Fixed in b00c62e7a, taking your suggestion verbatim including the trailing note:

```bash
csvtobin gul -i gul.csv -o gul.bin -S 100        # -S/--max_sample_index is required
```

docs/source/reference/kernel/ValidationComponents.md
sstruzik: Fixed in b00c62e7a, and I took the `-N` point too since you were right that the flag deserves naming on a page about validation:

```bash
csvtobin footprint -i footprint.csv -o footprint.bin -x footprint.idx -m 50        # validates by default
csvtobin footprint -N -i footprint.csv -o footprint.bin -x footprint.idx -m 50     # -N skips validation
```

docs/source/how-to/ground-up-losses.md
sstruzik: All three fixed in b00c62e7a — the line is now your suggestion exactly:

```bash
evepy 1 1 | gulmc -S 100 -a 0 --random-generator 2 -o gulmc.bin
```

`evepy` rather than the ktools `eve`, no `modelpy` CDFs piped into `gulmc`, and no `-i -`. The surrounding prose now states that `gulmc` reads an event stream directly, so the distinction from the `gulpy` pipeline is explicit rather than something a reader has to infer.

docs/source/how-to/ground-up-losses.md
sstruzik: Fixed in b00c62e7a — both sections reframed as opt-out, as you suggested. The heading is now "The ground-up engine (gulmc is the default)" and the body reads:

> `gulmc`, the full Monte-Carlo Python engine, is the **default** — no flag is needed […] To opt out and fall back to the CDF-based `gulpy` engine, pass `--gulmc False` (config `"gulmc": false`).

`--do-disaggregation` likewise now shows `--do-disaggregation False` for turning it off. This also removes the disagreement with `CoreComponents.md:18` that you pointed at.

docs/source/reference/kernel/CoreComponents.md
sstruzik: Fixed in b00c62e7a — rule `3` added with your multiplicative description:

> | 3 | combine sub-peril losses multiplicatively, `tiv * (1 - (1-A)(1-B)…)`, back-allocated in proportion to losses |

Rules 1 and 2 were reworded at the same time so the distinction you drew (`split_tiv_classic` capping versus `split_tiv_multiplicative`) is visible in the table rather than only in the code.

docs/source/reference/kernel/Specification.md
sstruzik: Partly fixed in b00c62e7a — **leaving this open**, because the prose was corrected but the table above it still contradicts it.

The description now says what you described (line 186):

> gulmc is the full Monte Carlo component which calculates ground up loss. It takes the evepy event stream as standard input and reads the model's footprint and vulnerability data directly (rather than a pre-computed cdf stream) […] (The alternative `gulpy` engine instead consumes the `modelpy` cdf stream described above.)

But the stream table at line 61 still declares a cdf input:

```
| gulmc      | 0/1 cdf        | 2/1 loss       | -i -a{} |
```

So the page now says one thing in prose and the opposite in the table, which is arguably worse than the original. The `-i` in that row is misleading too, given `gulmc` reads stdin when `-i` is omitted.

Before I change it: the other rows use numbered kernel stream types, and the evepy event stream isn't one of those — so what would you prefer in the input column? Something like `event ids (stdin)`, or a dash with a footnote pointing at the prose? Happy to go with your call rather than invent a convention for that table.

sstruzik: Fixed in ea23ccc6e — and to answer the header question directly, since it changed the shape of the fix.

**The event stream has no header.** `evepy`'s `stream_events()` writes only the ids:

```python
for e in events:
    stream_out.write(np.int32(e).tobytes())
```

and `modelpy` consumes them as bare 4-byte ints (`getmodel/manager.py:915` → `sys.stdin.buffer`, reading into a 4-byte `event_id_mv`). So there is nothing to assign a stream_id to.

**I did not renumber it to 0**, because `common/event_stream.py:19` defines `CDF_STREAM_ID = 0` — 0 is the cdf stream in code, and documenting it as the event stream would put the page back in conflict with the source. A headerless stream has no id, so it is shown as `event_id list (no header)`, and the paragraph introducing stream_ids now says why:

> The one exception is the event stream that starts a pipeline: `evepy` writes a bare sequence of 4-byte `event_id` integers with **no header**, so it carries no stream_id.

**The table now reads:**

| Component | Standard input | Standard output | Stream option parameters |
|:--|:--|:--|:--|
| evepy | none | event_id list (no header) | -o (default `-`, i.e. stdout) |
| modelpy | event_id list | 0/1 cdf | none |
| gulpy | 0/1 cdf | 2/1 loss | -a{} |
| gulmc | event_id list | 2/1 loss | -a{} (reads model data directly, not a cdf stream) |
| fmpy | 2/1 loss | 2/1 loss | none |
| summarypy | 2/1 loss | 3/1 summary | -i input from gulmc, -f input from fmpy |
| outputcalc | 3/1 summary | none | none |

Three changes beyond the row you flagged, all from the same root cause — the table had no way to name the event stream, so the rows either mislabelled it or omitted it:

- **`modelpy` said `none`** for input, but it reads the event stream on stdin. Same class of error as the gulmc row.
- **`evepy` had no row at all**, even though line 25 of this page calls it "the stream-initiating component".
- **Added a `gulpy` row.** The cdf stream had a producer in the table but no consumer, which is part of what made the gulmc row look plausible. `gul/io.py:73` rejects any stream whose source type isn't `CDF_STREAM_ID`, so that is where the cdf stream goes.

Also dropped `-i` from gulmc's option column, per your earlier point: its own help calls it "filename of input stream (list of events from `eve`)" — a file path, not a stream option — and gulmc reads stdin when it's omitted.

Build is 0 warnings and I checked the rendered table rather than just the source.

One adjacent thing I left alone rather than widen the scope: the higher-byte tables at lines 41-44 and 51-52 label stream 1 as "gulmc (deprecated)". Stream 1 is `GUL_STREAM_ID`, the legacy gul stream that's deprecated — gulmc itself isn't. Worth a separate tidy if you agree.

docs/source/reference/index.md
sstruzik: Fixed in 11e13827c — you were right, and it was staler than it looked: the note claimed two subsystems while `conf.py` had grown to nine, and this change takes it to ten.

It now lists what is actually generated rather than a count that drifts:

> Currently included: the Financial Module (`oasislmf.pytools.fm`), the ground-up Monte-Carlo engine (`oasislmf.pytools.gulmc`), the shared stream/data layer (`oasislmf.pytools.common`), the output modules (`elt`, `lec`, `plt`, `aal`, `summary`, `pla`) and the keys/lookup framework (`oasislmf.lookup`).

docs/source/explanation/pre-analysis-adjustments.rst
sstruzik: Fixed in da457db88 — now `:doc:`Geocoding </explanation/geocoding>``, matching the disaggregation bullet directly above it, which was already linked. Verified in the rendered HTML (`href="geocoding.html"`), not just the source.

docs/source/index.rst
sstruzik: Fixed in da457db88. It was an inline literal (``` ``DOCS_STRATEGY.md`` ```) rather than a link, so there was nothing to click. Now points at the file on GitHub, matching how `explanation/index.md` already refers to the same document.

docs/source/index.rst
sstruzik: Fixed in da457db88, and there was a concrete cause worth recording.

Tutorials, Reference and Explanation each point their toctree at a single `*/index` page, which Furo renders as an expandable entry. How-to guides listed its eight pages directly in `index.rst`, and **no `how-to/index` existed** — so there was nothing for Furo to expand. Added one in the same style as `tutorials/index.md` and pointed the toctree at it. The sidebar now reads `Tutorials / How-to guides / Reference / Explanation` with all four expandable, which I checked in the built HTML.

One wrinkle I left alone: three of the eight pages (`installation`, `building-and-running-models`, `logging-configuration`) live at the top level rather than under `how-to/`, so the new index references them as `../installation`. Works, but moving them would be tidier — happy to do that separately if you want the tree to match the section.

oasislmf/pytools/common/event_stream.py
sstruzik: Fixed in 11e13827c. `autoapi_dirs` covered nine paths and none was `oasislmf/pytools/common`, so the shared stream/data layer — which defines the stream ids, headers and special sample indexes — had no reference pages. Added, and wired `api/oasislmf/pytools/common/index` into the reference toctree. `event_stream` now renders at `reference/api/oasislmf/pytools/common/event_stream/`.

Worth flagging what that exposed: adding it took the build from 0 to **17 warnings**, because `pytools/common` docstrings had never been through the RST sweep the other subsystems had. Fixed in the same commit, docstring-only:

- `data.py` — bare `*_header` / `*_dtype` / `*_fmt` read as RST emphasis start-strings
- `hashmap.py` — three API/perf sketches were indented under a single `:` and needed `::` literal blocks; a nested bullet list needed blank lines either side; `numpy.str_` has a trailing underscore, which RST read as a reference
- `input_files.py` — two Google-style *named* returns (`intervals (type): …`) made napoleon treat the **name** as the return type and emit a Python xref, which matched two attributes elsewhere in the package ("more than one target found")

That last one is the same defect you flagged separately on `gulmc/aggregate.py`, so there is now a proven pattern for it.

Back to 0 warnings, and the three touched modules are AST-identical to HEAD once docstrings are blanked, so no behaviour changed.

docs/source/explanation/keys-service.rst
sstruzik: Fixed in da457db88 — five blocks were mislabelled, and fixing the labels turned up two real problems.

Four Python snippets were marked `text`, and one shell command (`oasislmf model run --config …`) was marked `python`. Relabelling then made Sphinx warn, because:

- the snippets used `class <model_id>Lookup(Lookup):`, which is not valid Python, so Pygments failed on the `<` and highlighting silently fell back. Replaced the placeholder with `MyModelLookup`, so the snippets are valid and copy-pasteable.
- **two of them contained a genuine bug**: `pd.concat([loc_missing, locations[~missing_height_with_storeys])` never closes the list. Fixed. Every `python` block in the file now passes `ast.parse`.

Two blocks are deliberately left as `text`: they are pseudo-JSON with `<placeholder>` values, and lexing them as `json` produces 188 and 54 Pygments error tokens, which is worse than no highlighting. Happy to revisit if you would rather see them highlighted.

docs/source/reference/kernel/RandomNumbers.md
sstruzik: Fixed in d9d69fbca — and thanks, because this was worse than "the examples use the old flags". Checking `gulmc/cli.py`, **none of those flags exist**: there is no `-R`, no `-r`, no `-s`, no random-number-file input and no seed option at all. Add `-S00` (a typo for `-S100`) and `-i -` (which reaches `open('-','rb')` → `FileNotFoundError`), and every example on the page failed rather than merely looking dated.

It went deeper than the flags: `ridx` appears **nowhere** in pytools, so the whole methodology section — a shared buffer of random numbers addressed by an index computed from primes — was ktools-only too.

Rewritten from `pytools/gul/random.py`:

- seeds are derived per `(group_id, event_id)`, so a run is repeatable with no seed flag and no buffer size to choose — that is the old "method 3", and it is the only one that survived;
- **hazard sampling has its own constants** keyed off `hazard_group_id`, which the page never mentioned even though `gulmc` samples hazard as well as damage;
- `--random-generator {0,1,2}` selects the generator (numpy MT19937 / Latin Hypercube / LH on Philox4x32-7, default `2`);
- `-d {1,2}` writes the hazard/damage random numbers instead of losses, which is how you now inspect what a run drew — the closest thing to examining the old buffer.

The three legacy modes are kept as a short, clearly-marked historical note so anyone arriving from the ktools docs isn't stranded.

Verified in the rendered HTML that no invalid flag survives, rather than just in the source.

docs/source/explanation/financial-module.rst
sstruzik: Fixed in d9d69fbca. "fmpy?" was right, and there were more than I first counted — **ten** lines still said `fmcalc` or `summarycalc`, not two.

I checked the flags against `fm/cli.py` rather than assuming they carried over, which was worth doing:

```
$ fmpy -p direct < guls.bin | fmpy -p ri1 -n > ri1_net.bin
```

`-p/--static-path` and `-n/--net-loss` both exist. `-n` is actually *richer* than the ktools flag: empty writes net instead of gross to `-o`, while `-n <path>` writes net to that path and gross to `-o`. The allocation examples become `fmpy -a 0 / -a 1 / -a 2` (`-a/--allocation-rule`, value as a separate token).

One `fmcalc` mention is deliberately left, at line 188 — it is explicitly labelled "(``fmcalc`` in the legacy ktools kernel)", so it is a correct historical reference rather than a stale one.

docs/source/how-to/mdk-builtin-lookup.rst
sstruzik: You were right, and it turned out to be the whole page rather than a few names — so I removed it in 6e64b2366 instead of repointing it.

What I found when I went through all 348 lines:

- it documents three classes — `OasisPerilLookup`, `OasisVulnerabilityLookup`, `OasisLookup` — under `oasislmf.model_preparation.lookup`, reached via `OasisLookupFactory`. None exist, **and there is nothing to repoint them to**: the built-in lookup is now one config-driven class, `oasislmf.lookup.builtin.Lookup`, whose peril and vulnerability behaviour comes from method factories rather than separate classes.
- it includes IPython transcripts of those classes (`Out[7]: <…OasisPerilLookup at 0x110fdb978>`) and a class diagram of how the three relate.
- its entire workflow is pre-generating an Rtree index file with `oasislmf model generate-peril-areas-rtree-file-index`, including a `-f` index path and a `"rtree_index"` config subsection. That command is gone **and nothing replaced it**: `build_rtree` reads a user-supplied geometry file with geopandas and joins locations at runtime (`lookup/builtin.py:640`, `gpd.sjoin`), so there is no index artefact to generate or configure.

`explanation/keys-service.rst` already documents the current design — the lookup config, `builtin_lookup_type`, the step methods, adding your own — so a rewrite would have largely duplicated it. I added a pointer at the top of the how-to index so the subject is still findable, and dropped the class diagram image, which nothing else referenced.

Two related things rather than fold them in silently:

- the pre-migration originals (`docs/mdk-builtin-lookup.rst` and its `.jpg`) still sit at the `docs/` root. They are outside `source/`, so nothing builds them, and they are one of a dozen legacy files there — raised as #2113 rather than removing one file from that pile here.
- this also makes the file lists in `DOCS_STRATEGY.md:58` and `MIGRATION_PLAN.md:325` stale, which folds into @vinulw's open thread about those two documents.

oasislmf/pytools/gulmc/aggregate.py
sstruzik: Fixed in 43ce05d17. Two shapes napoleon does not read as a Returns section:

- one `Returns:` listing three named values, which renders as a type plus continuation text instead of three items;
- an inline `Returns: (np.ndarray[oasis_float]) …`, which is not parsed as a section **at all** — napoleon needs an indented block on the next line.

Both rewritten description-first, the three-value one as a bullet list.

Worth recording because it caught me out mid-fix: in a Google-style `Returns:` block napoleon treats everything before the **first colon on the first line** as the return type. My initial lead-in ended in a colon, so the whole sentence rendered under "Return type" and only the bullets landed under "Returns". Dropping the colon fixes it — and it was only visible in the rendered HTML, not the source.

Same defect as the two named returns in `pytools/common/input_files.py` (11e13827c), so there is a consistent pattern now.

Docstring-only: AST-identical to HEAD once docstrings are blanked, so no behaviour changed.

DOCS_STRATEGY.md
sstruzik: Done in 2116a7653, though not quite as proposed — one fact reframes the choice.

**Neither file is on `main`.** Both were added by this branch (7b9643178), so this isn't two existing documents to rebalance; it's 807 lines of planning material the PR would *introduce* (`DOCS_STRATEGY.md` 306, `MIGRATION_PLAN.md` 501). That opens a third option, which is what we went with: publish the durable half, drop the transient half.

The two have different lifespans:

- **`MIGRATION_PLAN.md` is transient nearly end to end** — the four content actions, phasing, rough effort, immediate next actions, a progress log. It describes work that finishes when this merges, and a 501-line progress log at the repo root forever makes little sense.
- **`DOCS_STRATEGY.md` is a mix.** §4 architecture, §5 content model, §6 tooling and §9 gotchas are durable — exactly the "reference for future documentation" you asked for. §1-3 (as-found state, problems), §7-8 (phased plan, what this branch changes) and §10 (open questions, all since answered) are only true until this lands.

So the durable half is now a published page — {doc}`explanation/documentation-strategy` — on the grounds that a contributor guide belongs *in* the docs rather than loose at the repo root, and the transient half is gone. The reasoning stays in git history at 7b9643178 for anyone who wants it.

**One thing worth calling out:** rewriting it as settled practice meant checking the claims, and the old §6 tooling table listed several decisions that were never implemented. Publishing that as guidance would have recreated the drift the page argues against. They're now a separate "agreed but not yet implemented" list:

- no `linkcheck` job exists in CI — dead external links build perfectly cleanly today
- `interrogate` (docstring coverage) is not configured
- mermaid is not configured
- there is no render check, which matters given gotcha #2 below

The gotchas section also picked up the traps found since it was written: the `{external+...}` role that drops its link text rather than degrading, the napoleon `Returns:` colon rule, and the docstring-RST rules.

Nine reference sites repointed, including `reference/kernel/index.md`, which cited "`MIGRATION_PLAN.md` §3.3". This also retires the stale file lists at `DOCS_STRATEGY.md:58` / `MIGRATION_PLAN.md:325` that surfaced when the mdk-builtin-lookup page was removed.

Build stays at 0 warnings; the new page renders and is linked from both the root index and the explanation index. Happy to reinstate `MIGRATION_PLAN.md` if you'd rather keep the record in-tree.

docs/source/reference/outputs/results.rst
sstruzik: Fixed in af677a7aa — you'd spotted a genuine collision. The first section was literally titled **"Output Components"**, while `reference/kernel/OutputComponents.md` is a *different* page with the same name in the same reference section: one documented the tools, this one the formats.

Renamed to **"Legacy (pre-ORD) output formats"**, which removes the ambiguity and says what it actually is. The tools page keeps the "Output Components" name, and this page now links to it rather than restating command lines.

docs/source/reference/outputs/results.rst
sstruzik: Fixed in af677a7aa. You were right, and the page's own wording made it worse: a reader met ~550 lines of deprecated formats first, and the ORD section then introduced itself as an afterthought — *"As well as the set of legacy outputs described above, Oasis also supports Open Results Data"*.

The page now reads:

```
Results
  On this page
  ORD outputs                       <- opens by stating it is the current standard
    Sample Event Loss Table (SELT)
    Moment and Quantile Event Loss Tables (MELT, QELT)
    Period Loss Tables (SPLT, MPLT, QPLT)
    Exceedance Probability Tables (EPT, PSEPT)
    Average Loss Tables (ALT, PALT)
  Output File Naming Conventions    <- ORD table first, legacy second
  Legacy (pre-ORD) output formats   <- moved to the end, behind an explanatory note
```

Two further problems surfaced while restructuring, both worth knowing:

**Even the ORD half was organised by retired binary** — "summarycalctocsv (ORD)", "eltcalc (ORD)", "ordleccalc (ORD)". It is now organised by the table a reader is actually looking for, with a lead-in mapping each `ord_output` flag to its table and to the pytools component that emits it (`elt_moment` → MELT → `eltpy`, and so on).

**Every command example was a ktools invocation** with ktools flags (`-M {filename.csv}`, `$ ordleccalc …`). I removed them rather than translating: the components are documented in pytools form in `reference/kernel/OutputComponents.md`, and duplicating command lines here is what let them rot. The page links there instead.

**What I deliberately kept:** all 15 field-specification tables. I had assumed they duplicated the ORD reference generated in `ODS_OpenResultsData` and could go — checking the built output showed that reference covers field names, tables and descriptions, but **not** the types, byte widths or examples. Those are the concrete encoding the Oasis toolchain writes, so they belong here. The split is now explicit: the ORD standard owns the definitions (linked out), this page owns what Oasis writes.

1434 → 1157 lines. Verified in the rendered HTML — section order as above, zero legacy binary invocations left, no duplicate or dangling anchors, build 0 warnings.

One dependency to note: the "ORD standard" link points at `oasislmf.github.io/ord/...`, which goes live when ODS_OpenResultsData#11 lands. Until then it 404s in a standalone build; in the assembled site the orchestrator rewrites it page-relative, so it resolves there.

docs/source/explanation/currency_conversion.rst
```diff
@@ -0,0 +1,119 @@
...
+
+.. _dict_based_conversion:
+
+File Based Currency Conversion
+******************************
+
+DictBasedCurrencyRates is a solution where all the rate are provided via files and stored internally in the package as a dictionary.
+
+Oasis LMF supports csv file (compressed or not) or a parquet file where they will be read as DataFrame.
+The file is referenced using a JSON configuration file with the ``--currency_conversion_json`` flag :
```
SkylordA: minor typo, should be `--currency-conversion-json` with dashes not underscore

docs/source/how-to/logging-configuration.rst
SkylordA: Claude claims: `OASIS_PACKAGE_LOG_LEVEL=20` actually raises `ValueError: Invalid log level: '20'...` (verified by running it) — env vars are strings, and the numeric branch in `log_config.py` only accepts a Python `int`.

docs/source/reference/kernel/ReferenceModelOverview.md
```diff
@@ -0,0 +1,77 @@
...
+
+* **[4.1 Core components](CoreComponents.md)**
+* **[4.2 Output components](OutputComponents.md)**
+* **[4.3 Data conversion components](DataConversionComponents.md)**
+* **[4.4 Stream conversion components](StreamConversionComponents.md)**
+* **[4.5 Validation components](ValidationComponents.md)**
+
+The set of **[core components](CoreComponents.md)** provided in this release is as follows;
+* **evepy** is the event distributing utility. Based on the number of events in the input and the number of processes specified as a parameter, evepy outputs subsets of the events as a stream. The output streams into modelpy.
+* **modelpy** generates a stream of effective damageability cdfs for the input stream of events. It generates cdfs from the model files footprint.bin and vulnerability.bin, and the user's exposures file which is called items.bin. modelpy streams into gulmc or can be output to a binary file.
```
SkylordA: `modelpy` does not stream into `gulmc` — `gulmc` reads the event stream directly. Contradicts `Introduction.md`/`Specification.md`/`CoreComponents.md`, already fixed elsewhere in this PR.

docs/source/reference/kernel/ReferenceModelOverview.md
```diff
@@ -0,0 +1,77 @@
...
+The **[data conversion components](DataConversionComponents.md)** section covers the formats of all of the required data files and explains how to convert data in csv format into binary format, and vice versa.
+
+The **[stream conversion components](StreamConversionComponents.md)** section explains how to convert the binary data stream output to csv, plus how to convert gulmc data in csv format into binary format. These components are useful when working with individual components at a more detailed level.
+
+The **[validation components](ValidationComponents.md)** section explains how to use the validation components to check the validity of the static and input files in csv format, before they are converted to binary format. There are both validation checks on individual files and cross checks for consistency across files.
+
+The version of the installed components can be found by using the command line parameter -v. For example;
+
+```
+$ gulmc -v
```
SkylordA: `gulmc` has no `-v` flag (only `-V/--version`), and doesn't print `gulmc : version: X.X.X`.

docs/source/reference/kernel/ReferenceModelOverview.md
```diff
@@ -0,0 +1,77 @@
...
+gulmc : version: 3.0.7
+```
+
+Component usage guidance is available using the parameter -h
+
+```
+$ fmpy -h
```
SkylordA, line 64: This `fmpy -h` flag list is fabricated — no `-M`/`-O`/`-d`, no version flag exists at all. Real flags: `-a`,`-n`,`-p`,`-i`,`-o`,`-l`,`--sort-output`,`--storage-method`,`--create-financial-structure-files`,`-v`(=logging-level),`-S`.

docs/README.md
```diff
@@ -1,38 +1,105 @@
...
-Automation
-----------
-Upon committing and pushing the docs to the GitHub repository, the documentation will be:
- - built for all PRs targeting `main`, with the HTML documentation being stored as a downloadable artifact (a `github-pages.zip` file);
- - built and deployed to GitHub pages (i.e., stored in the `gh-pages` branch of the repository) for all push commits on the `main` branch.
+The Python API reference is generated from source docstrings by `sphinx-autoapi`,
+which is **AST-based** — it parses the source and never imports the package, so
+the heavy runtime dependencies (numba, etc.) are not imported at build time.
 
+`autoapi_dirs` in `source/conf.py` controls the scope. It is currently **scoped
```
SkylordA: Autoapi covers 10 subsystems now (`conf.py`'s `autoapi_dirs`), not just the Financial Module — same stale note already fixed in `reference/index.md`.

docs/source/how-to/api-client.rst
```diff
@@ -0,0 +1,393 @@
...
+
+.. list-table::
+   :widths: 35 65
+   :header-rows: 1
+
+   * - Option
+     - Description
+   * - ``--server-url URL``
+     - Base URL of the platform (default: ``http://localhost:8000``)
+   * - ``--server-version {v1,v2}``
```
SkylordA: `{v1,v2}` implies enforced choices, but `server_version` (`computation/run/platform.py:46`) has no choices — any string is accepted.

docs/source/how-to/model-development-kit.rst
```diff
@@ -0,0 +1,114 @@
...
+You can also install from a specific branch ``<branch name>`` using:
+
+.. code-block:: sh
+
+    pip install [-v] git+{https,ssh}://git@github.com/OasisLMF/OasisLMF.git@<branch name>#egg=oasislmf
+
+|
+
+.. note::
+    Starting from 1st January 2019, Pandas will no longer be supporting Python 2. As Pandas is a key dependency of the MDK
```
SkylordA, line 106: Says release 1.3.4 (2019) requires Python 3.10 minimum — impossible, Python 3.10 didn't exist until Oct 2021.

docs/source/tutorials/explore-model-data.md
```diff
@@ -0,0 +1,109 @@
...
+fig.tight_layout()
+```
+
+## Exposure: items and coverages
+
+**Coverages** hold the insured values (TIV); **items** link each coverage to an
+area-peril and a vulnerability function (the join between exposure and model data).
+
+```{code-cell} python
+items = pd.read_csv(DATA / "items.csv")
```
SkylordA, line 96: Printed TIV ($3.4M) is `coverages['tiv'].sum()` over all 10 coverages, not the joined `exposure`. Only coverage_ids 1-5 actually join; their real TIV is $1.45M.

docs/source/explanation/currency_conversion.rst
sstruzik: Fixed in 3aaf20e2. Confirmed at `oasislmf/cli/command.py:261`, which builds every flag as `param['name'].replace('_', '-')` — so `--currency-conversion-json` is the only spelling that has ever worked.

docs/source/how-to/logging-configuration.rst
sstruzik: Confirmed by running it, and fixed in 3aaf20e2. `_parse_level('20')` raises `ValueError: Invalid log level: '20'. Available levels: DEBUG, INFO, WARNING, ERROR`. The int branch at `log_config.py:266` needs a real `int`, which an environment variable can never be. I removed the example and stated the constraint, so the next person does not re-add it.

docs/source/reference/kernel/ReferenceModelOverview.md
sstruzik: Fixed in 3aaf20e2 — it now reads "modelpy streams into gulpy … not needed for gulmc, which reads the event stream directly". `gulmc`'s own `--file-in` help says "list of events from `eve`", and this now matches what `Introduction.md` and `CoreComponents.md` already said.

docs/source/reference/kernel/ReferenceModelOverview.md
sstruzik: Fixed in 3aaf20e2. Both halves were wrong: the flag is `-V`/`--version`, and it prints the bare `oasislmf` version (`gulmc/cli.py:27` → `action='version', version=oasis_version`), not `gulmc : version: 3.0.7`. I also noted that not every component has one — `fmpy` has no version flag at all (0 matches for "version" in `pytools/fm/cli.py`).

docs/source/reference/kernel/ReferenceModelOverview.md
sstruzik: Fixed in 3aaf20e2 — the list now comes from `pytools/fm/cli.py`. Your reading is exactly right: no `-M`, no `-O`, no `-d`, and `-v` is `--logging-level`, not version. I included the long forms and the flags the old list omitted (`--sort-output`, `--storage-method`, `--create-financial-structure-files`, `--low-memory`, `--static-path`).

docs/README.md
sstruzik: Fixed in 3aaf20e2. `autoapi_dirs` has ten entries — `pytools/fm`, `gulmc`, `common`, the output modules (`elt`, `lec`, `plt`, `aal`, `summary`, `pla`) and `lookup` — so the README now describes it as per-subsystem and names them, matching `reference/index.md`.

docs/source/how-to/api-client.rst
sstruzik: Fixed in 3aaf20e2. Confirmed: in `computation/run/platform.py` only `auth_type` declares `choices`; `server_version` is a plain string param, so any value is passed through. Now written as `--server-version VERSION` with `v1`/`v2` described as the meaningful values rather than an enforced set.

docs/source/how-to/model-development-kit.rst
sstruzik: Fixed in 3aaf20e2. You are right that it cannot be both — 1.3.4 shipped in 2019 and Python 3.10 arrived in October 2021. Split into the historical Python 2 note, which is what the release actually said, and the current minimum stated separately.

docs/source/tutorials/explore-model-data.md
sstruzik: Fixed in 3aaf20e2, and your 1.45M is the right figure — but worth recording that the obvious fix is also wrong. Summing the joined frame gives **2.9M**, because all ten items sit on those five coverages (two each), so each TIV is counted twice. The cell now sums the coverages the items reference, which prints 1,450,000, with a comment explaining both traps. Verified in the built page.

Ha-Ree:
A few unused pics:
aalcalc.jpg, banner.jpg, fmcalc.jpg, fmcalcrequireddata.jpg, gultobin.jpg, key.jpg, leccalc.jpg, msys2packages.jpg, multiplefmcalc.jpg, pltcalc.jpg, streamtocsv.jpg, summarycalc.jpg, windowspath.jpg.

sstruzik:
Thanks — all 13 removed in e361f3949.

Worth recording what they were, since "unused" undersells it. They came in with the ktools drain (59398cecf), which copied the whole image directory: of the 21 files in `_static/images/kernel/`, only 8 are referenced by the pages that were actually migrated. None of the 13 exists on `main`, and none was ever referenced — so they were over-import rather than orphans left behind by a rework.

They are also obsolete by content, which is why none of the migrated pages wanted them:

- `aalcalc`, `fmcalc`, `fmcalcrequireddata`, `leccalc`, `multiplefmcalc`, `pltcalc`, `summarycalc`, `gultobin`, `streamtocsv` — the ktools binaries the pytools kernel replaces
- `msys2packages`, `windowspath` — Windows install screenshots
- `banner`, `key` — site furniture Furo provides

The 8 kept are the conceptual diagrams that survived the rewrite: `framework`, `gulandfm`, `fm1`–`fm3`, `ktoolsrequireddata`, `outputsrequireddata`, `eltcalc`.

Checked for references across md, rst, py, html, css and yml before deleting — none. Build is clean on a fresh `-E`, ships the remaining 8, and every `` in the generated site resolves.

SkylordA (approved):
pending ci

Ha-Ree (approved):

PR #2071, opened 2026-07-23
Fixes for small issues

Ha-Ree:
1) lookup/builtin uses wrong sign for min 
2) notatrisk not propogated properly
3) error messages without f string
4) missed 2>/dev/null in bash
5) get_signature using None for falsey defaults
6) GeoTIFF written/read in wrong orders
7) Fix wrong README for tests
8) refactored shared code
9) exec(locals) has been silently changed in 3.13 and needed changing (https://github.com/python/cpython/issues/118888)


SkylordA (approved):
why is pip-tools 7.6.0 excluded?

sambles (approved):

PR #2079, opened 2026-07-29
vectorisation of cond_info

### LINKED ISSUE #2100: Vectorisation of slower functions

Ha-Ree:
Part of #2100.

`get_cond_info` resolved the condition hierarchy with three nested Python loops over
`to_dict('records')` scans, including an O(n²) insertion sort to order each location's
conditions by priority. This replaces them with frame operations: a `groupby` for the
per-`(acc_id, CondTag)` attributes, a `drop_duplicates` + dense `rank` for the per-location
priority ordering, and a merge for the cond-tag × layer coverage check.

## Behaviour

Output is unchanged — `level_conds` partitioning, `level_conds` dict insertion order,
`extra_accounts` values and `extra_accounts` row order all match the loop implementation
exactly — for any input where `locations_df` has no exactly-duplicated
`(loc_id, acc_id, CondTag)` rows.

Both orderings are load-bearing and are preserved deliberately: `get_levels` iterates
`level_conds.items()` and never reads the level number, so dict insertion order is what
assigns cond FM levels, and `extra_accounts` row order flows into
`pd.concat([accounts_df, pd.DataFrame(extra_accounts)])` and thence into FM row order.

### One deliberate behaviour change

**Duplicated location rows now count once, which can merge two cond tags into a single FM
level rather than only renumbering levels. This changes loss numbers.**

The loop built its per-location list by appending one entry per location row, duplicates
included, so every duplicate pushed higher-priority tags down a position. This version
de-duplicates first and uses a dense rank. Worked example — tags A (priority 1), B and C
(both priority 2), with location rows `(10,A), (10,A), (10,B), (11,A), (11,C)`:

| | result |
|---|---|
| before | `{1: [A], 2: [C], 3: [B]}` — three cond FM levels |
| after | `{1: [A], 2: [B, C]}` — two, with B and C sharing one |

Two conditions on the same FM level have their terms applied at one aggregation node
instead of nested. The de-duplicated answer is taken to be the intended one, because
`get_levels` already de-duplicates on exactly those three columns before use. Exactly
duplicated `(loc_id, acc_id, CondTag)` triples are a data error rather than a designed
input; nothing upstream rejects them today.

`test_duplicate_location_rows_can_merge_two_tags_into_one_level` pins this, asserting both
the new answer and the loop's old answer so the divergence stays visible if either side is
touched.

## Performance

Measured on 80k account rows (20k accounts × 3 layers × 2 cond tags) against 1M location
rows / 500k distinct locations, with identical output in all three runs:

| implementation | time |
|---|---|
| `main` (the loops this replaces) | 8.06 s |
| this PR | **2.18 s (3.7x)** |

The speedup is shape-dependent: it comes from removing the two `to_dict('records')` scans
and the insertion sort, so it scales with the location count rather than with the
`cond_tags × layers` product. `extra_accounts` construction is still per-row Python at that
same cardinality, as it was before.

One line dominated an earlier revision of this PR: `has_excl` was aggregated with a lambda,
which has no cython path in pandas and fell into `_aggregate_series_pure_python` — one
Python call per `(acc_id, layer_id)` group. Resolving the exclusion test into a boolean
column before the `groupby` and aggregating it with `'max'` took that aggregation from
5.22 s to 0.026 s in isolation, and the whole function from 7.25 s to 2.18 s.

## Tests

`tests/preparation/test_get_cond_info.py` keeps a verbatim copy of the loop implementation
as `reference_get_cond_info` and diffs the two over randomly generated account/location
pairs, so the new code is pinned against the old behaviour rather than against itself. On
top of that:

- targeted tests for the branches the FM acceptance fixtures never reach — the same-priority
  conflict raise, priority-0 handling, the `FullFilter` exclusion path and the filler rows
  for uncovered layers;
- `test_duplicate_location_rows_can_merge_two_tags_into_one_level` and
  `test_duplicate_location_rows_match_the_deduplicated_frame` for the behaviour change above.
  The random-case generator de-duplicates its locations frame deliberately, since feeding
  duplicated rows to the reference comparison would assert the old answer; these two tests
  cover that input class instead.

Local: 11 unit tests pass, and `tests/fm/test_fmpy.py` is 21/21 against the `validation/`
golden `fm_*.csv` fixtures.


sstruzik (approved):

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+            held = int(np.argmax(((loc['loc_id'] == loc_key) & (loc['priority'] == prio_val)).to_numpy()))
+            keys = [tuple(row) for row in loc[['acc_id', 'CondTag']].iloc[[arriving, held]].to_numpy().tolist()]
+            raise OasisException(f"{keys[0]} and {keys[1]} have same priority in {loc_key}")
+
+        # cond_level_start = max over a cond_tag's locations of its 1-based rank (priorities are distinct within a location)
+        loc['pos'] = loc.groupby('loc_id', observed=True)['priority'].rank(method='dense').astype(int)
+        level_map = {(r.acc_id, r.CondTag): int(r.pos)
+                     for r in loc.groupby(['acc_id', 'CondTag'], observed=True)['pos'].max().reset_index().itertuples()}
+
+        # every cond_tag key to emit results for: those defined on accounts plus those referenced by
```
SkylordA, line 401: Bug: no validation that every `acc_id` in `all_cond_keys` exists in `accounts_df`. A NaN acc_id (unresolved location→account match) raises `ValueError` here instead of the old `KeyError`; an acc_id absent from `accounts_df` entirely is silently dropped (line 288) instead of raising, producing a NaN-filled row downstream in `get_levels()`. Suggest validating before the loop, matching the existing CondTag check style (lines 214-215):
```python
missing_acc = all_cond_keys.loc[~all_cond_keys['acc_id'].isin(accounts_df['acc_id'])]
if missing_acc.shape[0]:
    raise OasisException(f'Those acc_id are present in locations but missing in the account file:\n{missing_acc}')
```

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
-                    break
-                elif cond_tag_priority == cond[PRIORITY_INDEX] and cond_key != cond[KEY_INDEX]:
-                    raise OasisException(f"{cond_key} and {cond[KEY_INDEX]} have same priority in {loc_key}")
+
+        # --- per (acc_id, CondTag) attributes taken from the first account row (0/blank priority -> 1) ---
+        tag_first = accounts_df.groupby(['acc_id', 'CondTag'], sort=False, observed=True).agg(
+            CondPriority=('CondPriority', 'first'),
+            CondPeril=('CondPeril', 'first'),
+        ).reset_index()
+        prio = pd.to_numeric(tag_first['CondPriority'], errors='coerce').fillna(1)
```
SkylordA: Bug: `pd.to_numeric(tag_first['CondPriority'], errors='coerce').fillna(1)` silently treats a garbage `CondPriority` (e.g. `'N/A'`) as priority 1; old code raised `TypeError`. If bad data should still fail loudly, use `errors='raise'` instead of coercing.

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+            raise OasisException(f"{keys[0]} and {keys[1]} have same priority in {loc_key}")
+
+        # cond_level_start = max over a cond_tag's locations of its 1-based rank (priorities are distinct within a location)
+        loc['pos'] = loc.groupby('loc_id', observed=True)['priority'].rank(method='dense').astype(int)
+        level_map = {(r.acc_id, r.CondTag): int(r.pos)
+                     for r in loc.groupby(['acc_id', 'CondTag'], observed=True)['pos'].max().reset_index().itertuples()}
+
+        # every cond_tag key to emit results for: those defined on accounts plus those referenced by
+        # locations (the latter adds default-'0' tags, which are synthetic priority-1 conds)
+        all_cond_keys = pd.concat([accounts_df[['acc_id', 'CondTag']], loc[['acc_id', 'CondTag']]]).drop_duplicates()
```
SkylordA: Simplification: re-derives unique `(acc_id, CondTag)` pairs from the full `accounts_df` when `tag_first` (built above) already has them.
```python
all_cond_keys = pd.concat([tag_first[['acc_id', 'CondTag']], loc[['acc_id', 'CondTag']]]).drop_duplicates()
```

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+        for r in all_cond_keys.itertuples():
+            cond_key = (int(r.acc_id), r.CondTag)
+            level_conds.setdefault(level_map.get((r.acc_id, r.CondTag), 1), set()).add(cond_key)
+
+        # extra 'filler' account rows for every (cond_tag, layer) the cond_tag does not already cover
+        for r in all_cond_keys.merge(account_layers, on='acc_id', how='inner').itertuples():
```
SkylordA: Efficiency: filler loop re-inflates a full `all_cond_keys × account_layers` cross join and processes it row-by-row via `itertuples()`, discarding most rows with a Python `in` check — undercuts the vectorization this PR is meant to deliver. Filter with a vectorized anti-join first (same `merge(..., indicator=True)` pattern already used at line 208-209), then build `extra_accounts` from the filtered frame.

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+            CondPeril=('CondPeril', 'first'),
+        ).reset_index()
+        prio = pd.to_numeric(tag_first['CondPriority'], errors='coerce').fillna(1)
+        tag_first['priority'] = prio.mask(prio == 0, 1)
+        peril = tag_first['CondPeril'].astype('object').fillna('')
+        tag_first['cond_peril'] = peril.mask(peril == '', 'AA1')
+        peril_map = {(r.acc_id, r.CondTag): r.cond_peril for r in tag_first.itertuples()}
+
+        # layers each cond_tag already appears in, and per (acc_id, layer_id) policy info + exclusion flag
+        layers_present = set(map(tuple, accounts_df[['acc_id', 'CondTag', 'layer_id']].drop_duplicates().to_numpy()))
```
SkylordA: Reuse: `layers_present` is a hand-rolled Python set + `in` check, duplicating the merge+indicator anti-join idiom already used at line 208-209.

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+        if collisions.any():
+            arriving = int(np.argmax(collisions))
+            loc_key, prio_val = loc['loc_id'].iloc[arriving], loc['priority'].iloc[arriving]
+            held = int(np.argmax(((loc['loc_id'] == loc_key) & (loc['priority'] == prio_val)).to_numpy()))
+            keys = [tuple(row) for row in loc[['acc_id', 'CondTag']].iloc[[arriving, held]].to_numpy().tolist()]
+            raise OasisException(f"{keys[0]} and {keys[1]} have same priority in {loc_key}")
+
+        # cond_level_start = max over a cond_tag's locations of its 1-based rank (priorities are distinct within a location)
+        loc['pos'] = loc.groupby('loc_id', observed=True)['priority'].rank(method='dense').astype(int)
+        level_map = {(r.acc_id, r.CondTag): int(r.pos)
```
SkylordA: Simplification: `peril_map`/`level_map` built via `itertuples()` dict comprehensions instead of `set_index(...).to_dict()` / `groupby(...).to_dict()`. Line 277's `int(r.pos)` is also a redundant re-cast — `pos` was already cast with `.astype(int)` on line 276.
Also for line 238 above

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
-                            'PolNumber': PolNumber,
-                            'LayerNumber': LayerNumber,
-                            'CondTag': cond_tag,
-                            'layer_id': layer_id,
-                            'CondNumber': '',
-                            'CondPeril': cond_peril,
-                        })
-            level_conds.setdefault(cond_level_start, set()).add(cond_key)
+                extra['CondNumber'] = ''
+            extra['CondPeril'] = peril_map.get((r.acc_id, r.CondTag), 'AA1')
```
SkylordA: Nit: `'AA1'` empty-peril sentinel hardcoded twice. Consider a shared constant.
Also line 237 above

oasislmf/preparation/il_inputs.py
```diff
@@ -229,83 +225,79 @@ def get_cond_info(locations_df, accounts_df):
...
+        account_layers = layer_cols.groupby(
+            ['acc_id', 'layer_id'], sort=False, observed=True)['has_excl'].max().reset_index()
+        account_layers = account_layers.merge(
+            layer_cols.drop_duplicates(['acc_id', 'layer_id'], keep='last').drop(columns='has_excl'),
+            on=['acc_id', 'layer_id'], how='left')
+
+        # --- resolve each location's cond priority, detect same-priority conflicts, rank within the location ---
+        loc = locations_df[['loc_id', 'acc_id', 'CondTag']].drop_duplicates()
+        loc = loc.merge(tag_first[['acc_id', 'CondTag', 'priority']], on=['acc_id', 'CondTag'], how='left')
+        loc['priority'] = loc['priority'].fillna(1)  # loc tags absent from accounts (e.g. default '0') -> priority 1
```
SkylordA: Nit: "missing/zero priority defaults to 1" implemented twice via two different idioms (lines 234-235 and 261).

SkylordA (changes requested):

oasislmf/preparation/il_inputs.py
Ha-Ree: Both are different- first is priority 0, second is location has no account

SkylordA (approved):
pending failing test

PR #2092, opened 2026-08-07
issue 2090 sparse event definition

### LINKED ISSUE #2090: Support for sparse event definition and hazard case files (dynamic footprints)
Currently the event definition and hazard case files are expected to include entries for every `section_id` that can be assigned by the keys server. If a location is associated with a `section_id` that does not exist in one of those files, the footprint generation and whole analysis will crash.

To work around this, we need to either identify all section ids that have no hazard or no events from all data via an awkward multi-pass process which leads to not-at-risk locations being flagged as failed and tightly couples several files, or we need to include all section ids in all datasets, potentially making them much larger than they need to be.

A cleaner solution would be for the footprint generation function to treat missing data as "not at risk", i.e. if a section id cannot be found in the event definition file, it means there are no events affecting that section and if a section id cannot be found in the hazard case file it means those sections are unaffected by the modelled perils (intensity = 0). I've drafted a solution at https://github.com/OasisLMF/OasisLMF/pull/2008 but it needs refining.

### LINKED ISSUE #2091: Unpartitioned event definition files cause model to crash
In OasisLMF main, `_get_event_flat` is invoked when event definition is a single (unpartitioned) parquet file, but the model still tries to read a section_id partition path, leading to a `FileNotFoundError`.

See https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/pytools/getmodel/footprint.py#L620-L624

Could those lines simply become:
```python
df_hazard_case = self.get_df_reader(
    hazard_case_filename,
    filters=[("areaperil_id", "in", self.areaperil_ids)]
).as_pandas()
```

It looks like the same is true of hazard_case files. There's no claim in docs or code that unpartitioned files are supported but it would be convenient if they were, for small demo and test models. 

See https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/pytools/getmodel/footprint.py#L586-L594. Same lines as affected by https://github.com/OasisLMF/OasisLMF/issues/2090.

Ha-Ree:


oasislmf/pytools/getmodel/footprint.py
```diff
@@ -555,76 +568,135 @@ def __enter__(self):
...
+            return df_sections
+
+        # not every reader engine accepts filters=None, so only pass them when there are some
+        reader_kwargs = {'filters': filters} if filters else {}
+
+        absent = []
+        df_section_list = []
+        for section in sections:
+            section_path = f'{filename}/section_id={section}'
+            if not self.storage.exists(section_path):
```
SkylordA: A section whose directory exists but is emptied entirely by `filters` is never reported absent here, unlike the flat branch (613-614) which diffs post-filter. Fixed by the collapse above; if kept separate, compute `absent` from the read result instead of `storage.exists()`.

oasislmf/pytools/getmodel/footprint.py
```diff
@@ -555,76 +568,135 @@ def __enter__(self):
...
-        df_hazard_case_list = []
-        for section in sections:
-            df_section = self.get_df_reader(
-                f'{hazard_case_filename}/section_id={int(section)}',
-                filters=[("areaperil_id", "in", self.areaperil_ids)]
-            ).as_pandas()
-            df_section['section_id'] = section
-            df_hazard_case_list.append(df_section)
-        df_hazard_case = pd.concat(df_hazard_case_list, ignore_index=True)
+        df_hazard_case = self._read_sections(
```
SkylordA: `_get_event_flat` re-lists partitions on every `get_event()` call. The fix above cuts this from one `exists()`+read per section to one listing per event, but it's still repeated per event, worth caching in `__enter__` if this path is hot.

oasislmf/pytools/getmodel/footprint.py
```diff
@@ -555,76 +568,135 @@ def __enter__(self):
...
+            filters (list): optional pyarrow filters pushed down to the reader
+
+        Returns: (pd.DataFrame) the concatenated sections with a section_id column, or an
+            empty DataFrame if none of the sections holds any data.
+        """
+        sections = sorted(int(section) for section in sections)
+        if not sections:
+            return pd.DataFrame()
+
+        if not partitioned:
```
SkylordA: The partitioned branch loops per-section with `storage.exists()`, but the reader already handles hive-partitioned directories in one call, same pattern the flat branch above uses. Suggest collapsing both branches:

```suggestion
        section_filters = (filters or []) + [("section_id", "in", sections)]
        df_sections = self.get_df_reader(filename, filters=section_filters).as_pandas()
        present = set() if df_sections.empty else set(df_sections['section_id'])
        self._report_absent_sections(filename, set(sections) - present)
        return df_sections
```
Leaves `partitioned` unused, worth dropping from the signature, docstring, and call sites as a follow-up. Also fixes the absence-detection issue below for free.

SkylordA (approved):

sstruzik (approved):

PR #2094, opened 2026-08-10
h3 vectorise

### LINKED ISSUE #2100: Vectorisation of slower functions

Ha-Ree:
Part of #2100.

`build_h3`'s lookup converted every location's coordinates to an H3 cell one row at a time and
then resolved the mapping with a left join. This keeps the per-cell Python loop — h3 has no array
API — but pays it once per *distinct* coordinate instead of once per location, and replaces the
join with a `reindex` over the distinct cells.

## Performance

200,000 locations at resolution 5, best of 5 runs, fully matched mapping:

| | all coordinates distinct | 2,000 distinct coordinates |
|---|---|---|
| `main` | 526.8 ms | 500.9 ms |
| this PR | **209.6 ms (2.5x)** | **16.2 ms (31x)** |

Both halves matter. The all-distinct case — the realistic one for a real portfolio — is already
2.5x faster purely from dropping `str_to_int`'s hex parse and the merge; the de-duplication is
what takes it further when a portfolio repeats coordinates.

`h3.api.basic_int.latlng_to_cell` is used in place of `h3.str_to_int(h3.latlng_to_cell(...))`.
I confirmed the two are bit-identical over resolutions 0–15, and H3 cell indices are at most 60
bits so the `int64` path cannot overflow.

## Behaviour

Row count, row order and the index reset all match the left join this replaces, and
`test_build_h3_matches_row_wise_reference` pins the result against a verbatim copy of the removed
row-wise loop over 5,000 rows with a 5% null-coordinate injection.

Three deliberate changes:

**A duplicate `h3_int64` in the mapping file is now an error.** A point falls in exactly one cell
at a given resolution, so a repeated cell index is a mapping file error. Previously it produced a
left-join row explosion — one key per duplicate row, silently changing the row count of the keys
file. It now raises `OasisException` at build time. The uniqueness is also load-bearing for the
new `reindex`. Documented in the `build_h3` docstring.

**Non-finite coordinates resolve to `OASIS_UNKNOWN_ID`.** The validity mask is `np.isfinite`
rather than `~np.isnan`. An infinite coordinate would otherwise survive an `isnan` test, and
`1j * inf` has a `nan` real part, so the complex key would reach `pd.factorize` as an NA and take
its `-1` sentinel — a legal numpy index that reads back from the *end* of the mapped array and
returns another location's `area_peril_id`. That would have been silent, row-order dependent, and
so dependent on multiprocessing chunk boundaries; `pd.read_csv` parses the literal `inf` in a
numeric column, so it is reachable straight from an exposure file. Infinite coordinates are now
treated as unknown, consistently with null ones, which `process_locations` reports as a
per-location `fail` status. Covered by `test_build_h3_infinite_coordinates_get_unknown_id`.

**`area_peril_id` stays in the integer domain.** The accumulator and the mapped array are `int64`
rather than `float64`, so an id above 2^53 keeps its exact value. An H3 model that uses the cell
index itself as the area peril id needs no mapping table beyond an identity file, and cells are
~6e17 — every id would otherwise have come back rounded. Covered by
`test_build_h3_area_peril_id_above_float64_exact_range` and `test_build_h3_cell_index_as_area_peril_id`.
This is also slightly faster, since it removes the float→`Int64` cast work in `set_id_columns`.

Two smaller fixes fall out of the rewrite and are worth noting:

- A mapping row with `h3_int64 == 0` no longer assigns its `area_peril_id` to every
  null-coordinate location. `0` was previously both the invalid-cell sentinel and a join key.
- Locations already carrying an `area_peril_id` column are overwritten cleanly, instead of
  producing `area_peril_id_x`/`area_peril_id_y` from the merge suffixes and then a third, all-`-1`
  column. This matters to chained and `combine` strategies.

## Tests

`tests/lookup/test_builtin_h3.py`, 22 tests. `_row_wise_h3_int64` keeps the deleted loop as a
reference implementation, and the equivalence test guards against a vacuous comparison by
asserting that both the matched and unmatched populations are well represented. Row count and
order, non-contiguous index reset, nullable `Float64` coordinates, parquet mappings, the `Int64`
output dtype and both build-time raises are covered.

### DIFF
--- a/oasislmf/lookup/builtin.py
+++ b/oasislmf/lookup/builtin.py
@@ -960,8 +960,7 @@ class Lookup(AbstractBasicKeyLookup, MultiprocLookupMixin):
         return geotiff_lookup
 
     def build_h3(self, resolution, file_path, file_type='csv', **kwargs):
-        """
-        Function factory to look up area_peril_id using H3 hexagonal grid indexing.
+        """Function factory to look up area_peril_id using H3 hexagonal grid indexing.
 
         Converts latitude/longitude to an H3 cell at the specified resolution,
         converts the cell to its int64 representation, then maps to an int32
@@ -970,6 +969,14 @@ class Lookup(AbstractBasicKeyLookup, MultiprocLookupMixin):
         The mapping file must contain at least two columns:
             - ``h3_int64``      : H3 cell index as a 64-bit integer
             - ``area_peril_id`` : Oasis area peril ID (int32)
+
+        ``h3_int64`` must be unique: a point falls in exactly one cell at a given resolution, so a
+        repeated cell index is a mapping file error and raises ``OasisException`` at build time.
+        A mapping file with duplicate cells previously emitted one key per duplicate row, which
+        changed the row count of the keys file rather than reporting the problem.
+
+        A location whose coordinates are null, infinite, or absent from the mapping file resolves
+        to ``OASIS_UNKNOWN_ID``, which the lookup reports as a per-location ``fail`` status.
 
         Config example::
 
@@ -982,18 +989,18 @@ class Lookup(AbstractBasicKeyLookup, MultiprocLookupMixin):
                 }
             }
 
-        Parameters
-        ----------
-        resolution : int
-            H3 resolution level (0–15). Higher values produce finer cells.
-        file_path : str
-            Path to the int64→area_peril_id mapping file.
-            Supports the ``%%KEYS_DATA_PATH%%`` placeholder.
-        file_type : str
-            Pandas read function suffix (``'csv'``, ``'parquet'``, etc.).
-            Defaults to ``'csv'``.
-        **kwargs
-            Additional keyword arguments forwarded to the pandas read function.
+        Args:
+            resolution (int): H3 resolution level (0–15). Higher values produce finer cells.
+            file_path (str): Path to the int64→area_peril_id mapping file.
+                Supports the ``%%KEYS_DATA_PATH%%`` placeholder.
+            file_type (str): Pandas read function suffix (``'csv'``, ``'parquet'``, etc.).
+                Defaults to ``'csv'``.
+            **kwargs: Additional keyword arguments forwarded to the pandas read function.
+
+        Returns:
+            function: function assigning an area_peril_id to each location from its latitude and
+                longitude, set to OASIS_UNKNOWN_ID where the H3 cell is missing from the mapping
+                file or the coordinates are null or non-finite.
         """
         if h3 is None:
             raise OasisException(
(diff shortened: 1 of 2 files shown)


oasislmf/lookup/builtin.py
```diff
@@ -976,29 +987,51 @@ def build_h3(self, resolution, file_path, file_type='csv', **kwargs):
...
-            locations['h3_int64'] = 0  # 0 correponds to an invalid H3 index
+        # a point falls in exactly one cell at a given resolution, so a repeated cell index is a
+        # mapping file error; fail here rather than silently emitting a key per duplicate row
+        duplicated_cells = h3_mapping_df['h3_int64'].duplicated()
+        if duplicated_cells.any():
+            raise OasisException(
+                f"H3 mapping file must have a unique 'h3_int64' per row, found {duplicated_cells.sum()} "
+                f"duplicate(s), starting with {h3_mapping_df.loc[duplicated_cells, 'h3_int64'].iloc[0]}"
+            )
+        area_peril_by_cell = h3_mapping_df.set_index('h3_int64')['area_peril_id']
```
SkylordA: `area_peril_by_cell = h3_mapping_df.set_index('h3_int64')['area_peril_id']` validates neither the dtype nor the nullness of `area_peril_id`, unlike `h3_int64` which gets an explicit uniqueness check just above. A single blank `area_peril_id` cell forces the whole column to float64, silently rounding any other id above 2^53 in that column:

```python
>>> pd.Series([1.0, 2.0, np.nan, 9007199254740993.0]).to_numpy(dtype='int64')
[1, 2, -9223372036854775808, 9007199254740992]   # NaN -> garbage; the big id already lost its +1
```

This defeats the "stays in the integer domain" guarantee at the source, before `h3_lookup` runs. The NaN case itself turns out to be masked downstream: `set_id_columns`'s `df.loc[(df[col].isna()) | (df[col] <= 0), col] = OASIS_UNKNOWN_ID` ([oasislmf/lookup/builtin.py:451](oasislmf/lookup/builtin.py#L451)) catches it, because NaN-to-int64 saturates to each ISA's "integer indefinite" value (`INT64_MIN` on x86-64, `0` on ARM64), and both are `<= 0`. So this is unlikely to be a live correctness bug today — but it's an accident of hardware cast semantics rather than a designed safeguard, and isn't guaranteed to survive a future numpy/pandas change to the cast path. Given the PR's whole thesis is closing exactly this class of silent-wrong-value bug (see the `isfinite` fix), an explicit non-null check on `area_peril_id` at build time — mirroring the `h3_int64` duplicate check — would replace incidental protection with a real one. Not blocking.

SkylordA (approved):

sstruzik (approved):

PR #2105, opened 2026-08-14
Feature/decouple websocket

sambles:
### Fixes to run platform without redis websocket

Works with https://github.com/OasisLMF/OasisPlatform/pull/1427 add an option to send event processing updates via a HTTP instead of websocket 

#### Bug: periodic ping timer never reset in `gul/manager.py`
- **`oasislmf/pytools/gul/manager.py`** — the periodic-ping loop reset `counter = 0` after each ping but never reset `timer`. Once `SERVER_UPDATE_TIME` first elapsed, the stale `timer` meant every subsequent event satisfied the "time elapsed" check, so pings fired on every event for the rest of the run instead of being throttled to the interval. `gulmc/manager.py` already did this correctly (`timer = time.time()`); `gul/manager.py` now matches.

#### `oasislmf/utils/ping.py` updates
- **`port_override` ordering bug**: `json.dumps(data)` was called *before*
  `data.pop('port_override', None)`, so the key leaked into the outgoing socket message despite
  the docstring claiming it's stripped first. Fixed by popping before serializing.
- **Narrow exception handling**: `oasis_ping_socket` only caught `ConnectionRefusedError`. Other
  realistic failures (`TimeoutError`, DNS resolution failures) were left uncaught and would
  crash the calling worker process just because a progress ping failed. Now catches
  `(ConnectionError, TimeoutError, socket.gaierror)`.
- **Logging routed to the wrong logger**: all `logging.error(...)` calls used the bare
  module-level shortcut, which logs on the **root** logger — a different, disconnected object
  from the `oasislmf.*` logger hierarchy that `gulmc`/`gulpy`/`modelpy` actually configure
  handlers on via `redirect_logging`/`OASIS_PYTOOLS_LOG_LEVEL`. Ping errors were reaching stderr
  only via Python's unformatted "handler of last resort", untagged and inconsistent with the
  rest of each pytool's log output. Fixed by using `logger = logging.getLogger(__name__)`
  throughout.
- **Added debug visibility**: a `logger.debug(...)` line before each send attempt (HTTP,
  websocket, and plain socket), logging the target and payload. Naturally gated by
  `OASIS_PYTOOLS_LOG_LEVEL` — silent unless that's set to `DEBUG`, no separate flag needed.

### DIFF
--- a/oasislmf/computation/generate/losses.py
+++ b/oasislmf/computation/generate/losses.py
@@ -682,7 +682,7 @@ class GenerateLosses(GenerateLossesDir):
             has_http = 'OASIS_ANALYSIS_STATUS_URL' in os.environ
             if 'analysis_pk' in self.kwargs and not (has_websocket or has_http):
                 self.logger.info("Set `OASIS_WEBSOCKET_URL`/`OASIS_WEBSOCKET_PORT` or `OASIS_ANALYSIS_STATUS_URL` "
-                                  "environment variables for run progress updates")
+                                 "environment variables for run progress updates")
             elif 'analysis_pk' in self.kwargs:
                 oasis_ping({"analysis_pk": self.kwargs["analysis_pk"], 'events_total': str(os.path.getsize("input/events.bin") // oasis_int_size)})
             else:
--- a/oasislmf/computation/generate/losses.py
+++ b/oasislmf/computation/generate/losses.py
@@ -693,8 +693,11 @@ class GenerateLosses(GenerateLossesDir):
         with setcwd(model_run_fp):
             socket_server_size = None
             socket_server_port = None
-            if 'analysis_pk' in self.kwargs and not all(item in os.environ for item in ['OASIS_WEBSOCKET_URL', 'OASIS_WEBSOCKET_PORT']):
-                self.logger.info("Set `OASIS_WEBSOCKET_URL` and `OASIS_WEBSOCKET_URL` environment variables for run progress updates")
+            has_websocket = all(item in os.environ for item in ['OASIS_WEBSOCKET_URL', 'OASIS_WEBSOCKET_PORT'])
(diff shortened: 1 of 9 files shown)


oasislmf/pytools/gul/manager.py
Ha-Ree: Worth changing here, `counter += 1` should never have been in the while loop, probably just after the log saying event done

Ha-Ree:
Also worth asking, is it worth just deprecating the websocket pings if we are adding this other method?

sambles:
> Also worth asking, is it worth just deprecating the websocket pings if we are adding this other method?

I think we should, but maybe not straight away.  We could flag it for deprecation and keep both until the next stable release?

oasislmf/computation/generate/losses.py
```diff
@@ -678,8 +678,11 @@ def run(self):
         with setcwd(model_run_fp):
             socket_server_size = None
             socket_server_port = None
-            if 'analysis_pk' in self.kwargs and not all(item in os.environ for item in ['OASIS_WEBSOCKET_URL', 'OASIS_WEBSOCKET_PORT']):
-                self.logger.info("Set `OASIS_WEBSOCKET_URL` and `OASIS_WEBSOCKET_URL` environment variables for run progress updates")
+            has_websocket = all(item in os.environ for item in ['OASIS_WEBSOCKET_URL', 'OASIS_WEBSOCKET_PORT'])
+            has_http = 'OASIS_ANALYSIS_STATUS_URL' in os.environ
```
Ha-Ree, line 682: Could be worth making a function for websocket/http as also used in bash.py

sambles:
comment from linked platform PR needs fixing here:
https://github.com/OasisLMF/OasisPlatform/pull/1427#issuecomment-5437426807

Ha-Ree (approved):
Only thing found is potentially adding a return status to async and resetting counter only on success

PR #2106, opened 2026-08-14
Fix for flaky test

### LINKED ISSUE #2093: Fix Flaky test in API client checks
https://github.com/OasisLMF/OasisLMF/actions/runs/31392744657/job/93468391644

**FileEndpointTests.test_download__file_exists_exception_raised**

[test-error.txt](https://github.com/user-attachments/files/30903233/test-error.txt)

sambles:
### Fix for flaky CI test
See https://github.com/OasisLMF/OasisLMF/issues/2093

### DIFF
--- a/tests/platform_api/test_client.py
+++ b/tests/platform_api/test_client.py
@@ -18,7 +18,9 @@ import pytest
 from posixpath import join as urljoin
 
 import requests
+from requests.adapters import HTTPAdapter
 from requests.exceptions import HTTPError
+from urllib3.util.retry import Retry
 from requests_toolbelt import MultipartEncoder
 import responses
 from responses.registries import OrderedRegistry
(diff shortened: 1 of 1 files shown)


tests/platform_api/test_client.py
```diff
@@ -69,6 +71,20 @@
 DISABLE_DATA_CHECKS = version.parse(responses_ver) >= version.parse("0.25.3")
 
 
+def fetch_with_retries(url, retries=5, backoff_factor=1.0):
+    session = requests.Session()
```
Ha-Ree: Probably should be
`with requests.Session() as session:`

tests/platform_api/test_client.py
```diff
@@ -69,6 +71,20 @@
...
+    retry = Retry(
+        total=retries,
+        connect=retries,
+        read=retries,
```
vinulw: minor: I think `total` takes precendence over all other counts so `connect` and `read` probably redundant?

tests/platform_api/test_client.py
```diff
@@ -69,6 +71,20 @@
 DISABLE_DATA_CHECKS = version.parse(responses_ver) >= version.parse("0.25.3")
 
 
+def fetch_with_retries(url, retries=5, backoff_factor=1.0):
```
vinulw: not sure if the default backoff factor is too high. With 5 retries the delay between retries in the worst case will be $\text{backoff factor} \cdot 2^{\text{max retries}} = \text{backoff factor} \cdot 32s = 32s$ and the gap between retries is quite large (2s, 4s ...).

sambles: good point, i'll drop it way down to `0.1`

vinulw (approved):
👍🏾

Ha-Ree (approved):
Thought I approved this one yesterday xD

PR #2107, opened 2026-08-17
Exposure summary vectorise

### LINKED ISSUE #2100: Vectorisation of slower functions

Ha-Ree:
Part of #2100.

`get_exposure_summary` filtered `df_summary_peril` down to one (field value, status, coverage
type) combination at a time and summed it, which is `n_fields × 7 statuses × n_values × 13
coverages` passes over the whole frame. This replaces that with a small number of grouped passes:
the key columns are coded to integers once up front, then each field is deduplicated and grouped
in one go.

## Performance

1,600,000 rows / 100,000 locations / 3 summary fields:

| implementation | time |
|---|---|
| `main` (the per-combination filter loop) | 18.70 s |
| this PR | **5.87 s (3.2x)** |

The coding step is what makes the rest cheap: every deduplication and grouping hashes its key
columns, and for the object columns holding peril ids, lookup statuses and OED field values that
hashing dominated the summary. After `pd.factorize` each pass hashes integers instead.

## Behaviour

Output is unchanged. `tests/preparation/test_exposure_summary_fields.py` keeps the old
per-combination implementation verbatim as `reference_exposure_summary_field` /
`reference_exposure_summary_fields` and diffs the whole nested summary leaf by leaf, so the new
code is pinned against the old behaviour rather than against itself. `assert_summaries_equal`
checks the key sets match first, uses `pytest.approx` for floats, and then re-checks every
integer-valued leaf both exactly and for `isinstance(..., int)` — so a `np.int64` leaking into the
JSON would fail.

Two things are worth a reviewer's attention:

**The deduplication is per field and is not shared across fields.** The field value is part of the
key the row-wise version deduplicated on, because it filtered by field value before deduplicating
on `(loc_id, peril_id)`. Hoisting one shared dedupe out of the field loop looks safe only if every
summary field holds one value per location — and that does not hold. `loc_id` identifies
`(PortNumber, AccNumber, LocNumber)` rather than a row, so two exposure rows with the same
`LocNumber` share one, and `DEFAULT_VALIDATION_CONFIG` has no uniqueness rule that rejects them.
Sharing the key would drop one of the two rows and take its entire field bucket to zero — every
leaf, every status, every coverage type — while the portfolio total stayed put, so the breakdown
would silently stop reconciling. `test_one_loc_id_carrying_two_field_values` covers it.

Note the flip side, which is pre-existing and deliberate: a location carrying two country codes
contributes to *both* buckets, so the country breakdown sums to more than the peril breakdown.
`test_one_loc_id_carrying_two_field_values_counts_it_under_both` pins that asymmetry so the
deduplication cannot quietly resolve it either way.

**The coded columns have their own namespace.** A summary field is named after an OED column —
`summary_report_fields` in the model settings, converted with `convert_col_name` — while the
summary sums columns of its own. `NumberOfBuildings` is an ordinary field to want a breakdown by,
and it arrives here as `number_of_buildings`, exactly the name `COUNT_AGGREGATION` sums. Writing
the codes under the raw column name would overwrite the values and then sum the codes, giving a
silently wrong building or risk count in a published report rather than a crash. The codes are
held under `code_column(name)` instead, which also makes `coverage_type_id` and `loc_id` usable as
summary fields. `test_field_named_after_a_summed_or_key_column` covers all four.

**TIV shifts by ~1 ULP.** `Series.sum()` on a float64 block uses numpy's pairwise summation,
whereas `groupby(...)['tiv'].sum()` accumulates sequentially per group, so the two are not
bit-identical — worst relative difference measured at 3.02e-16, e.g. `5046643.199999999` becoming
`5046643.2`. This is visible in a diff of `exposure_summary_report.json` even though it is
numerically meaningless. Nothing in the repo compares that file and the new tests use
`pytest.approx`, but it is worth knowing about if you diff the report downstream.

## Tests

16 tests. Beyond the reference comparison: field values absent from the data, an empty frame,
categorical key columns, duplicate keys, statuses summing to `all`, the shared-`loc_id` cases and
the colliding-field-name cases above. I verified the six new tests are discriminating — five of
them fail against the shared-dedupe version and pass here.


oasislmf/preparation/summaries.py
```diff
@@ -971,19 +1019,37 @@ def get_exposure_summary(
...
             for field_value in field_list:
-                exposure_summary = get_exposure_summary_field(
-                    df_summary_peril_status, exposure_summary, field_name,
-                    field_value, status
-                )
+                field_summary = exposure_summary[field_name][field_value][status]
+                field_code = codes[field_name].get(field_value)
+
+                for coverage_type, coverage_type_id in coverage_ids:
+                    coverage_key = (field_code,) + group_key + (coverage_type_id,)
```
SkylordA, line 1064: A summary field named `'status'` gets every non-`'all'` count and TIV silently zeroed.

**Cause**: `get_exposure_summary_stats` collapses `group_keys` to one element when `field_name == 'status'` (avoiding a duplicate `code_column('status')`), so `overall`/`tiv`/`by_coverage` end up keyed by shorter tuples than normal. The caller, `get_exposure_summary_fields`, builds its lookup keys (`coverage_key` at line 1041, and line 1050) without the same collapse — one element too long for `field_name == 'status'`. Every `.get()` misses, so every non-`'all'` bucket returns the zeroed default. `'all'` takes a different path and stays correct. `field_name == 'peril_id'` doesn't collide and is unaffected.

**Confirmed** by extracting the four new functions and running them directly, two seeds:

```
field_name='status', field_value='success':
  all      tiv=11736.73     <- correct
  success  tiv=0.00         <- should equal the 'all' row (field_value == status)
  fail     tiv=0.00         <- correctly 0 (empty intersection)
```

**Reachable in production**: `get_exposure_summary` builds `oed_categories` from `additional_fields` run through `convert_col_name` (`'Status'` → `'status'`), so any exposure file with a column that converts to `'status'`, passed via `additional_fields`, triggers this.

**Untested**: `oed_categories_for` (used by every test in the new file) only ever supplies `'peril_id'` and `'country_code'` — worth adding `'status'` alongside the existing `number_of_buildings`/`number_of_risks`/`coverage_type_id`/`loc_id` collision cases.

SkylordA (changes requested):

oasislmf/preparation/summaries.py
Ha-Ree: Fixed with a test to confirm

SkylordA (approved):

oasislmf/preparation/summaries.py
```diff
@@ -807,46 +883,25 @@ def get_exposure_totals(df):
...
-            "tiv": portfolio_tiv,
-            "number_of_locations": portfolio_num,
-            "number_of_buildings": portfolio_num_buildings,
-            "number_of_risks": portfolio_num_risks,
-        }
+    within_scope = df['status'].isin(OASIS_KEYS_STATUS_MODELLED)
+    scopes = {
+        "modelled": df[within_scope],
+        "not-modelled": df[~within_scope],
+        "portfolio": df,
```
vinulw: minor: might be better to just store the masks instead of copying multiple instances of the entire df

oasislmf/preparation/summaries.py
```diff
@@ -753,46 +753,122 @@ def generate_summaryxref_files(
...
+    coverage_keys = group_keys + ['coverage_type_id']
+
+    # each frame is deduplicated from the one above it, which is already the first row per group
+    tiv_rows = encoded.drop_duplicates(group_keys + ['loc_id', key_column('peril_id'), 'coverage_type_id'])
+    coverage_rows = tiv_rows.drop_duplicates(coverage_keys + ['loc_id'])
+    location_rows = coverage_rows.drop_duplicates(group_keys + ['loc_id'])
+
+    tiv = tiv_rows.groupby(coverage_keys, observed=True, sort=False)['tiv'].sum()
+    by_coverage = coverage_rows.groupby(coverage_keys, observed=True, sort=False).agg(**COUNT_AGGREGATION)
+    overall = location_rows.groupby(group_keys, observed=True, sort=False).agg(**COUNT_AGGREGATION)
```
vinulw: minor: not sure if `COUNT_AGGREGATION` needs to be a global variable yet, not used elsewhere and makes it a bit difficult to understand whats going on here when the declaration is global

tests/preparation/test_exposure_summary_fields.py
```diff
@@ -0,0 +1,451 @@
...
+import pytest
+
+from oasislmf.preparation.summaries import get_exposure_summary_fields, get_exposure_totals
+from oasislmf.utils.coverages import SUPPORTED_COVERAGE_TYPES
+from oasislmf.utils.status import OASIS_KEYS_STATUS, OASIS_KEYS_STATUS_MODELLED
+
+STATUSES = ['all'] + list(OASIS_KEYS_STATUS.keys())
+
+
+def reference_exposure_summary_field(df, exposure_summary, field_name, field_value, status):
```
vinulw: makes sense to use reference implementation to check outputs for the rewrite but I dont think we need to add this to the permanent test suite. 

It would be better to serialise the df / dict produced by the reference function and unittest the functions against the serialised outputs, then we don't need the reference implementation to live here (same comment applied to `reference_exposure_summary_fields`).

vinulw (commented):
Main comment is the one about the testing against a reference.

vinulw (approved):

PR #2108, opened 2026-08-17
Dummy model vectorise

### LINKED ISSUE #2100: Vectorisation of slower functions

Ha-Ree:
Part of #2100.

The dummy model generators built their files a row at a time, packing each row with
`struct.pack` and writing it. This generates each file as numpy arrays instead, in chunks, and
writes the chunks straight out with `ndarray.tofile`.

## Performance

`FootprintBinFile` at 100 events × 400 areaperils × 25 intensity bins: **15.18 s → 0.07 s (216x)**,
files identical.

## Output is byte-for-byte identical

Every generator keeps its row-by-row `generate_data` as the reference implementation, and
`tests/computation/test_dummy_model_generate.py` pins the array path against it per file class,
including dtype. This is the whole correctness argument, so it is worth being precise about what
holds it up:

- The chunk boundaries never move the RNG stream. Each generator draws in blocks whose shape keeps
  the per-item draws contiguous, so drawing N items at once takes the same numbers from the
  generator as drawing them one at a time.
- `array_dtype` builds a **packed** dtype, matching `struct.pack('=iqq')` and the reader's
  `EventIndexBin_dtype` in `oasislmf/pytools/getmodel/common.py`. `align=True` would silently give
  24-byte index records instead of 20.
- `ItemsFile` and `OccurrenceFile` keep their per-row draw order. `ItemsFile` does get a
  `generate_arrays`, but its areaperil and vulnerability draws alternate per location and are
  bounded differently, so only the assembly is vectorised and the draws themselves stay per
  location. `OccurrenceFile` keeps row-by-row generation entirely — its row count per event is
  itself random — and is assembled whole by the base class `np.fromiter`. They are the residual
  hot spots.

One caveat to "byte for byte identical": with `num_events=0` the footprint now produces an empty
`footprint.idx`, where before no idx file was created at all. A footprint with no index file is not
readable; an empty index is at least well-formed, so this is kept rather than restored.

## Memory

`CHUNK_ROWS` bounds how many rows are built before being written. `FootprintBinFile` is the only
file that gets genuinely large, and its peak memory is unbounded in
`areaperils_per_event × num_intensity_bins` unless the chunking also applies *within* an event —
the draws alone are `(areaperils_per_event, 2, num_intensity_bins)` float64, roughly 68 bytes of
transient RAM per output row against 12 bytes on disk.

So an event is taken in chunks of areaperils rather than all at once, and the index entry is
written once the last chunk of that event is done. Measured peak RSS, one event, sparseness 1.0:

| areaperils | bins | rows | footprint.bin | peak RSS before | peak RSS after |
|---|---|---|---|---|---|
| 50,000 | 20 | 1.0M | 11.4 MB | 83.8 MB | 85.1 MB |
| 200,000 | 20 | 4.0M | 45.8 MB | 236.4 MB | 121.9 MB |
| 500,000 | 20 | 10.0M | 114.4 MB | 544.6 MB | 131.9 MB |

`oasislmf test generate-model-files -a 2000000 -i 50` — 2M areaperils, 50 intensity bins, all
reachable through documented CLI flags — extrapolates to roughly 5.4 GB unchunked, an OOM kill on a
4 GB runner. It now writes its 1.1 GB `footprint.bin` in **152.8 MB** of peak RSS.

I verified the chunking is stream-neutral over 144 configurations (sparseness 0.0/0.3/0.7/1.0 ×
`no_intensity_uncertainty` × `areaperils_per_event` above and below `num_areaperils` × 1/4/9
intensity bins × 0/1/5 events) crossed with `CHUNK_ROWS` forced to 1, 2, 3, 7, 13 and 10⁶ — 864
SHA-256 comparisons of `footprint.bin` and `footprint.idx` against the unchunked generator, zero
mismatches. `test_footprint_chunking_does_not_change_the_output` pins that with a `CHUNK_ROWS`
monkeypatch, and `test_footprint_index_describes_whole_events_when_chunked` pins that the index
still describes whole events rather than chunks.

### What is not bounded

`CHUNK_ROWS` is applied to three classes — `FootprintBinFile`, `VulnerabilityFile` and
`LossFactorsFile`. The other generators build the whole file in a single array, plus full-length
int64 `repeat`/`tile` temporaries, where the row path was constant-memory. On the portfolio side
that is a regression rather than an improvement:

| generator | inputs | file size | peak RSS before | peak RSS after |
|---|---|---|---|---|
| `FMXrefFile` | `-l 2000000 -c 4 --num-layers 3` (24M rows) | 288 MB | 30 MB | 549 MB |
| `ItemsFile` | `-l 2000000 -c 4` (8M rows) | 160 MB | 33 MB | 384 MB |
| `OccurrenceFile` | `-e 2000000` (2M rows) | 24 MB | 33 MB | 56 MB |

`-l 8000000 -c 4 --num-layers 3` extrapolates to roughly 2.7 GB for `fm_xref.bin` alone — an OOM
kill on the same 4 GB runner this section uses as its yardstick for the footprint, on a run that
previously peaked around 30 MB.

These are the easy cases to chunk: the ids are pure `arange`/`repeat`/`tile` with no RNG
involved, so they need no stream-order argument at all, just `CHUNK_ROWS` rows at a time with the
`arange` starts offset, exactly as `LossFactorsFile.generate_arrays` already does. `ItemsFile`
already loops over locations, so it only needs to yield every N locations rather than
accumulate. Follow-up rather than part of this PR — but the section above should not be read as
"peak memory is now bounded generally", because outside those three classes it is not.

## Also fixed

Re-running the generator into the same directory used to append to `footprint.idx` while
truncating `footprint.bin`, leaving a corrupt model — idx sizes went 100 → 200 → 300 bytes over
three runs of a 5-event footprint. `write_array` truncates, so it is now 100 → 100 → 100.

## Tests

`tests/computation/test_dummy_model_generate.py`, 76 tests. It parametrises over every `ModelFile`
subclass asserting `generate_arrays` equals `generate_data` including dtype, repeats over
multi-chunk sizes, sweeps sparseness and `no_intensity_uncertainty`, checks the footprint index
offsets against each event's `nbytes`, and round-trips every file back off disk with
`np.fromfile`. Retaining `generate_data` is what makes this possible — it is the executable
specification, not dead code, and `debug_write_file` still uses it.

### DIFF
--- a/oasislmf/computation/data/dummy_model/generate.py
+++ b/oasislmf/computation/data/dummy_model/generate.py
@@ -26,6 +26,14 @@ from math import erf
 import numpy as np
 
 
+# the file layouts are declared as struct format characters, which the arrays written to
+# the files have to match exactly, so the two are kept in step through this mapping
+STRUCT_TO_NUMPY_FORMAT = {'i': 'i4', 'f': 'f4', 'q': 'i8'}
+
+# rows generated before being written, keeping the memory used by a chunk in the tens of MBs
+CHUNK_ROWS = 10 ** 6
+
+
 class ModelFile:
     """Base class for all dummy model files.
 
@@ -33,20 +41,30 @@ class ModelFile:
     typical order of execution is as follows:
         1. Initialise class attributes and methods (__init__).
         2. Set random seed (seed_rng).
-        3. Generate random data (generate_data).
-        4. Convert random data to binary format and write to file. This step is
-        done as each line of data is generated to minimise memory use
-        (write_file).
+        3. Generate random data (generate_arrays, or generate_data for the
+        classes that generate their data a row at a time).
+        4. Write the generated data to file in binary format, in chunks to
+        minimise memory use (write_file).
 
     Attributes:
         seed_rng: Seed random number generator.
         write_file: Write data to output file in binary format.
         debug_write_file: Write data to screen in csv format.
-        generate_data: Generate dummy model data.
+        generate_data: Generate dummy model data row by row.
+        generate_arrays: Generate dummy model data in chunks.
     """
 
     def __init__(self):
         pass
+
+    @property
+    def array_dtype(self):
+        """Get the numpy dtype of the rows written to the file.
+
+        Returns:
+            numpy.dtype: packed dtype matching the file's struct format
+        """
+        return np.dtype([(name, STRUCT_TO_NUMPY_FORMAT[fmt]) for name, fmt in self.dtypes.items()])
 
     def seed_rng(self):
         """Seed random number generator.
@@ -69,16 +87,15 @@ class ModelFile:
     def write_file(self):
         """Write data to output file in binary format.
 
-        General method to convert generated data to binary format and write to
-        file. Calls chlid class-specific generate_data method.
+        General method to write the generated data to file. Calls child
+        class-specific generate_arrays method.
         """
         with open(self.file_name, 'wb') as f:
             if self.start_stats:
                 for stat in self.start_stats:
                     f.write(struct.pack(stat['dtype'], stat['value']))
-            dtypes_list = ''.join(self.dtypes.values())
-            for line in self.generate_data():
-                f.write(struct.pack('=' + dtypes_list, *(line)))
+            for chunk in self.generate_arrays():
+                chunk.tofile(f)
 
     def debug_write_file(self):
         """Write data to screen in csv format.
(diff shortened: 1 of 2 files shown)


SkylordA (approved):
Small note: the footprint index was opened in append mode before and is now overwritten, so re-running into an existing directory no longer duplicates rows. Worth a line in the PR description.

oasislmf/computation/data/dummy_model/generate.py
```diff
@@ -782,6 +1000,39 @@ def generate_data(self):
...
+        two draws alternate per location, so they have to stay in that order to take the same
+        numbers from the generator. Only the ids and the packing are built with numpy.
+
+        Yields:
+            numpy.ndarray: rows of item data, in the file's dtype
+        """
+        super().seed_rng()
+        shape = (self.num_locations, self.coverages_per_location)
+        areaperils = np.empty(shape, dtype='int64')
+        vulnerabilities = np.empty(shape, dtype='int64')
```
vinulw: minor: can infer these from the `self.array_dtype`. This comment applies in one place below (marked dtype)

oasislmf/computation/data/dummy_model/generate.py
```diff
@@ -902,6 +1167,22 @@ def generate_data(self):
...
+            numpy.ndarray: rows of programme data, in the file's dtype
+        """
+        num_aggs = self.num_locations * self.coverages_per_location
+        agg_id = np.arange(1, num_aggs + 1)
+
+        programme = np.empty(num_aggs * 2, dtype=self.array_dtype)   # 2 from number of levels
+        programme['from_agg_id'] = np.tile(agg_id, 2)
+        programme['level_id'] = np.repeat([1, 2], num_aggs)
+        # Site coverage FM level aggregates to itself, policy layer FM level to the programme
+        programme['to_agg_id'] = np.concatenate([agg_id, np.ones(num_aggs, dtype='int64')])
```
vinulw: dtype

vinulw (approved):
Suggestion about dtypes is minor, happy to approve with out without update.

PR #2111, opened 2026-08-17
Array operations speedup

### LINKED ISSUE #2100: Vectorisation of slower functions

### LINKED ISSUE #2114: join-summary-info crashes on parquet when no SummaryId matches the summary info
## Summary

`join-summary-info` crashes on a parquet file when every `SummaryId` in it is beyond the end of the summary info, instead of joining the rows to empty summary columns:

```
ValueError: Columns must be same length as key
```

The CSV path of the same tool handles this case without complaint, so the two formats disagree.

## Reproduce

```python
import pandas as pd
from pathlib import Path
from tempfile import TemporaryDirectory
from oasislmf.pytools.join_summary_info.manager import main

with TemporaryDirectory() as d:
    summary_info, data, out = Path(d, "gul_summary-info.parquet"), Path(d, "aalgul_ord.parquet"), Path(d, "joined.parquet")
    pd.DataFrame({"summary_id": [1, 2], "PortNumber": ["1", "2"], "tiv": [10.0, 20.0]}).to_parquet(summary_info, index=False)
    # every SummaryId is past max_summary_id
    pd.DataFrame({"SummaryId": [7, 8], "MeanLoss": [1.0, 2.0]}).to_parquet(data, index=False)
    main(summaryinfo=summary_info, data=data, output=out)   # ValueError
```

Swapping `.parquet` for `.csv` throughout writes the file with empty summary columns, as expected.

## Cause

In the parquet branch of `run()` (`oasislmf/pytools/join_summary_info/manager.py`), a summary id past the end of the summary info joins to the empty string `""`, and the joined column is then split out with `.str.split(",", expand=True)`.

`""` splits into **one** field, whatever the number of summary columns. When at least one row joins normally, the split still widens to the full number of columns and the unmatched row's trailing columns come out missing. When *no* row joins, the split produces a single column, and assigning it to N headers raises.

This also means a partly-unmatched file is written inconsistently today: an unmatched row gets `""` in the first summary column and null in the rest, whereas a summary id that falls in a **gap** within the summary info gets `""` in every column (`load_summary_info` fills gaps with `"," * (len(headers) - 1)`). Both are the same situation — no summary info for this id — so both should give the same thing, and the CSV path already writes empty for every column in both.

## Fix

Join out-of-range summary ids to the same `"," * (len(summary_headers) - 1)` that gaps already use, so the split always yields one field per header. That removes the crash and makes the parquet output agree with the CSV output and with the gap case.

Note this changes the written output for the partly-unmatched case: the trailing summary columns of an unmatched row become empty strings rather than nulls.

## Affected

Present on `main`; not a regression. Reproduced on both the pandas object dtype and pandas 3's `StringDtype`.

Ha-Ree:
Part of #2100.

Seven small row-wise passes over data that scales with the exposure or the output, each replaced by the array operation it was spelling out by hand:

- gul/gulmc build_structures: the peril filter mask was a python set lookup per keys row, now np.isin (2M keys rows: 0.68s -> 0.006s). The keys dtype and the filter itself were duplicated between the two callers and are now shared as `KEYS_DTYPE` / `filter_area_peril_id` in `pytools/common/input_files.py`, which is also what makes them testable. This fixes a latent crash: on an empty keys table the list comprehension produced a float64 array, so indexing with it raised `IndexError: arrays used as indices must be of integer (or boolean) type`.
- gulmc get_dynamic_footprint_adjustments: the zero-adjustment fallback built one tuple per item, now np.zeros with the item ids assigned in
- il_inputs: the StepTriggerType sub-type was resolved by a row-wise apply with try/except, now a lookup indexed by (steptriggertype, coverage_type_id). The five lines that apply it are extracted as `assign_sub_step_trigger_types` so the tests exercise the production code rather than a copy of it; a pair with no sub-type keeps its trigger type as before.
- join_summary_info: the summary info was scattered by index in a loop and, for parquet, joined per row by agg(",".join) and indexed per row by apply
- getmodel _build_footprint: groupby.transform(lambda x: x / x.sum()) per areaperil, now a divide by transform('sum')
- exposure run: apply(str) per column, now astype(str)

None of the changed lines is inside an `@njit` function, so none of the numba hazards — nopython support, object-mode fallback, allocation inside prange, or the known gulpy-vs-gulmc fastmath codegen divergence — are in play here.

## The one that is not bit-for-bit equal

The footprint probabilities now divide by a groupby sum, which differs from summing each group separately in the last bits of a float64. Measured over 400 randomised trials the largest relative difference is 5.97e-16, one to two ULP, appearing at group sizes 10, 100 and 100000.

`probability` in `EventDynamic` is `oasis_float`, which defaults to f4, and over 301,689 rows **0** values differed after the cast — so the written footprint is identical in practice. Two caveats worth stating rather than papering over: that f4 absorption is empirical, not a proof (a 1e-16 perturbation straddles an f4 rounding boundary with probability ~1e-9 per value, which stops being negligible on a model with hundreds of millions of footprint rows); and under `OASIS_FLOAT=f8`, which is supported and exercised in the test suite, the written probabilities *will* differ in the last bits. The values are a renormalisation, so a 1-ULP shift carries no meaning either way.

## Correction to an earlier version of this description

An earlier version of this description said `validate_vuln_csv_contents`' old `isinstance(x, (int, float))` check "rejected an integer or float32 column because np.int64 and np.float32 are not int or float subclasses". That is wrong. While `isinstance(np.int64(1), int)` is indeed False, `Series.apply` maps over `.astype(object)`, which unboxes to python `int`/`float` — so `int64`, `int32`, `float32` and `float64` columns were **all** accepted already, and none of the cited scenarios was reachable anyway, since the sole call path is `pd.read_csv(file_path)` with no `dtype=`.

The change to `np.issubdtype(dtype, np.number)` is a safe speedup on every reachable input. Its real deltas are all in the stricter direction and none is reachable from `read_csv`:

- a `bool` column: accepted before (`True` -> 1, inside [0,1]), rejected now. The new answer is the better one.
- an `object` column holding python numbers: accepted before, rejected now.
- a pandas extension dtype such as `Float64`: `np.issubdtype` raises `TypeError`, which the function's outer `except Exception` turns into a warning and a False.

All three are now pinned by tests, which matters because `tests/utils/test_data.py` mocks `pandas.read_csv` and injects hand-built frames, so a fixture can reach them even though a real csv cannot.

## Fixes #2114

Found while checking the join_summary_info change against pandas 3's string dtype. A summary id past the end of the summary info joined to `""`, which splits into one field however many summary columns there are: a partly-unmatched file was written with the first summary column empty and the rest null, and a file where nothing matched raised `ValueError: Columns must be same length as key`. It now joins to the same empty columns a gap in the summary info already joins to, which is also what the CSV path already writes. The crash is on main too, not from this branch.

Note this half is a **user-visible output change**, not only a crash fix: a consumer reading the joined parquet now gets `""` where it previously got null, so `.isna()`/`.dropna()`/`fillna` behaviour flips. Since `CHANGELOG.rst` is generated at publish time from PR titles, that change would otherwise reach users under the heading "Array operations speedup". Happy to split it into its own PR if that reads better — see the discussion on #2100.

## Three edges closed while the files were open

Each was reachable and silently wrong or a crash, and each has a test that fails without the fix:

- **A negative `SummaryId` joined to the wrong summary.** `summary_id <= max_summary_id` is true for a negative id, which is a legal numpy index, so the row silently took its summary info counted back from the end of the table. Pre-existing, in both the parquet and csv paths, and now bounded below in both.
- **A summary-info file holding nothing but `summary_id` crashed the parquet path.** `.iloc[:, 0]` on a frame with no remaining columns raises `IndexError`. The csv path already handled it, and the data file now comes back unchanged rather than the join failing.
- **A repeated `--extra-summary-cols` crashed `oasislmf exposure run`.** `extra_summary_cols` comes straight from the CLI and was never checked against the level's own `summary_cols`, so naming a column the level already has gave groupby a grouper that is not 1-dimensional (and, on the gul-only path, gave the frame-wide `astype(str)` a duplicate key). The columns are de-duplicated now. Also pre-existing.

## Tests

`tests/fm/test_fmpy.py` 21 passed (covers step policies end to end), gul/gulmc 1272 passed, and `tests/computation/test_run_exposure.py` 58 passed against five unchanged expected-output CSVs.

New:
- 14 tests for the shared peril filter, diffed against a verbatim copy of the set-membership loop it replaces across list/set/tuple/case-mismatch/prefix/absent filter shapes, plus the empty-keys case — which asserts the *old* implementation raises, pinning the crash the change fixes.
- 12 for the sub-step trigger type lookup and its application, parametrised over `int64`/`int32`/`float64`/`object`/`Int32`/`Int64`. `Int32` is what the real pipeline carries, and it is the one a numpy `astype` cannot handle.
- 7 for the `validate_vuln_csv_contents` probability dtype branch, whose reject path no test reached before.
- 7 for the join, covering the #2114 fix plus the negative-id and no-info-column edges above, with the parquet and csv paths cross-checked against each other.

### DIFF
--- a/oasislmf/computation/run/exposure.py
+++ b/oasislmf/computation/run/exposure.py
@@ -304,8 +304,7 @@ class RunExposure(ComputationStep):
                         total_gul, total_il, total_ri_ceded)
 
             # Convert output cols to strings for formatting
-            for c in group_by_cols:
-                all_losses_df[c] = all_losses_df[c].apply(str)
+            all_losses_df[group_by_cols] = all_losses_df[group_by_cols].astype(str)
 
             if self.print_summary:
                 cols_to_print = all_loss_cols.copy()
--- a/oasislmf/computation/run/exposure.py
+++ b/oasislmf/computation/run/exposure.py
@@ -231,7 +231,7 @@ class RunExposure(ComputationStep):
         elif self.output_level == 'peril_item':
             summary_cols = ['output_id'] + lowest_id_cols + [policy_num, 'coverage_type_id', 'peril_id']
 
-        summary_cols += self.extra_summary_cols
+        summary_cols = list(dict.fromkeys(summary_cols + self.extra_summary_cols))
         for col in self.extra_summary_cols:
             if col in calculated_summary_cols:
                 all_losses_df = calculated_summary_cols[col](all_losses_df)
--- a/oasislmf/pytools/join_summary_info/manager.py
+++ b/oasislmf/pytools/join_summary_info/manager.py
@@ -132,7 +132,11 @@ def run(
 
             summary_id = df["SummaryId"].to_numpy()
             in_range = summary_id <= max_summary_id
-            joined_summary_data = np.full(len(df), "", dtype=object)
+            # a summary id past the end of the summary info has no summary info, the same as
+            # one that falls in a gap within it, so it joins to the same empty columns. They
+            # have to be spelled out, as a row splitting into fewer columns than the headers
+            # is an error rather than a short row
+            joined_summary_data = np.full(len(df), "," * (len(summary_headers) - 1), dtype=object)
             joined_summary_data[in_range] = summary_data[summary_id[in_range]]
             df[summary_headers] = pd.Series(
                 joined_summary_data, index=df.index).str.split(",", expand=True)
--- a/oasislmf/pytools/join_summary_info/manager.py
+++ b/oasislmf/pytools/join_summary_info/manager.py
@@ -3,6 +3,7 @@
 import logging
 import shutil
 import numpy as np
+import pandas as pd
 from pathlib import Path
 import pyarrow.parquet as pq
 import pyarrow as pa
@@ -53,7 +54,9 @@ def load_summary_info(stack, summaryinfo_file):
             raise ValueError("Missing 'summary_id' column in summary info file.")
 
         summary_ids = df["summary_id"].to_numpy(dtype=np.int64)
-        summary_data = df.drop(columns=["summary_id"]).astype(str).agg(",".join, axis=1).to_numpy(dtype=object)
+        summary_columns = df.drop(columns=["summary_id"]).astype(str)
+        summary_data = summary_columns.iloc[:, 0].str.cat(
+            summary_columns.iloc[:, 1:], sep=",").to_numpy(dtype=object)
         headers = [col for col in df.columns if col != "summary_id"]
     else:
         raise ValueError(f"Unsupported file format {summaryinfo_file.suffix}.")
@@ -63,8 +66,7 @@ def load_summary_info(stack, summaryinfo_file):
 
     max_summary_id = summary_ids.max()
     full_summary_data = np.full((max_summary_id + 1,), "," * (len(headers) - 1), dtype=object)
-    for i in range(len(summary_ids)):
-        full_summary_data[summary_ids[i]] = summary_data[i]
+    full_summary_data[summary_ids] = summary_data
 
     return full_summary_data, headers, max_summary_id
 
@@ -128,9 +130,12 @@ def run(
             if "SummaryId" not in df.columns:
                 raise ValueError("Missing 'SummaryId' column in data file.")
 
-            df[summary_headers] = df["SummaryId"].apply(
-                lambda sid: summary_data[sid] if sid <= max_summary_id else ""
-            ).str.split(",", expand=True)
+            summary_id = df["SummaryId"].to_numpy()
+            in_range = summary_id <= max_summary_id
+            joined_summary_data = np.full(len(df), "", dtype=object)
+            joined_summary_data[in_range] = summary_data[summary_id[in_range]]
+            df[summary_headers] = pd.Series(
+                joined_summary_data, index=df.index).str.split(",", expand=True)
             pq.write_table(pa.Table.from_pandas(df), temp_output_file)
         else:
             raise ValueError(f"Unsupported file format {data_file.suffix}.")
--- a/oasislmf/pytools/join_summary_info/manager.py
+++ b/oasislmf/pytools/join_summary_info/manager.py
@@ -55,8 +55,11 @@ def load_summary_info(stack, summaryinfo_file):
 
         summary_ids = df["summary_id"].to_numpy(dtype=np.int64)
         summary_columns = df.drop(columns=["summary_id"]).astype(str)
-        summary_data = summary_columns.iloc[:, 0].str.cat(
-            summary_columns.iloc[:, 1:], sep=",").to_numpy(dtype=object)
+        if summary_columns.shape[1]:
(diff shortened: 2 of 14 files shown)


oasislmf/pytools/join_summary_info/manager.py
```diff
@@ -117,7 +122,7 @@ def run(
                 for line in data_fin:
                     row = line.strip().split(",")
                     summary_id = int(row[summary_id_col_idx])
-                    if summary_id > max_summary_id:
+                    if summary_id < 0 or summary_id > max_summary_id:
                         fout.write(line.strip() + ("," * len(summary_headers)) + "\n")
                     else:
                         fout.write(line.strip() + "," + summary_data[summary_id] + "\n")
```
SkylordA: Same gap the parquet branch had before this PR's `if summary_headers:` fix (line 142): when `summary_headers` is empty, this line always adds one comma while line 126 adds zero, producing ragged CSV. Not reachable via the pipeline's own generators, but `join-summary-info` is a public CLI taking arbitrary files, and `test_join_parquet_with_no_summary_info_columns` already covers this for parquet — worth the same guard/test here for symmetry.

oasislmf/utils/data.py
```diff
@@ -960,7 +960,8 @@ def validate_vuln_csv_contents(file_path):
         ):
             logger.warning("vulnerability_id, intensity_bin_id, and damage_bin_id columns must contain integer values.")
             return False
-        if vuln_df['probability'].apply(lambda x: isinstance(x, (int, float))).all():
+        probability = vuln_df['probability']
+        if pd.api.types.is_numeric_dtype(probability) and not pd.api.types.is_bool_dtype(probability):
             if not (vuln_df['probability'].between(0, 1).all()):
```
SkylordA: `.between(0, 1).all()` defaults to `skipna=True`, so a nullable Float64/Int64 probability column with a missing value passes as valid. Not reachable today (the only caller uses plain `pd.read_csv`, no `dtype_backend`), but `test_nullable_extension_dtype_probability_is_accepted` treats nullable dtypes as supported input — `.all(skipna=False)` would close the gap if that's intentional.

oasislmf/computation/run/exposure.py
```diff
@@ -231,7 +231,7 @@ def run(self):
         elif self.output_level == 'peril_item':
             summary_cols = ['output_id'] + lowest_id_cols + [policy_num, 'coverage_type_id', 'peril_id']
 
-        summary_cols += self.extra_summary_cols
+        summary_cols = list(dict.fromkeys(summary_cols + self.extra_summary_cols))
```
vinulw, line 248: minor: cant you just dedupe using a `set` and then convert to `list`

vinulw (approved):

oasislmf/computation/run/exposure.py
Ha-Ree: Needs ordering unfortunately

SkylordA (approved):

PR #2119, opened 2026-08-20
RI in exposure run propagation

### LINKED ISSUE #2118: exposure run: reinsurance losses only generated for the first loss factor
## Summary

`GenerateLossesDeterministic` only produces reinsurance losses for the *first* loss factor. With more than one `--loss-factor`, the returned RI frame contains rows for `loss_factor_idx == 0` only, so `exposure run` reports `total ri ceded` equal to `total il` (i.e. net RI of 0) for every loss factor after the first. GUL and IL are unaffected.

## To reproduce

```
oasislmf exposure run -s validation/reinsurance1 -l 0.5 1.0
```

```
Losses (loss factor=50.00%; total gul=14,500,000,000; total il=9,516,400,215; total ri ceded=585,380,006)
Losses (loss factor=100.00%; total gul=29,000,000,000; total il=19,196,400,190; total ri ceded=19,196,400,190)
```

`ri ceded` for the second factor is exactly `total il`, which only happens when net RI is 0. Inspecting the run directory confirms the RI frame is short:

```
guls  loss_factor_idx counts: {0: 28710, 1: 28710}
ils   loss_factor_idx counts: {0: 28710, 1: 28710}
rils  loss_factor_idx counts: {0: 28710}          <- no rows for loss factor 2
```

## Cause

The IL path converts the GUL CSV to binary keeping one sample per loss factor:

https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/computation/generate/losses.py#L883

```python
            csvtobin(guls_fp, guls_bin_fp, "gul", stream_type=self.il_stream_type, max_sample_index=len(self.loss_factor))
```

`run_ri_layer` then regenerates the same binary with `max_sample_index=1`:

https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/computation/generate/losses.py#L940

```python
                            csvtobin(guls_fp, guls_bin_fp, "gul", stream_type=self.il_stream_type, max_sample_index=1)
```

`max_sample_index` does double duty in `gul_tobin` — it filters rows to `sidx <= max_sample_index` *and* is written into the stream header as the sample count — so every loss factor beyond the first is dropped, and `rils['loss_factor_idx'] = rils['sidx'] - 1` is always 0.

### How the truncation reaches RI

The truncating `csvtobin` only rewrites `guls_bin_fp`, and nothing in the RI chain reads that file — RI layer 1 reads `ils_bin_fp`, layer *n* reads `ri{n-1}.bin`. The layer-1 `fmpy_run` immediately below it is what carries the damage across: it recomputes `ils_bin_fp` *from the just-truncated* `guls_bin_fp`, overwriting the full multi-sample `ils.bin` written by the IL step at L883-894. That overwritten one-sample `ils.bin` is what feeds RI.

```
L883    guls.csv --csvtobin(S=n)--> guls.bin (full)
L884    guls.bin --fmpy_run(IL)---> ils.bin  (full)
L895    ils.bin  --bintocsv-------> raw_ils.csv  ->  losses['il']   correct, banked here
--- RI section ---
L940    guls.csv --csvtobin(S=1)--> guls.bin  TRUNCATED
L942    guls.bin --fmpy_run(IL)---> ils.bin   overwritten, 1 sample   <- carrier
L957    ils.bin  --fmpy_run(RI)---> ri1.bin   only ever 1 loss factor
```

This is also why GUL and IL escape: `raw_ils.csv` is written and read into `losses['il']` at L895, before the RI section clobbers `ils.bin`.

Note the `fmpy_run` is guarded by `if layer == 1`, so for layers >= 2 only the `csvtobin` re-runs. There the truncation is harmless (layer 2+ reads `ri{n-1}.bin`) but wasted — the damage is inflicted once, at layer 1.

## Origin

Not a regression from the conversion-tools work. `gultobin -S 1` was correct when written in 2019 (1fb8d16b7), when the deterministic workflow had a single loss factor selected via special sidx values. #543 (63666c91f, 2020-04-17) added multi-loss-factor support and updated the IL call site from `-S 1` to `-S {len(self.loss_factor)}`, but left the RI path's hardcoded `-S 1` untouched. #1693 (4037be445) later translated it faithfully into `max_sample_index=1`. So this is a call site missed in 2020 that has been silently wrong since.

## Suggested fix

Correcting the argument to `len(self.loss_factor)` fixes it. But once it matches the IL step, both the `csvtobin` and the layer-1 `fmpy_run` in `run_ri_layer` become byte-level no-ops — same source, same args, and `guls_bin_fp` is untouched in between — so the better fix is to delete both and let RI layer 1 consume the `ils_bin_fp` that already exists. That also drops the per-layer re-conversion. Verified: every loss artifact (`guls/ils/rils.csv`, `guls.bin`, `ils.bin`, `raw_*.csv`, all `ri*.bin`/`ri*.csv`) is byte-identical between the two approaches across reinsurance1/2/4.

The deletion relies on the IL step at L871-895 being unconditional (only the RI block at L917 is guarded), so `guls.bin` and `ils.bin` always exist in full before RI runs.

Two things I flagged as risks when filing this, both since resolved:

- **Was `max_sample_index=1` load-bearing for the single-loss-factor case?** No. For one loss factor `len(self.loss_factor)` *is* 1, so the value is unchanged by construction; `guls.csv`, `ils.csv` and `rils.csv` are byte-identical to `main` for single-factor runs on reinsurance1 and reinsurance4.
- **Do the reinsurance validation fixtures need regenerating?** No. Only `validation/insurance_bi` and `validation/insurance_account` have a `loss_factors.csv`, and neither has RI files — no golden fixture anywhere combines RI with multiple loss factors. `tests/fm/test_fmpy.py` passes unchanged (21 passed).

That fixture gap is also why this survived five years, and it is the remaining hole: nothing tests multi-factor RI at all.

## Notes

Found while fixing #2109 (`total gul=0` in the same summary header). That issue is a separate defect in `RunExposure.run` and its fix does not address this one; the original description of #2109 wrongly claimed `total ri ceded` was unaffected.

Ha-Ree:
Fixes #2118.

`oasislmf exposure run` with more than one `--loss-factor` produced reinsurance losses for the first loss factor only, so `total ri ceded` was reported as the whole of `total il` (net RI of 0) for every factor after the first.

`validation/reinsurance1 -l 0.5 1.0`:

```
total ri ceded=585,380,006      (lf 50%,  unchanged)
total ri ceded=19,196,400,190  ->  1,156,380,008   (lf 100%, was == total il)

rils loss_factor_idx: {0: 28710}  ->  {0: 28710, 1: 28710}
```

## Cause

`run_ri_layer` regenerated `guls.bin` with `max_sample_index=1` where the IL step above it uses `max_sample_index=len(self.loss_factor)`. In `gul_tobin` that argument both filters rows to `sidx <= max_sample_index` and is written into the stream header as the sample count, so everything past the first loss factor was dropped.

The truncating `csvtobin` alone was harmless — nothing in the RI chain reads `guls.bin`, since layer 1 reads `ils.bin` and layer *n* reads `ri{n-1}.bin`. The layer-1 `fmpy_run` directly below it carried the damage across: it recomputed `ils.bin` *from the truncated* `guls.bin`, overwriting the full multi-sample `ils.bin` that the IL step had already written, and that one-sample `ils.bin` is what fed RI.

```
L883    guls.csv --csvtobin(S=n)--> guls.bin (full)
L884    guls.bin --fmpy_run(IL)---> ils.bin  (full)
L895    ils.bin  --bintocsv-------> raw_ils.csv  ->  losses['il']   correct, banked here
--- RI section ---
L940    guls.csv --csvtobin(S=1)--> guls.bin  TRUNCATED
L942    guls.bin --fmpy_run(IL)---> ils.bin   overwritten, 1 sample   <- carrier
L957    ils.bin  --fmpy_run(RI)---> ri1.bin   only ever 1 loss factor
```

That also explains why GUL and IL escaped: `raw_ils.csv` is written and read into `losses['il']` at L895, before the RI section clobbers `ils.bin`.

This is not a regression from the conversion-tools work. `gultobin -S 1` was correct in 2019 (`1fb8d16b7`) when the deterministic workflow had a single loss factor. #543 (`63666c91f`, 2020-04-17) added multi-loss-factor support and updated the IL call site to `-S {len(self.loss_factor)}` but missed the RI one; #1693 (`4037be445`) later translated it faithfully to `max_sample_index=1`. A call site missed in 2020.

## Change

Rather than correcting the argument, both the `csvtobin` and the layer-1 `fmpy_run` are deleted. Once `max_sample_index` matches the IL step they are byte-level no-ops — same source file, identical `fmpy_run` arguments, and `guls_bin_fp` untouched in between — so RI layer 1 can consume the `ils_bin_fp` that already exists. This also drops a re-conversion that ran once per RI layer.

The duplication is what caused the bug in the first place: two call sites performing the same GUL conversion, one updated for multiple loss factors and one missed. Correcting the argument would leave the second site there to drift again; deleting it removes the class of failure rather than the instance. It rests on the IL step at L871-895 being unconditional (only the RI block at L917 is guarded), so `guls.bin`/`ils.bin` always exist in full before RI runs — which upstream `ril` requiring `il` guarantees today.

## Verification

- Equivalence of the deletion vs simply correcting the argument: every loss artifact byte-identical across reinsurance1, reinsurance2 (multi-layer) and reinsurance4 — `guls/ils/rils.csv`, `guls.bin`, `ils.bin`, `raw_*.csv`, and all RI layer files (8/8, 22/22, 2/2).
- Single loss factor unchanged vs `main`: `len(self.loss_factor)` *is* 1 there, so identical by construction, and confirmed byte-identical `guls.csv`/`ils.csv`/`rils.csv` on reinsurance1 and reinsurance4.
- Multi-layer chaining through `ri1.bin` verified: reinsurance2, reinsurance3 and reinsurance4 all emit every loss factor.
- `tests/fm/test_fmpy.py`: 21 passed. `test_generate_losses.py` + `test_run_exposure.py`: 76 passed, with the same 2 pre-existing `test_generate_losses.py` failures present on `main`.
- Modest speedup from dropping the per-layer re-conversion: 21.8s -> 20.6s (reinsurance1), 55.1s -> 53.4s (reinsurance2).

## No fixture regeneration

I flagged this as a risk on the issue; it does not apply. Only `validation/insurance_bi` and `validation/insurance_account` carry a `loss_factors.csv`, and neither has RI files, so no golden fixture combines RI with multiple loss factors.

## Follow-up

That gap is why this went unnoticed for five years, and it is still open: there is no test covering multi-factor RI. Worth adding one asserting the RI frame carries every loss factor — not included here to keep this PR to the fix.

Chasing the byte comparison also turned up unrelated run-to-run nondeterminism in file preparation: `fm_summary_map.csv` emits its columns in varying order between runs (`layer_id,PolNumber` vs `PolNumber,layer_id`), as do `nodes_array_2.npy`, `ri_info.parquet` and `ri_layers.json`. A control run of identical code reproduced it, so it is not from this change, but it will make any byte-level fixture comparison flaky.


oasislmf/computation/generate/losses.py
```diff
@@ -937,19 +937,7 @@ def run_ri_layer(layer):
...
-                                    files_in=[guls_bin_fp],
-                                    files_out=[ils_bin_fp],
-                                    low_memory=self.fmpy_low_memory,
-                                    sort_output=self.fmpy_sort_output,
-                                    net_loss=None,
-                                    storage_method='sparse',
-                                )
                                 ri_input_fp = ils_bin_fp
                             else:
                                 ri_input_fp = os.path.join(output_dir, f'ri{layer - 1}.bin')
```
vinulw: minor, can move the if else clause out of the `try` now and restrict the scope of the try to the `fmpy_run`

vinulw: minor, can move the if else clause out of the `try` now and restrict the scope of the try to the `fmpy_run`. Also the exception message could be a bit clearer (e.g. specify this is in the reinsurance part of the run)

tests/computation/test_run_exposure.py
```diff
@@ -347,6 +347,22 @@ def test_all_files_output_matches_expected(self):
...
+            oed_accounts_csv=ACCOUNTS,
+            oed_info_csv=RI_INFO,
+            oed_scope_csv=RI_SCOPE,
+            loss_factor=[0.5, 1.0],
+        )
+        losses = pd.read_csv(out)
+        ri_by_factor = losses.groupby('loss_factor_idx')['loss_ri'].sum()
+
+        self.assertEqual(list(ri_by_factor[ri_by_factor > 0].index), [0, 1])
+        self.assertGreater(ri_by_factor[1], ri_by_factor[0])
```
vinulw: i think it would be worth having a general integration test that verifies that multiple loss factors work overall (not just for `ri`). You could use the same pattern as all the other tests and create a `expected_loss_factor_multiple_loss_factors.csv` in `test_exposure_run/` and run `_assert_output_matches` on it

vinulw (commented):
main comment = one on modifying the added test to just do a full integration test of the multiple `loss_factor`

SkylordA (approved):

vinulw (approved):
👍🏾

PR #2132, opened 2026-09-02
Fix fmpy back-allocation rule 1 for layers above layer 1

### LINKED ISSUE #2131: fmpy back-allocation rule 1 is wrong for every layer above layer 1
## Issue Description

fmpy back-allocation rule 1 (allocate in proportion to ground up loss) is wrong for any programme with more than one layer. Layer 1 is always correct; every layer above it is allocated using weights that belong to other nodes, so the back-allocated item losses do not sum back to that layer's loss.

On a two item, two layer test case the layer 2 items come to 1,312.50 against a layer loss of 1,000. On the [#2055](https://github.com/OasisLMF/OasisLMF/issues/2055) reproduction (one account, four single layer policies) fmpy `-a1` totals **148,407.46** where the correct total is **491,749.50**. `fmcalc -a1` gives the right answer in both cases. Rules 2 and 3 are unaffected and agree exactly between the two engines.

Alloc rule 1 is not the default (IL defaults to 2, RI to 3), so this should not be affecting normal runs — but it is silently wrong for anyone who sets `ktools_alloc_rule_il=1` on a multi layer account.

Two items in one aggregation with ground up losses of 1,000 and 3,000, one level, two layers limited to 3,000 and 1,000.

`fm_programme.csv`
```csv
from_agg_id,level_id,to_agg_id
1,1,1
2,1,1
```

`fm_policytc.csv`
```csv
layer_id,level_id,agg_id,profile_id
1,1,1,2
2,1,1,3
```

`fm_profile.csv`
```csv
profile_id,calcrule_id,deductible1,deductible2,deductible3,attachment1,limit1,share1,share2,share3
1,100,0,0,0,0,0,0,0,0
2,14,0,0,0,0,3000,0,0,0
3,14,0,0,0,0,1000,0,0,0
```

`fm_xref.csv`
```csv
output_id,agg_id,layer_id
1,1,1
2,1,2
3,2,1
4,2,2
```

`guls.csv`
```csv
event_id,item_id,sidx,loss
1,1,-3,1000
1,1,-2,0
1,1,-1,0
1,1,1,1000
1,2,-3,3000
1,2,-2,0
1,2,-1,0
1,2,1,3000
```

1. Convert the inputs and run both engines:

```
fmprogrammetobin < fm_programme.csv > fm_programme.bin
fmpolicytctobin  < fm_policytc.csv  > fm_policytc.bin
fmprofiletobin   < fm_profile.csv   > fm_profile.bin
fmxreftobin      < fm_xref.csv      > fm_xref.bin
gultobin -S1     < guls.csv         > guls.bin

fmcalc -p . -a1 < guls.bin | fmtocsv

python -m oasislmf.pytools.fm.cli -p . -a 1 --create-financial-structure-files
python -m oasislmf.pytools.fm.cli -p . -a 1 -i guls.bin -o fmpy.bin
fmtocsv < fmpy.bin
```

2. Compare the sample 1 rows. Layer 2's loss is `min(4000, 1000) = 1000`, allocated in proportion to the ground up losses of 1,000 and 3,000, so items 1 and 2 should get 250 and 750.

| output_id | item | layer | expected | fmcalc `-a1` | fmpy `-a1` |
|---|---|---|---|---|---|
| 1 | 1 | 1 | 750  | 750  | 750 |
| 2 | 1 | 2 | 250  | 250  | **562.50** |
| 3 | 2 | 1 | 2250 | 2250 | 2250 |
| 4 | 2 | 2 | 750  | 750  | 750 |

Layer 1 sums to 3,000 as it should. Layer 2 sums to 1,312.50 against a layer loss of 1,000.

## Root cause

`is_allocation_rule_a1` in [`oasislmf/pytools/fm/compute_sparse.py`](https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/pytools/fm/compute_sparse.py#L895-L938) makes two passes over the base children. The pass that builds the denominator reads the child's ground up loss with no profile offset:

```python
child_loss = loss_val[loss_indptr[child['net_loss']]: loss_indptr[child['net_loss']] + child_val_count]
temp_node_loss[:, child_sidx[val_i]] += child_loss[val_i]
```

The pass that applies the factor reads the same field **with** a profile offset:

```python
child_net = loss_val[loss_indptr[child['net_loss'] + profile_i]
    : loss_indptr[child['net_loss'] + profile_i] + child_val_count]
...
child_loss[val_i] = child_net[val_i] * temp_node_loss[profile_i, child_sidx[val_i]]
```

But `net_loss` is allocated **one slot per node**, not one per layer — [`financial_structure.py:702-704`](https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/pytools/fm/financial_structure.py#L702-L704):

```python
node['loss'], loss_i = loss_i, loss_i + node['layer_len']   # one per layer
if level == start_level:
    node['net_loss'], loss_i = loss_i, loss_i + 1           # one, total
```

So `child['net_loss'] + profile_i` is only in range for `profile_i == 0`. For any higher layer it walks past that node's single slot and into the next node's loss buffer. The numerator and denominator are then inconsistent, the weights no longer sum to 1, and the layer's loss is not conserved.

The numbers above fall straight out of the slot layout, which for this case is `[item1.loss_L1, item1.loss_L2, item1.net_loss, item2.loss_L1, item2.loss_L2, item2.net_loss, agg.loss_L1, ...]`. The layer 2 factor is `1000 / 4000 = 0.25`. Item 1 reads `net_loss + 1`, which is item 2's layer 1 output of 2,250, giving `2250 * 0.25 = 562.50`. Item 2 reads the aggregate node's layer 1 loss of 3,000, giving 750, which happens to be right.

## Suggested fix

Drop the offset so both passes read the one slot that exists:

```python
child_net = loss_val[loss_indptr[child['net_loss']]
    : loss_indptr[child['net_loss']] + child_val_count]
```

With that change fmpy returns 750 / 250 / 2250 / 750 on the case above, and 491,749.50 on the #2055 reproduction — matching fmcalc in both. I have not raised a PR; happy to if you'd like, with a multi layer rule 1 case added to the FM test suite.

Two related things worth a look while that code is open:

* the denominator is not guarded against being zero. `node_loss[val_i] / temp_node_loss[...]` is guarded on the numerator (`if node_loss[val_i]:`) but not the denominator, so a node with loss but no ground up loss would divide by zero under `fastmath`.
* nothing in `validation/` exercises a non default allocation rule, which is why this was never caught. Adding a rule 1 case would be worthwhile independently of the fix.

* OasisLMF `main` at 2.5.7, and long standing — the Feb 2026 rename in #1871 only changed the loop variable from `p` to `profile_i`, the offset was already there.
* Ubuntu 24.04, Python 3.12, ktools 3.12.4 for the fmcalc comparison.
* Only affects `--ktools-alloc-rule-il 1` (or `-a 1` on `oasislmf.pytools.fm.cli`) with more than one layer. Single layer programmes always take `profile_i == 0` and are correct.

## Example data / logs

The five CSVs above are the complete input. Found while investigating [#2055](https://github.com/OasisLMF/OasisLMF/issues/2055) — fmpy was being used as the reference to check an fmcalc fix ([ktools#412](https://github.com/OasisLMF/ktools/pull/412)) and rules 2 and 3 matched exactly, which is what made rule 1 stand out.

Ha-Ree:
Fixes #2131.

## The bug

`is_allocation_rule_a1` in `compute_sparse.py` makes two passes over the base children. The pass that builds the denominator reads each child's ground up loss with no profile offset, and the pass that applies the factor reads the same field **with** one:

```python
child_loss = loss_val[loss_indptr[child['net_loss']]: ...]                  # denominator
...
child_net  = loss_val[loss_indptr[child['net_loss'] + profile_i]: ...]      # numerator
```

`net_loss` is allocated one slot per node, not one per layer (`financial_structure.py`, `node['net_loss'], loss_i = loss_i, loss_i + 1`), so `+ profile_i` is only in range for `profile_i == 0`. For any layer above the first it reads past that node's single slot and into the next node's loss buffer, so the weights have nothing to do with that item's ground up loss and no longer sum to 1. Layer 1 was always correct; every layer above it was allocated with another node's losses and did not add back up to the layer's loss.

## The fix

Drop the offset so both passes read the one slot that exists. This is the last `net_loss` access in the module that carried an offset — `load_net_value` and the `keep_input_loss` write already used it unoffset.

## Results

Two items in one aggregation (ground up 1,000 and 3,000), two layers limited to 3,000 and 1,000. Layer 2's loss is 1,000 and should split 1:3:

| output_id | item | layer | expected | before | after |
|---|---|---|---|---|---|
| 1 | 1 | 1 | 750  | 750 | 750 |
| 2 | 1 | 2 | 250  | **562.50** | 250 |
| 3 | 2 | 1 | 2250 | 2250 | 2250 |
| 4 | 2 | 2 | 750  | 750 | 750 |

On the #2055 reproduction (one account, four single layer policies) fmpy `-a1` goes from **148,407.46** to **491,749.50**, against `fmcalc -a1` on 491,749.83.

Checked more broadly against ktools `fmcalc -a1` over 120 randomly generated multi level, multi layer structures (81 of them multi layer):

| | matching fmcalc | differing |
|---|---|---|
| before | 46 | **74** |
| after | **120** | 0 |

## Tests

New `tests/pytools/fm/test_back_allocation.py` runs a small multi layer structure end to end through `manager.run` for rules 1, 2 and 3, and asserts both the exact per item losses and the invariant that each layer's item losses add back up to that layer's loss. The two rule 1 tests fail on `main` and pass with this change; the rule 2 and 3 cases pass either way and are there as controls.

The static input is written as CSV in a `tmp_path`, which `load_static` accepts, so the test needs no ktools binaries. There was no end to end coverage of the fm calculation in `tests/pytools/fm` before this, and nothing anywhere exercises a non default allocation rule, which is why this went unnoticed.

## Not addressed

`temp_node_loss` is not guarded against being zero in the same branch — the division is guarded on the numerator (`if node_loss[val_i]:`) but not the denominator, so a node with loss but no ground up loss would divide by zero under `fastmath`. Left alone as it is a separate concern and I have no case that reaches it; noted in #2131.

### DIFF
--- a/tests/pytools/fm/test_back_allocation.py
+++ b/tests/pytools/fm/test_back_allocation.py
@@ -1,0 +1,118 @@
+"""End to end tests of the fm back-allocation rules on a small multi layer structure."""
+import struct
+
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+from oasislmf.pytools.common.event_stream import ITEM_STREAM, LOSS_STREAM_ID, stream_info_to_bytes
+from oasislmf.pytools.fm import manager
+
+# two items in one aggregation, one level, two layers limited to 3000 and 1000
+FM_PROGRAMME = """from_agg_id,level_id,to_agg_id
+1,1,1
+2,1,1
+"""
+
+FM_POLICYTC = """level_id,agg_id,layer_id,profile_id
+1,1,1,2
+1,1,2,3
+"""
+
+FM_PROFILE = """profile_id,calcrule_id,deductible1,deductible2,deductible3,attachment1,limit1,share1,share2,share3
+1,100,0,0,0,0,0,0,0,0
+2,14,0,0,0,0,3000,0,0,0
+3,14,0,0,0,0,1000,0,0,0
+"""
+
+# output ids in item then layer order
+FM_XREF = """output,agg_id,layer_id
+1,1,1
+2,1,2
+3,2,1
+4,2,2
+"""
+
+GULS = {1: 1000.0, 2: 3000.0}
+
+
+def write_static(path):
+    for name, content in [('fm_programme', FM_PROGRAMME), ('fm_policytc', FM_POLICYTC),
+                          ('fm_profile', FM_PROFILE), ('fm_xref', FM_XREF)]:
+        (path / f'{name}.csv').write_text(content)
+
+
+def write_gul_stream(path, guls, max_sidx=1):
+    """Write a loss stream holding one event with a single sample per item."""
+    with open(path, 'wb') as stream:
+        stream.write(stream_info_to_bytes(LOSS_STREAM_ID, ITEM_STREAM))
+        stream.write(np.int32(max_sidx).tobytes())
+        for item_id, loss in guls.items():
+            stream.write(struct.pack('=ii', 1, item_id))
+            for sidx in range(1, max_sidx + 1):
+                stream.write(struct.pack('=if', sidx, loss))
+            stream.write(struct.pack('=if', 0, 0.))  # delimiter
+    return path
+
+
+def read_loss_stream(path):
+    """Return {(output_id, sidx): loss} for the sampled sidx only."""
+    data = np.fromfile(path, dtype='b')
+    losses = {}
+    cursor = 8  # stream_type + max_sidx
+    while cursor < data.nbytes:
+        _, output_id = struct.unpack_from('=ii', data, cursor)
+        cursor += 8
+        while cursor < data.nbytes:
+            sidx, loss = struct.unpack_from('=if', data, cursor)
+            cursor += 8
+            if sidx == 0:
+                break
+            if sidx > 0:
+                losses[(output_id, sidx)] = loss
+    return losses
+
+
+def run_fm(tmp_path, allocation_rule):
+    write_static(tmp_path)
+    gul_path = write_gul_stream(tmp_path / 'guls.bin', GULS)
+    out_path = tmp_path / 'fm.bin'
+    manager.run(create_financial_structure_files=True, allocation_rule=allocation_rule,
+                static_path=str(tmp_path))
+    manager.run(create_financial_structure_files=False, allocation_rule=allocation_rule,
+                static_path=str(tmp_path), files_in=[str(gul_path)], files_out=[str(out_path)],
+                net_loss=None, storage_method='sparse', low_memory=False, sort_output=True,
+                stepped=None)
+    return read_loss_stream(out_path)
+
+
+@pytest.mark.parametrize('allocation_rule, expected', [
+    # layer 1 loss is min(4000, 3000) = 3000, layer 2 loss is min(4000, 1000) = 1000.
+    # rule 1 splits each layer by the ground up losses of 1000 and 3000, so both layers
+    # split 1:3. rules 2 and 3 have nothing below them here, so they agree.
+    (1, {1: 750., 2: 250., 3: 2250., 4: 750.}),
+    (2, {1: 750., 2: 250., 3: 2250., 4: 750.}),
+    (3, {1: 750., 2: 250., 3: 2250., 4: 750.}),
+])
+def test_back_allocation_is_per_layer(tmp_path, allocation_rule, expected):
+    losses = run_fm(tmp_path, allocation_rule)
+
+    assert_allclose([losses[(output_id, 1)] for output_id in sorted(expected)],
+                    [expected[output_id] for output_id in sorted(expected)], rtol=1e-6)
+
+
+@pytest.mark.parametrize('allocation_rule', [1, 2, 3])
+def test_back_allocated_losses_sum_to_the_layer_loss(tmp_path, allocation_rule):
+    """Each layer's item losses must add back up to that layer's loss.
+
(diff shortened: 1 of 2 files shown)


tests/pytools/fm/test_back_allocation.py
```diff
@@ -0,0 +1,118 @@
...
+GULS = {1: 1000.0, 2: 3000.0}
+
+
+def write_static(path):
+    for name, content in [('fm_programme', FM_PROGRAMME), ('fm_policytc', FM_POLICYTC),
+                          ('fm_profile', FM_PROFILE), ('fm_xref', FM_XREF)]:
+        (path / f'{name}.csv').write_text(content)
+
+
+def write_gul_stream(path, guls, max_sidx=1):
```
sstruzik: I would replace those custom write and load by the one we have in our code.
from oasislmf.pytools.converters.bintocsv.manager import bintocsv
from oasislmf.pytools.converters.csvtobin.manager import csvtobin
with a csv input like
GULS = """event_id,item_id,sidx,loss
1,1,1,1000.0
1,2,1,3000.0
"""

sstruzik (approved):

SkylordA (approved):

PR #2140, opened 2026-09-07
fix/csvtobin_pipe

### LINKED ISSUE #2139: csvtobin: default_tobin() crashes on non-seekable streams (breaks runner.py rerun() diagnostic fallback)
## Issue Description
`csvtobin/manager.py::default_tobin()` writes each chunk via `chunk.tofile(file_out)`. `numpy.ndarray.tofile()` requires a seekable destination (it calls `.tell()` internally), so it always fails with `OSError: obtaining file position failed` when `file_out` is a pipe. `file_type="eve"` has no entry in `TOBIN_FUNC_MAP`, so it always falls through to this broken `default_tobin`.

This is hit via `execution/runner.py::rerun()`, the automatic single-event crash-diagnosis fallback: when gulmc raises any exception mid-run, `computation/generate/losses.py` catches the resulting `CalledProcessError` and calls `rerun()`, which pipes a one-line event CSV through `csvtobin eve | gulmc ...` to replay it in isolation. Since that pipe hits the bug above, the diagnostic fallback itself crashes with this unrelated error, masking whatever originally failed.

Confirmed in both 2.5.5 and 2.5.7 (unchanged code path). No test exercises `runner.py::rerun()` or a non-seekable `csvtobin` destination, and PiWind's CI never triggers `rerun()` (needs a genuine mid-run failure), so this has effectively zero coverage.

**Fix**: `file_out.write(chunk.tobytes())` instead of `chunk.tofile(file_out)` — verified correct on a pipe (round-trips identical bytes, no seek required).

1. `r, w = os.pipe(); wf = os.fdopen(w, 'wb')`
2. `np.array([1], dtype=np.int32).tofile(wf)` → `OSError: obtaining file position failed`
3. In a real run: cause gulmc to raise mid-run (e.g. a downstream ktools stage closing its output FIFO early → `BrokenPipeError` in `gulmc/manager.py`'s `stream_out.write()`)
4. `losses.py` catches it, calls `rerun()`, which builds `printf 'event_id\n <id>\n' | csvtobin eve | gulmc ...` → crashes with the error above instead of a clean diagnostic

- OasisLMF 2.5.5 and 2.5.7 (both affected)
- Linux 6.8.0-136-generic, x86_64
- Python 3.12.12, numpy 2.4.3
- Not a regression — pre-existing, untested code path

## Example data / logs
```
  File ".../bin/csvtobin", line 7, in <module>
    sys.exit(main())
  File ".../oasislmf/pytools/converters/csvtobin/cli.py", line 66, in main
    csvtobin(file_in, file_out, file_type, **kwargs)
  File ".../oasislmf/pytools/converters/csvtobin/manager.py", line 60, in csvtobin
    tobin_func(stack, file_in, file_out, file_type, **kwargs)
  File ".../oasislmf/pytools/converters/csvtobin/manager.py", line 46, in default_tobin
    chunk.tofile(file_out)
OSError: obtaining file position failed
```

Original exception that triggered `rerun()` (from `gulmc`'s own log):
```
File ".../oasislmf/pytools/gulmc/manager.py", line 452, in run
    write_start += stream_out.write(memoryview(byte_mv[write_start: compute_info['cursor']]))
BrokenPipeError: [Errno 32] Broken pipe
```

Fix:
```diff
 def default_tobin(stack, file_in, file_out, file_type):
     dtype = TOOL_INFO[file_type]["dtype"]
     for chunk in iter_csv_as_ndarray(stack, file_in, dtype):
-        chunk.tofile(file_out)
+        file_out.write(chunk.tobytes())
```

SkylordA:
### fix/csvtobin_pipe
fixes and closes https://github.com/OasisLMF/OasisLMF/issues/2139

### DIFF
--- a/oasislmf/pytools/converters/csvtobin/manager.py
+++ b/oasislmf/pytools/converters/csvtobin/manager.py
@@ -43,7 +43,7 @@ TOBIN_FUNC_MAP = {
 def default_tobin(stack, file_in, file_out, file_type):
     dtype = TOOL_INFO[file_type]["dtype"]
     for chunk in iter_csv_as_ndarray(stack, file_in, dtype):
-        chunk.tofile(file_out)
+        file_out.write(chunk.tobytes())
 
 
 def csvtobin(file_in, file_out, file_type, **kwargs):
--- a/oasislmf/pytools/converters/csvtobin/utils/amplifications.py
+++ b/oasislmf/pytools/converters/csvtobin/utils/amplifications.py
@@ -25,8 +25,8 @@ def amplifications_write_bin(data, file_out, *, _write_header=True, _prev_item_i
             raise ValueError('Item IDs are not contiguous')
 
     if _write_header:
-        np.array([0], dtype="i4").tofile(file_out)
-    data.tofile(file_out)
+        file_out.write(np.array([0], dtype="i4").tobytes())
+    file_out.write(data.tobytes())
 
 
 def amplifications_tobin(stack, file_in, file_out, file_type):
--- a/oasislmf/pytools/converters/csvtobin/utils/complex_items.py
+++ b/oasislmf/pytools/converters/csvtobin/utils/complex_items.py
@@ -40,7 +40,7 @@ def complex_items_tobin(stack, file_in, file_out, file_type):
     try:
         items_df = pd.read_csv(file_in)
     except pd.errors.EmptyDataError:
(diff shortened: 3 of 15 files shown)


Ha-Ree:
Good change but does this need patching elsewhere (eg same file write still exists in parquettobin)?

Ha-Ree (approved):

PR #2141, opened 2026-09-08
fix/elt-plt-reset-state-buffer-full

### LINKED ISSUE #2147: eltpy/pltpy crash with spurious IndexError when MELT/QELT/MPLT/QPLT output buffer fills at a summary boundary
## Issue Description
`eltpy` (`oasislmf/pytools/elt/manager.py`) and `pltpy` (`oasislmf/pytools/plt/manager.py`) crash with a spurious `IndexError` when their fixed-size output buffer (`DEFAULT_BUFFER_SIZE = 1_000_000`) fills up exactly while finishing a summary's MELT/QELT (or MPLT/QPLT) record. This only happens when a single event's summaries alone push one of these outputs past 1,000,000 rows before the event naturally ends (≥500,000 non-zero-`impacted_exposure` summaries) — large-scale exposure with a wide event footprint, which is why it never shows up against PiWind or other small reference models. The crash cascades into a `BrokenPipeError` in `summarypy`/`gulmc`.

1. Run a model with `elt_moment: true` (or `plt_moment: true`) against exposure large enough that a single event's non-zero-`impacted_exposure` summaries cross ~500,000 before the event ends.
2. Observe `eltpy`'s/`pltpy`'s log: `IndexError: index <event_id-like large number> is out of bounds for axis 0 with size <sample_size>`.
3. `summarypy`/`gulmc` logs show a `BrokenPipeError` around the same time — fallout from the reader's crash.

- OasisLMF 2.5.7; affected code path unchanged since at least 2.5.5, so likely also affects earlier releases.
- Reproduced with `kernel_num_processes` at both `4` and `1`, and with `NUMBA_DISABLE_JIT=1` — ruling out a multiprocessing race or a JIT-only artifact.

## Example data / logs
Confirmed sequence from custom debug logging added to `read_buffer` to trace the crash (not standard output, abridged):
```
DBGTRACE read_buffer ENTRY: event_id_in=1000719084 cursor=0 valid_buff=65524 reading_losses=True summary_id=3818385
DBGTRACE read_event: YIELD event_id=1000719084 cursor_before_compact=44844 valid_buff=65524 will_compact=True
DBGTRACE read_buffer ENTRY: event_id_in=0 cursor=0 valid_buff=65536 reading_losses=True summary_id=3820941
DBGTRACE CRASH: event_id=0 summary_id=3820941 sidx=1000719084 loss=5.354288564664981e-39 cursor=8 valid_buff=65536
  bytes_around_cursor=ecc2a53b944d3a00f0372a4dfbffffff33e15949fcffffff0000803fffffffff33e1d94801000000
```

SkylordA:
### Fix eltpy/pltpy crash on MELT/QELT/MPLT/QPLT output-buffer overflow

`eltpy`/`pltpy` failed to reset `state["reading_losses"]` on the buffer-full early-return taken when the MELT/QELT (or MPLT/QPLT) output buffer fills right at a summary's terminator. This left the reader's state stuck mid-summary, so the next call misread the following summary's header bytes as sample data, crashing with a spurious `IndexError`.

This PR reworks the fix to reserve buffer capacity for a summary's worst-case output *before* reading its header, rather than checking after writing. This removes every capacity check from the terminator write path, so the missing-reset bug can't recur.

It also fixes an independent bug the above depends on: buffer-full and new-event returns now read the persisted `state["current_event_id"]` instead of the `event_id` parameter, since the caller always passes `event_id=0` on a fresh call. Previously this could write `EventId=0` to output after any buffer-full resume.

Adds regression tests covering a buffer overflow spanning multiple summaries, an overflow landing right before a genuine new event, and the SELT reservation margin.

closes https://github.com/OasisLMF/OasisLMF/issues/2147

### DIFF
--- a/oasislmf/pytools/elt/manager.py
+++ b/oasislmf/pytools/elt/manager.py
@@ -209,34 +209,32 @@ def read_buffer(
             sdloss = np.float64(0.0)
         return meanloss, sdloss
 
+    def _reservation_overflows(idx, reservation, capacity, name):
+        # Buffer genuinely too small for even one summary (idx == 0, i.e. buffer is
+        # already empty): flushing can never make room, so return would loop forever.
+        if idx + reservation > capacity:
+            if idx == 0:
+                raise ValueError(
+                    f"{name} reservation of {reservation} rows for a single summary exceeds the "
+                    f"output buffer capacity of {capacity}; increase OASIS_DEFAULT_BUFFER_SIZE."
+                )
+            return True
+        return False
+
     while cursor < valid_buff:
         if not state["reading_losses"]:
             # Reserve room for the next summary's worst-case output before reading
             # anything of it, so writes below can never run past the end of a buffer.
             # +2 (not +1): both MEAN_IDX and NUMBER_OF_AFFECTED_RISK_IDX can each add
             # one extra SELT row on top of the len_sample real samples.
-            if state["compute_selt"] and si + state["len_sample"] + 2 > selt_data.shape[0]:
-                if si == 0:
-                    raise ValueError(
-                        f"SELT reservation of {state['len_sample'] + 2} rows for a single summary exceeds the "
-                        f"output buffer capacity of {selt_data.shape[0]}; increase OASIS_DEFAULT_BUFFER_SIZE."
-                    )
-                _update_idxs()
-                return cursor, state["current_event_id"], item_id, 1
-            if state["compute_melt"] and mi + 2 > melt_data.shape[0]:
-                if mi == 0:
-                    raise ValueError(
-                        f"MELT reservation of 2 rows for a single summary exceeds the output buffer "
-                        f"capacity of {melt_data.shape[0]}; increase OASIS_DEFAULT_BUFFER_SIZE."
-                    )
-                _update_idxs()
-                return cursor, state["current_event_id"], item_id, 1
-            if state["compute_qelt"] and qi + len(intervals) > qelt_data.shape[0]:
-                if qi == 0:
-                    raise ValueError(
-                        f"QELT reservation of {len(intervals)} rows for a single summary exceeds the "
-                        f"output buffer capacity of {qelt_data.shape[0]}; increase OASIS_DEFAULT_BUFFER_SIZE."
-                    )
+            buffer_full = False
+            if state["compute_selt"] and _reservation_overflows(si, state["len_sample"] + 2, selt_data.shape[0], "SELT"):
+                buffer_full = True
+            if state["compute_melt"] and _reservation_overflows(mi, 2, melt_data.shape[0], "MELT"):
+                buffer_full = True
+            if state["compute_qelt"] and _reservation_overflows(qi, len(intervals), qelt_data.shape[0], "QELT"):
+                buffer_full = True
+            if buffer_full:
                 _update_idxs()
                 return cursor, state["current_event_id"], item_id, 1
 
--- a/oasislmf/pytools/elt/manager.py
+++ b/oasislmf/pytools/elt/manager.py
@@ -213,9 +213,8 @@ def read_buffer(
         if not state["reading_losses"]:
             # Reserve room for the next summary's worst-case output before reading
             # anything of it, so writes below can never run past the end of a buffer.
-            # +2 (not +1) since both MEAN_IDX and NUMBER_OF_AFFECTED_RISK_IDX can each
-            # add one extra SELT row on top of the len_sample real samples (see
-            # elt_selt_special_sidx_issue.md - a separate, pre-existing issue).
+            # +2 (not +1): both MEAN_IDX and NUMBER_OF_AFFECTED_RISK_IDX can each add
+            # one extra SELT row on top of the len_sample real samples.
             if state["compute_selt"] and si + state["len_sample"] + 2 > selt_data.shape[0]:
                 _update_idxs()
                 return cursor, state["current_event_id"], item_id, 1
--- a/oasislmf/pytools/elt/manager.py
+++ b/oasislmf/pytools/elt/manager.py
@@ -52,6 +52,7 @@ class ELTReader(EventReader):
             ('compute_selt', np.bool_),
             ('compute_melt', np.bool_),
             ('compute_qelt', np.bool_),
+            ('current_event_id', event_id_dtype),
             ('summary_id', summary_id_dtype),
             ('impacted_exposure', oasis_float),
             ('non_zero_samples', oasis_int),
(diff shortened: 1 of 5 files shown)


oasislmf/pytools/elt/manager.py
```diff
@@ -209,6 +211,20 @@ def _get_mean_and_sd_loss():
 
     while cursor < valid_buff:
         if not state["reading_losses"]:
+            # Reserve room for the next summary's worst-case output before reading
+            # anything of it, so writes below can never run past the end of a buffer.
+            # +2 (not +1): both MEAN_IDX and NUMBER_OF_AFFECTED_RISK_IDX can each add
+            # one extra SELT row on top of the len_sample real samples.
+            if state["compute_selt"] and si + state["len_sample"] + 2 > selt_data.shape[0]:
```
Ha-Ree: Claude has found:

If `max_records_per_event * (len_sample + 1) > splt_data.shape[0]`, it will do no work and cause an infinite loop

Ha-Ree (changes requested):

oasislmf/pytools/elt/manager.py
SkylordA: added an error with `DEFAULT_BUFFER_SIZE` being modifiable to make these `_data` arrays bigger if needed

oasislmf/pytools/elt/manager.py
```diff
@@ -216,12 +216,27 @@ def _get_mean_and_sd_loss():
             # +2 (not +1): both MEAN_IDX and NUMBER_OF_AFFECTED_RISK_IDX can each add
             # one extra SELT row on top of the len_sample real samples.
             if state["compute_selt"] and si + state["len_sample"] + 2 > selt_data.shape[0]:
+                if si == 0:
```
Ha-Ree: Can't help but feel like this is basically 3 copied if statements that could be simplified somehow to reduce the identical code
Maybe take it into a function so its just
```
if state["compute_selt"] and si + state["len_sample"] + 2 > selt_data.shape[0]:
    return _some_good_function_name_(si, "SELT", cursor, state, item_id)
if ...
    return _some_good_function_name(mi, "MELT", ...
```

Ha-Ree:
I am happy here with the changes but there are 2 minors Claude suggested that I will share

Minor — the migration to state is only half done: the local event_id used for all writes still comes from the parameter. elt/manager.py:169 / plt/manager.py:230, used at e.g. elt/manager.py:277, plt/manager.py:337. last_event_id and every return now read state["current_event_id"], but if a call enters with reading_losses == True and no header is read, writes still use the parameter — which read_streams passes as 0 (common/event_stream.py:354). Post-PR the only way to reach that is a truncated/malformed stream (the EOF path at event_stream.py:391-398 can return with reading_losses True, then the leftover loop calls read_buffer(..., 0, 0)), so impact is low — but seeding event_id = state["current_event_id"] at function entry closes the class completely instead of leaving one path open.

Minor — current_event_id lives on the reader, which is shared across all input streams. elt/manager.py:55, plt/manager.py:64; read_streams interleaves streams via the selector and eltpy -i accepts nargs='+' (elt/cli.py:13). With two input files, file A's last event id becomes last_event_id for file B's first header, firing a spurious new-event return that yields A's event id again with zero rows and skips one boundary detection. Output content stays correct (each row carries its own EventId); only flush granularity and the logged event id are affected. Worth a comment at minimum. Context: multi-file reading is already unreliable here — read_buffer resets read_summary_set_id = False on every file switch (elt/manager.py:100-104), so returning to a file whose summaryset id was already consumed eats 4 stream bytes. Pre-existing and out of scope, but it means the shared-state assumption isn't new.

SkylordA:
> I am happy here with the changes but there are 2 minors Claude suggested that I will share
> 
> Minor — the migration to state is only half done: the local event_id used for all writes still comes from the parameter. elt/manager.py:169 / plt/manager.py:230, used at e.g. elt/manager.py:277, plt/manager.py:337. last_event_id and every return now read state["current_event_id"], but if a call enters with reading_losses == True and no header is read, writes still use the parameter — which read_streams passes as 0 (common/event_stream.py:354). Post-PR the only way to reach that is a truncated/malformed stream (the EOF path at event_stream.py:391-398 can return with reading_losses True, then the leftover loop calls read_buffer(..., 0, 0)), so impact is low — but seeding event_id = state["current_event_id"] at function entry closes the class completely instead of leaving one path open.
> 
> Minor — current_event_id lives on the reader, which is shared across all input streams. elt/manager.py:55, plt/manager.py:64; read_streams interleaves streams via the selector and eltpy -i accepts nargs='+' (elt/cli.py:13). With two input files, file A's last event id becomes last_event_id for file B's first header, firing a spurious new-event return that yields A's event id again with zero rows and skips one boundary detection. Output content stays correct (each row carries its own EventId); only flush granularity and the logged event id are affected. Worth a comment at minimum. Context: multi-file reading is already unreliable here — read_buffer resets read_summary_set_id = False on every file switch (elt/manager.py:100-104), so returning to a file whose summaryset id was already consumed eats 4 stream bytes. Pre-existing and out of scope, but it means the shared-state assumption isn't new.

added a fix for these anyway

oasislmf/pytools/elt/manager.py
```diff
@@ -207,8 +206,35 @@ def _get_mean_and_sd_loss():
             sdloss = np.float64(0.0)
         return meanloss, sdloss
 
+    def _reservation_overflows(idx, reservation, capacity, name):
```
sstruzik, line 209: this function is repeated twice, can we move it into common, maybe in oasislmf/pytools/common/event_stream.py

tests/pytools/eltpy/test_eltpy.py
```diff
@@ -1,19 +1,56 @@
...
+import oasislmf.pytools.elt.manager as elt_manager
 from oasislmf.pytools.elt.manager import main
-from oasislmf.pytools.common.data import (oasis_int, oasis_float)
+from oasislmf.pytools.common.data import (oasis_int, oasis_float, quantile_interval_dtype)
+from oasislmf.utils.exceptions import OasisStreamException
 
 TESTS_ASSETS_DIR = Path(__file__).parent.parent.parent.joinpath("assets").joinpath("test_eltpy")
 
 
+def _build_summary_stream(sample_size, summaries):
```
sstruzik, line 22: i thing we should use csvtobin(file_type="summarycalc") to create the stream instead of this custom one

tests/pytools/pltpy/test_pltpy.py
```diff
@@ -1,17 +1,60 @@
...
+from oasislmf.pytools.common.id_index import build as id_index_build
+from oasislmf.pytools.common.input_files import OccurrenceCSR
+import oasislmf.pytools.plt.manager as plt_manager
 from oasislmf.pytools.plt.manager import main
+from oasislmf.utils.exceptions import OasisStreamException
 
 TESTS_ASSETS_DIR = Path(__file__).parent.parent.parent.joinpath("assets").joinpath("test_pltpy")
 
 
+def _build_summary_stream(sample_size, summaries):
```
sstruzik, line 22: same, can we use the csvtobin(file_type="summarycalc") ?

sstruzik (approved):

Ha-Ree (approved):

PR #2142, opened 2026-09-08
fix/rerun-stale-fifo-redirect

### LINKED ISSUE #2146: rerun() hangs indefinitely due to stale FIFO redirect in extracted gul command
## Issue Description
`oasislmf.execution.runner.rerun()` — used to replay a failing event outside numba (`NUMBA_DISABLE_JIT=1`) for a clean traceback after a kernel crash — hangs forever instead of producing one.

## Root cause
`rerun()` extracts `gul_cmd` verbatim from `run_kernel.sh`, which still contains the original run's output redirect (e.g. `> /tmp/<random>/<fifo>`). `-o {gul_output}` is then appended after this stale redirect rather than replacing it:
```python
gul_cmd = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('gul'))].pop(0)
gul_pipe = f"{eve_cmd} | {gul_cmd} -o {gul_output}"
```
The FIFO's original reader (e.g. `summarypy`) is long gone by the time `rerun()` runs, so opening it for writing blocks forever per POSIX FIFO semantics. `rerun()` never returns and has to be killed manually. Confirmed via `strace` and `/proc/<pid>/wchan`, which showed the process parked waiting for a FIFO reader that will never arrive.

1. Trigger a kernel crash that causes `event_error.json` to be written.
2. Call `rerun()` (or let the crash-handling path invoke it).
3. Observe the process hangs indefinitely, blocked in `open`/`openat` on a FIFO path, with no output ever written to `gul_errors.log`.

- OasisLMF 2.5.7 (code path unchanged since at least 2.5.5, so likely affects earlier releases too)
- Reproduces consistently, regardless of `kernel_num_processes` or `NUMBA_DISABLE_JIT`

## Example data / logs
Relevant code: `oasislmf/execution/runner.py`, `rerun()` — `gul_cmd` retains the original run's `> /tmp/<random>/<fifo>` redirect, so `gul_pipe` ends up with two output destinations: the stale FIFO (blocks forever) followed by `-o {gul_output}` (never reached).

## Proposed fix
Strip any trailing output redirect from `gul_cmd` before appending `rerun()`'s own output flag:
```python
gul_cmd = re.sub(r'>\s*\S+', '', gul_cmd).strip()
```
Verified: after the fix, `rerun()` completes and writes a real traceback to `gul_errors.log` instead of hanging.

SkylordA:
### fix/rerun-stale-fifo-redirect
rerun() reconstructs the gul command from run_kernel.sh to replay a single failing event without numba, but the extracted command still carries the original run's output redirect to a FIFO whose reader has already exited. Writing to that orphaned FIFO blocks forever. Strip any trailing output redirect before pointing the command at rerun()'s own output file instead.

closes https://github.com/OasisLMF/OasisLMF/issues/2146

### DIFF
--- a/oasislmf/execution/runner.py
+++ b/oasislmf/execution/runner.py
@@ -142,9 +142,9 @@ def rerun():
     gul_cmd = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('gul'))].pop(0)
     fm_cmds = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('fm'))]
 
-    # strip any stale output redirect from the extracted command (e.g. to a fifo whose
+    # strip a trailing output redirect from the extracted command (e.g. to a fifo whose
     # reader has already exited in the main run) before pointing it at our own output file
-    gul_cmd = re.sub(r'>\s*\S+', '', gul_cmd).strip()
+    gul_cmd = re.sub(r'\s*\d*>>?\s*\S+$', '', gul_cmd).strip()
 
     pipe_output = "/tmp/il_P1"
     summary_output = "/tmp/il_S1_summary_P1"
@@ -156,7 +156,7 @@ def rerun():
 
     fm_input = gul_output
     for i in range(len(fm_cmds)):
-        fm_cmd = re.sub(r"-\s*>\s*\S+", f"-o 64_ri{i + 1}.bin", fm_cmds[i])
+        fm_cmd = re.sub(r'\s*\d*>>?\s*\S+$', '', fm_cmds[i]).strip()
         fm_output = f"{event_error}_fm{i + 1}.bin"
         fm_pipe = f"{fm_cmd} -o {fm_output} -i {fm_input}"
         with open("fm_errors.log", "a") as error_log:
--- a/oasislmf/execution/runner.py
+++ b/oasislmf/execution/runner.py
@@ -142,6 +142,10 @@ def rerun():
     gul_cmd = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('gul'))].pop(0)
     fm_cmds = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('fm'))]
 
+    # strip any stale output redirect from the extracted command (e.g. to a fifo whose
+    # reader has already exited in the main run) before pointing it at our own output file
+    gul_cmd = re.sub(r'>\s*\S+', '', gul_cmd).strip()
+
     pipe_output = "/tmp/il_P1"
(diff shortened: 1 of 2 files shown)


oasislmf/execution/runner.py
Ha-Ree: Should probably strip this too?

oasislmf/execution/runner.py
```diff
@@ -142,6 +142,10 @@ def rerun():
     gul_cmd = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('gul'))].pop(0)
     fm_cmds = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('fm'))]
 
+    # strip any stale output redirect from the extracted command (e.g. to a fifo whose
+    # reader has already exited in the main run) before pointing it at our own output file
+    gul_cmd = re.sub(r'>\s*\S+', '', gul_cmd).strip()
```
Ha-Ree: Got Claude to look at the regex and it thought:

The substitution is global and unanchored, not "trailing" as the comment claims. oasislmf/execution/runner.py:145–147. `re.sub(r'>\s*\S+', '', gul_cmd)` rewrites every `>` in the segment and, for any non-simple redirect, converts the target into a stray positional argument rather than removing it:` 2>> log/gul_stderror.err` → `2 log/gul_stderror.err` (the match is `>>`, since `\S+` stops at the space), and 1> f → 1. Currently unreachable for the gul segment in generated scripts (all fixture shapes were checked; the only > is the single trailing FIFO redirect), so this is latent rather than live — but `re.sub(r'\s*\d*>>?\s*\S+$', '', gul_cmd)` (or splitting on the last unquoted >) matches the stated intent and is failure-proof if get_main_cmd_gul_stream ever gains a guard.

Ha-Ree (changes requested):

oasislmf/execution/runner.py
```diff
@@ -142,6 +142,10 @@ def rerun():
     gul_cmd = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('gul'))].pop(0)
     fm_cmds = [cmd.strip() for cmd in kernel_pipeline if cmd.strip().startswith(('fm'))]
 
+    # strip a trailing output redirect from the extracted command (e.g. to a fifo whose
+    # reader has already exited in the main run) before pointing it at our own output file
+    gul_cmd = re.sub(r'\s*\d*>>?\s*\S+$', '', gul_cmd).strip()
```
Ha-Ree: Could be nice to also lift the regex out so both come from a shared place  (line 159)

Ha-Ree (approved):

Issue #2143, opened 2026-09-08
Remove chance_of_loss (sidx -4): computed, carried through fmpy, then discarded


sstruzik:
## Summary

`chance_of_loss` (sidx `-4`) is computed by the ground-up tools, carried through the financial
module as a pass-through, and then **discarded**. Nothing in the pytools pipeline consumes it.
It should be removed from the computation and from the stream.

## Evidence that nothing reads it

Traced end to end:

| stage | what happens to `-4` |
|---|---|
| `gulpy` / `gulmc` | computed in `compute_mean_loss` (`gul/core.py:181`) and written |
| `plapy` | iterated over; deliberately **not** scaled since #<this PR> |
| `fmpy` | stored in `pass_through`, re-emitted unchanged (`fm/stream_sparse.py:382`) |
| `summarypy` | **dropped** — `summary/manager.py:219` accumulates only `sidx > 0` and `-1`, `-3`, `-5` |
| `aalpy` / `lecpy` / `pltpy` | read `-4` as the **affected-risk count**, which `summarypy` writes into the same slot |
| `eltpy` | does not reference it |

The `ChanceOfLoss` column in MELT/MPLT output does **not** come from this value — `eltpy` and
`pltpy` compute it themselves from the sample losses:

```python
chance_of_loss = state["non_zero_samples"] / state["len_sample"]   # elt/manager.py:242
state["chance_of_loss"] += (loss > 0) / state["len_sample"]        # plt/manager.py:416
```

So the special is dead data occupying a record per item, per event, in every ground-up and
financial-module stream.

## What removal would involve

- `gul/core.py` — drop `chance_of_loss` from `compute_mean_loss` and its return tuple
- `gul/manager.py`, `gulmc/manager.py` — drop the unpack and the `losses[CHANCE_OF_LOSS_IDX]` write
- `gul/common.py` — drop `-4` from `SPECIAL_SIDX`. **Careful**: `NUM_IDX = SPECIAL_SIDX.shape[0]`
  sizes the `losses` array, which is addressed by the negative constants, so `NUM_IDX` must stay
  at 5 or `MAX_LOSS_IDX == -5` starts pointing at the wrong row.
- `fm/` — remove the `pass_through` machinery entirely (~28 references across `manager.py`,
  `stream_sparse.py`, `compute_sparse.py`), which exists solely to carry this one value. This is
  the bulk of the win.

## Why this is an issue and not a PR

[...]

PR #2145, opened 2026-09-09
Building-packed loss streams: buildings in the sample dimension

### LINKED ISSUE #1870: Improve time and memory performance for big Portfolio
revisit the different part of the oasislmf execution process to improve memory and time performance.

## v2.5.1
- [x] rewrite insurance loss fm file generation to process level one at a time https://github.com/OasisLMF/OasisLMF/pull/1782
- [x] Optimize il_inputs and improve gul/il_inputs https://github.com/OasisLMF/OasisLMF/pull/1862
- [x] Refactor extract_financial_structure https://github.com/OasisLMF/OasisLMF/pull/1864
- [x] Look for optimisation in GULMC https://github.com/OasisLMF/OasisLMF/pull/1869
- [x] Look for optimisation in FMPY: no easy optimisation, only improve documentation. https://github.com/OasisLMF/OasisLMF/pull/1871

## v2.5.2
- [x] add monitoring functionality to track performance during the run phase https://github.com/OasisLMF/OasisLMF/pull/1898
- [x] create minimal warm-up for jit cache https://github.com/OasisLMF/OasisLMF/pull/1899
- [x] remove pandas from summarypy to save on memory usage https://github.com/OasisLMF/OasisLMF/pull/1903

## v2.5.3 (intermediary I/O + footprint)
- [x] dtype coherence between file generation and pytools (less upcasting / fewer copies) https://github.com/OasisLMF/OasisLMF/pull/1827
- [x] use parquet for intermediary data instead of CSV https://github.com/OasisLMF/OasisLMF/pull/1834
- [x] write binary files directly, bypass the CSV intermediary https://github.com/OasisLMF/OasisLMF/pull/1876
- [x] optimize FootprintParquetDynamic for partitioned parquet https://github.com/OasisLMF/OasisLMF/pull/1919
- [x] (related) reduce FootprintBin/FootprintBinZ memory & skip footprint load when data_server active https://github.com/OasisLMF/OasisLMF/pull/1956

## In progress
- [x] gulmc: replace numba dicts with precomputed array-backed structures — ~45% faster, ~30% less peak USS on PiWind 10K https://github.com/OasisLMF/OasisLMF/pull/1964
- [x] ODS_Tools (submodule): vectorize `check_conditional_requirement` — ~127x speedup on exposure validation (7.24s → 0.06s) https://github.com/OasisLMF/ODS_Tools/pull/266

sstruzik:
Multiplexes a location's buildings into the **sample dimension** of one item instead of expanding
one item row per building, so a model can treat buildings as independent risks without paying N
times the item count.

## What changes

`--disaggregation` replaces the boolean `--do-disaggregation`, which could only express two of the
three ways a location's `NumberOfBuildings` can be represented:

| value | representation |
|---|---|
| `none` | one item holding the whole location; buildings implicitly perfectly correlated |
| `items` | one item per building — the historical default, unchanged |
| `samples` | one item per (location, peril, coverage_type), buildings in the sample dimension |

`do_disaggregation` still works and maps onto the new values with a deprecation warning.

Under `samples`, building *b* sample *s* rides at `sidx = (b-1)*S + s`, and specials at
`-(b-1)*NUM_SPECIAL_SIDX`. **Building 1 is the identity encoding**, so any run that is not packed
is byte-for-byte unchanged.

The financial module applies the site levels' terms **per building** and collapses them once the
last risk-keyed level has been applied; back-allocation then works on the collapsed leaves.

## Format changes

- `correlations.bin` 20 → 24 bytes: one **signed** `packed_buildings`, magnitude = count, a
  negative sign meaning the buildings must reach the financial module separately. Older files are
  rejected with a clear message rather than mis-parsed.
- `fm_structure_info.bin`, new, 8 bytes: `site_collapse_level` and `max_buildings`. The financial
  module cannot derive either — fm levels are compacted, so the numbering varies per portfolio,
  and the building counts live on a table it does not read.
- `gul_summary_map.csv` / `fm_summary_map.csv` gain `number_of_buildings`. A map written before
  this change has no such column and is handled as one building per row, so existing run
  directories keep working.

## Measured

729,670-location flood portfolio carrying 5,751,561 buildings (up to 630,510 on one location),
20 samples, gulmc, 4 processes, against row disaggregation. Loss time fitted from runs at
5 / 50 / 150 events and extrapolated to the model's full 1,447-event set:

| | |
|---|---|
| items | 5,729,760 → 676,119 (8.5x fewer) |
| generated inputs | 6.7x smaller |
| generate-files wall | 195s → 133s (1.5x faster) |
| peak RSS | 2.5x lower (loss run) |
| end-to-end wall | 2.23x faster (campus), 1.07x (aggregate) |

Losses agree to 0.010% (GUL) and 0.098% (IL), which is sampling noise at 20 samples.

The size and memory gains are unconditional. The runtime gain is not: packing reorganises the
building x sample work rather than removing it, so it only shows where a location's buildings can
be summed before they reach the stream. Where they must stay apart for the financial module the
per-event cost is unchanged (1.81 vs 1.85 s/event) and what remains is the fixed saving from
loading smaller inputs.

## One path, not two

Packing started as a second path through both ground-up engines — its own RNG setup, its own copy
of the sampling loop, its own writer. Since the encoding is built so that **N == 1 reproduces the
unpacked case exactly**, that was one algorithm maintained twice. The engines now always take the
packed route, with N == 1 when nothing is packed.

That removed the duplicated `binary_search`/`get_gul`/threshold loop from gulpy, both branches from
gulmc, and left `write_losses`/`write_losses_packed` as a single writer. Net effect across the two
managers is ~165 lines deleted.

Verified byte-identical, not argued: `evepy | modelpy | gulpy` and `evepy | gulmc` over 3 allocation
rules x 3 generators x correlation (or effective-damageability) on/off, 18/18 identical each,
A/B'd against the pre-collapse code. The one output that *does* change is `gulmc --debug 1|2`,
which previously raised `ValueError` for S > 1 and overwrote the special rows for S == 1 — a latent
bug the merge fixes, now pinned by a test that fails 4/4 on the old code.

## Known limitations

- **Building-level summaries are not available.** Packing removes `building_id` from the item
  table, so a summary keyed on it cannot be produced.
- **The affected-risk count differs between modes.** A fully damaged 3-building location reports 1
  under `none`, 3 under `items`, 1 under `samples`. `samples` reproduces the `none` answer and
  `items` is the outlier; `IsAggregate` does not change any of them, because the count keys on
  `building_id` rather than `risk_id`. Which answer is correct is a modelling question, tracked in
  #2153 and documented on the disaggregation page. Losses are unaffected.
- **A custom ground-up binary must not be combined with `samples`.** `model_custom_gulcalc`
  bypasses `gulmc`/`gulpy`, and nothing currently stops the combination. It fails quietly:
  generation has already divided TIV by the building count, so a binary that does not implement
  the packed encoding writes ordinary sample indices and the losses come out N times too small.
  Documented with a warning admonition; a runtime guard is possible since `correlations.bin`
  carries the count at loss time.
- **`exposure run` does not pack.** The deterministic generator has no sample dimension, so
  `samples` runs as `none` there with an INFO line rather than failing — the same settings can be
  used to check a portfolio before launching the real run.

## Not blocking, but read before claiming parity

#2131 (`fmpy` back-allocation rule 1 wrong above layer 1) is pre-existing on `main` and will block
a packed-vs-disaggregated parity claim at allocation rule 1 until it is fixed. Rules 0, 2 and 3 are
unaffected.

#2143 was spun out of this work: `chance_of_loss` is computed, carried through `fmpy`, and then
discarded by `summarypy`. Not removed here because it invalidates 1,034 reference fixtures.

## Fixed along the way, all pre-existing on `main`

- **Deprecation notices never reached anyone.** Python shows a `DeprecationWarning` only when it is
  attributed to `__main__`, and attribution follows `stacklevel`, so anything raised below the
  entry point is dropped. `oasislmf version --verbose` printed no notice at all. A shared
  `warn_deprecated` now backs the CLI flag, both `lookup/builtin.py` parameters and the new one;
  the first three were invisible before and are verified visible now.
- **`redirect_logging`** left every `oasislmf.*` logger at the redirect level, so later code in the
  same process silently lost INFO logging. This was the real cause of 16 `test_platform` /
  `test_generate_losses` failures that only appeared in certain suite orderings.
- **`plapy`** multiplied `chance_of_loss` by the amplification factor, turning a probability of 0.8
  into 1.6.
- An **out-of-bounds read** in the cdf converter, introduced mid-branch and caught by review:
  `read_getmodel_stream` has two callers, and removing a bounds guard for one broke the other.
  `NUMBA_BOUNDSCHECK=1` raises on it; a subprocess test now guards that path, since without bounds
  checking the read is silent.

## Testing

3,262 tests pass; `ruff` and `autopep8` clean. Roughly 3,000 lines of new tests, including
packed-vs-disaggregated parity for every allocation rule, a deductible chosen to bind on one
building and not another (the only regime where per-building terms differ from location-total
terms), per-building Latin Hypercube stratification with a control proving the rejected
alternative loses it, and the percentage-of-TIV factors at both site and above-site levels.

The random-number scheme was reviewed empirically end to end: building 1 byte-identical to the
unpacked draw for all three generators and every `skip_seeds`; the Latin Hypercube property holding
for **every** building rather than just the first; per-building independence by Fisher z-scores
against N(0,1); no shared stream segments across 180,000 values; Philox counter distinctness across
2,000 buildings of one seed; and the correlation path achieving exactly the configured rho.

Four independent review passes were run over the branch. Everything they found is fixed here except
the items listed above, and several fixes carry a note in the commit message about the mutation
that proves the test catches the defect.

* Added `--disaggregation` (`none` / `items` / `samples`), replacing the deprecated
  `--do-disaggregation`. `samples` multiplexes a location's buildings into the sample dimension of
  a single item instead of creating one item per building, giving equivalent losses with far fewer
  items. Requires the pure-Python `gulmc`/`gulpy` pipeline.
* `correlations.bin` gained a per-item building count; files generated before this change are
  rejected with a message asking for regeneration rather than being mis-read.
* Fixed summary-info reporting a packed location's TIV as one building's share rather than the
  location total. Losses were never affected, only the reported exposure.
* Fixed deprecation warnings never reaching the user, including the `--verbose` notice, which had
  never been displayed.
* Fixed `plapy` applying the post-loss-amplification factor to chance-of-loss, which is a
  probability and could be pushed above 1.
* Fixed pytools logging redirection not restoring logger state, which silently suppressed INFO
  logging for anything running the tools in-process.


oasislmf/pytools/fm/compute_sparse.py
```diff
@@ -833,12 +1043,18 @@ def compute_event(compute_info,
...
+                            # back_alloc's one-base-child shortcut writes the post-profile loss straight to loss_in,
+                            # valid only when that child IS the storage node. The forced aggregation above breaks
+                            # that, so tell it which case this is.
+                            storage_is_base_child = (
+                                base_children_count == 1
+                                and nodes_array[temp_children_queue[0]]['node_id'] == storage_node['node_id'])
                             if is_allocation_rule_a2:
                                 ba_children_count = base_children_count
                             else:
                                 ba_children_count = 1
```
Ha-Ree, line 1055: Worth checking and updating tests to catch this: Claude found
The new` storage_is_base_child `guard on `back_alloc_a2`'s shortcut is computed from the real base-children count, but rules 0/1 force `ba_children_count = 1`. So whenever a node genuinely has >1 base children, the function sees `count == 1, storage_is_base_child == False`, skips `loss_in[:] = loss_out`, and rescales the first base child instead. On main that shortcut was unconditional at `count == 1.`

exposure run --kernel-alloc-rule-il 1 on validation/insurance changes 22 IL rows against main (output_id 1362: 2,644,768.75 → 675,061.12); insurance_conditions and reinsurance2 shift too. Setting `storage_is_base_child = True` inside the else: `ba_children_count` = 1 branch at both call sites (~1054 and ~1084) restores byte-identical output.` tests/fm/test_fmpy.py` can't catch it — `RunFmTest` only drives IL 2 / RI 3, so the branch's 21/21 pass is real but blind here.

validation/.~lock.insurance_conditions_test_case_list.xlsx#
```diff
@@ -0,0 +1 @@
+,sstruzik,maxwellflitton-Apollo,09.07.2026 14:05,file:///home/sstruzik/.config/libreoffice/4;
```
Ha-Ree, line 1: This looks out of place?

oasislmf/pytools/gul/manager.py
```diff
@@ -476,61 +523,138 @@ def compute_event_losses(event_id, coverages, coverage_ids, items_data,
...
+                # random samples for this building
+                for sample_idx in range(1, sample_size + 1):
+                    loss = building_losses[sample_idx - 1, item_j, b - 1]
+                    if loss >= loss_threshold:
+                        cursor = mv_write_sidx_loss(byte_mv, cursor, encode_sidx(b, sample_idx, sample_size), loss)
+        else:
+            # summed at source: an ordinary unpacked item covering all nb_item buildings
+            for special_idx in SPECIAL_SIDX:
+                value = losses[special_idx, item_j]
+                if special_idx != CHANCE_OF_LOSS_IDX:
```
Ha-Ree, line 647: Also need adjustment for STD_DEV_INDEX which scales at sqrt(nb_item)

oasislmf/pytools/pla/streams.py
```diff
@@ -51,6 +51,14 @@ def read_buffer(byte_mv, cursor, valid_buff, event_id, item_id, items_amps, plaf
...
+                # Chance-of-loss is a probability, not a loss -- amplifying it can push it above 1. Every other
+                # special scales with the loss, and tiv is scaled deliberately so an amplified loss is not
+                # clipped by the cap. Decoded rather than compared to -4 because a packed item carries one per
+                # building, at -4, -9, -14 ...; only negatives can be specials and the decode ignores the
+                # sample size for those.
+                if sidx < 0 and decode_local_sidx(sidx, 0) == CHANCE_OF_LOSS_IDX:
+                    continue
+
                 loss = sidx_loss_view[k]['loss']
                 loss = 0 if np.isnan(loss) else loss
```
Ha-Ree, line 63: Is it fine this nan catch is skipped?

Ha-Ree (changes requested):

oasislmf/pytools/fm/compute_sparse.py
sstruzik: claude is correct, fixed

validation/.~lock.insurance_conditions_test_case_list.xlsx#
sstruzik: Untracked in 5f1a6386e — it was swept in by a broad git add.

oasislmf/pytools/gul/manager.py
sstruzik: Fixed — scales by sqrt(n) now, with a test that fails on the old n scaling.

Ha-Ree (approved):

PR #2154, opened 2026-09-11
fix(cli): print help when oasislmf api get is called without options set

sambles:
### fix(cli): print help when oasislmf api get is called without options set
Minor fix for platform CLI, print help when missing required option.


SkylordA (approved):
pending missing coverage

Ha-Ree (approved):

PR #2155, opened 2026-09-14
feature/not-modelled-locations

### LINKED ISSUE #2116: Allow locations to be marked not-at-risk and not-modelled in the standard keys server
Oasis has a `notatrisk` status that can be assigned in the keys server, however in the standard keys server it can only be assigned via the `split_loc_perils_covered` builtin. In that situation, it is assigned to all rows where the peril is not included in the model. This is questionable because if a peril isn't modelled, we don't know whether it's at risk. A `notmodelled` status would be more appropriate.

Apart from this, the standard keys server doesn't allow model developers to explicitly set statuses. `success` and `failure` are assigned based on the presence or absence of `areaperil_id` or `vulnerability_id`.

Could a mechanism be provided so that model developers can mark locations as `notatrisk` or `notmodelled` based on their own criteria? Perhaps they might fill in a special `status` column in the lookup based on a spatial mask or table. This would allow the stats to be a lot more meaningful.

SkylordA:
### feature/not-modelled-locations
Adds a `notmodelled` keys status, distinct from `notatrisk`, for locations whose peril falls outside a model's scope. `build_split_loc_perils_covered` now assigns `notmodelled` instead of `notatrisk` in this case, and a new `build_set_status` builtin lookup step lets model developers set status/message explicitly from their own data.

closes https://github.com/OasisLMF/OasisLMF/issues/2116


oasislmf/lookup/builtin.py
```diff
@@ -606,12 +619,48 @@ def fct(locations):
...
+            if unknown_status.any():
+                raise OasisException(
+                    f"unknown status value(s) {sorted(locations.loc[unknown_status, status_column].unique())} "
+                    f"in column {status_column}, must be one of {sorted(valid_status_ids)}")
+
+            locations.loc[set_status, 'status'] = locations.loc[set_status, status_column]
+            if message_column is not None:
+                if message_column not in locations.columns:
+                    raise OasisException(f"missing column {message_column} in location")
+                locations.loc[set_status, 'message'] = locations.loc[set_status, message_column]
```
sstruzik: claude propose to add this guard, so empty cell resolving to nana in the message_column don't appear as 'NaN' in the message column
set_message = set_status & ~is_empty(locations, message_column)
locations.loc[set_message, 'message'] = locations.loc[set_message, message_column]

oasislmf/utils/status.py
```diff
@@ -20,7 +21,8 @@
     'fail_ap': {'id': OASIS_KEYS_FA, 'desc': 'Failure areaperil'},
     'fail_v': {'id': OASIS_KEYS_FV, 'desc': 'Failure vulnerability'},
     'notatrisk': {'id': OASIS_KEYS_NR, 'desc': 'Modelled but not at risk'},
-    'noreturn': {'id': OASIS_KEYS_XX, 'desc': 'No key returned from lookup'}
+    'noreturn': {'id': OASIS_KEYS_XX, 'desc': 'No key returned from lookup'},
+    'notmodelled': {'id': OASIS_KEYS_NMOD, 'desc': 'Not modelled'}
```
sstruzik, line 25: As this is a new feature, can you update the lookup doc with this new status.
And optionally if you have the energy (/token :) ), Refresh the entire lookup doc to be in line with the current code

sstruzik (changes requested):

PR #2157, opened 2026-09-14
Parallelise the ExposurePreAnalysis hook by account group

### LINKED ISSUE #2151: Pre-analysis hook constructs the user's class twice and prints kwargs to stdout
## Issue Description

`ExposurePreAnalysis.run` in [`oasislmf/computation/hooks/pre_analysis.py`](https://github.com/OasisLMF/OasisLMF/blob/main/oasislmf/computation/hooks/pre_analysis.py#L122-L127) has two debug `print` statements sitting immediately before the real call:

```python
print(kwargs)
print(_class(**kwargs))
_class_return = _class(**kwargs).run()
```

They are leftovers, not deliberate logging, and they cause two separate problems.

**1. The user's pre-analysis class is constructed twice.**

`print(_class(**kwargs))` builds a full instance purely to print its `repr`, and then the instance is thrown away and a second one is built for the actual `.run()`. Anything a model's `__init__` does happens twice — reading reference data, opening a connection, checking out a licence, logging, writing a scratch file. `exposure_data` is passed by reference, so a class that modifies the exposure in `__init__` rather than in `run()` applies its modification twice, and the extra application is invisible in the logs. Nothing in the documented contract says `__init__` has to be free of side effects, so this is the caller's bug, not the model's.

The printed value is just the default object `repr`, which carries no information:

```
<exposure_pre_analysis_simple.ExposurePreAnalysis object at 0x7a167bfe9fd0>
```

**2. `print(kwargs)` dumps the whole keyword dict to stdout.**

That is the `OedExposure` object, `input_dir`, `model_data_dir`, `user_data_dir`, `analysis_settings_json`, the full model `settings` dict and the parsed contents of `exposure_pre_analysis_setting_json`:

```
{'exposure_data': <ods_tools.oed.exposure.OedExposure object at 0x7a167c171010>, 'input_dir': '/tmp/tmp5gnm_wnp', 'model_data_dir': None, 'user_data_dir': None, 'analysis_settings_json': None, 'settings': {}, 'exposure_pre_analysis_setting': {'BuildingTIV_multiplyer': 2}}
```

It goes to stdout rather than through `self.logger`, so it is not tagged, not levelled and not suppressible by log configuration, and it lands in worker stdout on the platform. Every other status line in this method uses `self.logger.info`. Whatever a model puts in its pre-analysis settings JSON — credentials for a data service, for instance — is printed verbatim.

1. Run any model with a pre-analysis hook configured, e.g. `oasislmf model run` on a model with `exposure_pre_analysis_module` set, or the existing unit test with capture off:

```
python -m pytest tests/model_preparation/test_exposure_pre_analysis.py::test_exposure_pre_analysis_simple_example -s
```

2. Both lines above appear on stdout.

3. To see the double construction, point `exposure_pre_analysis_module` at a class that records each `__init__`:

```python
class ExposurePreAnalysis:
    def __init__(self, exposure_data, exposure_pre_analysis_setting, **kwargs):
        self.exposure_data = exposure_data
        self.exposure_pre_analysis_setting = exposure_pre_analysis_setting
        with open('/tmp/init_count.txt', 'a') as counter:
            counter.write('init\n')

    def run(self):
        self.exposure_data.location.dataframe['BuildingTIV'] *= self.exposure_pre_analysis_setting['BuildingTIV_multiplyer']
```

`/tmp/init_count.txt` contains two lines for one pre-analysis run.

## Suggested fix

Delete both lines. The `original_files` / `modified_files` `logger.info` calls either side of them already report what the hook did, so nothing is lost.

There is no test coverage that would have caught this — `tests/model_preparation/test_exposure_pre_analysis.py` only checks the resulting OED file, and never how many times the class was built or what reached stdout. Worth adding a case that asserts a single `__init__` and that the settings do not appear on stdout, so the prints cannot come back.

* OasisLMF `main` at 2.5.7, and every release since **1.27.0** — both lines were introduced in [#1171](https://github.com/OasisLMF/OasisLMF/pull/1171) (Jan 2023) and have never been touched since.
* Ubuntu 24.04, Python 3.12.
* Affects any run with `exposure_pre_analysis_module` set. Runs without a pre-analysis hook never reach this code.

## Example data / logs

The stdout above is from `tests/model_preparation/test_exposure_pre_analysis.py::test_exposure_pre_analysis_simple_example` run with `-s` on `main`.

### LINKED ISSUE #2156: Parallelise the ExposurePreAnalysis hook (currently single-process, unlike the keys/lookup step)
## Problem

The Oasis workflow already parallelises two of its slowest steps:

- the keys/lookup service splits the location file into chunks and runs them across a
  multiprocessing pool (`oasislmf/lookup/factory.py`, `lookup_num_chunks`/`lookup_num_processes`/
  `lookup_multiprocessing`), and
- the ktools loss calculation is chunked by event across separate OS processes
  (`oasislmf/execution/bash.py`).

`ExposurePreAnalysis` (`oasislmf/computation/hooks/pre_analysis.py`) - the model-supplier-defined
hook used for geocoding, disaggregation and other exposure enhancement, invoked before
`GenerateFiles`/the keys step in `RunModel`/`GenerateOasisFiles` - has no such chunking. It always
runs once, serially, in a single process on the whole location+account exposure
(`_class(**kwargs).run()` in `ExposurePreAnalysis.run()`). For large portfolios this hook is now
the long pole in the pipeline: it takes as long on a many-core machine as on a single core,
while the two neighbouring steps in the same pipeline already scale with available cores.

## Proposal

Chunk the pre-analysis step the same way the keys/lookup step is chunked, reusing its
"number of chunks / number of processes" configuration style so behaviour is consistent and
predictable across both steps:

- Split `exposure_data.location.dataframe` (and `exposure_data.account.dataframe`, when present)
  into N chunks and run the user's hook class once per chunk across a multiprocessing pool,
  merging the resulting location/account dataframes back together afterwards.
- Unlike the keys/lookup service, which is a pure per-location function and can safely split by
  individual `loc_id`, a pre-analysis hook is an arbitrary user-supplied plugin that may need to
  see all of a single account's locations together (e.g. account-level TIV/exposure
  adjustments). Chunks should therefore be formed from unique `(PortNumber, AccNumber)` groups,
  never splitting one account's rows across two chunks.
- Add new `exposure_pre_analysis_multiprocessing` / `exposure_pre_analysis_num_processes` /
  `exposure_pre_analysis_num_chunks` parameters, each defaulting to the corresponding
  `lookup_multiprocessing` / `lookup_num_processes` / `lookup_num_chunks` value already
  resolved for the run, so the two steps parallelise consistently out of the box while still
  allowing independent tuning (or disabling) per stage.
- Document the compatibility limits this implies for pre-analysis hooks: no cross-account state,
  no reliance on `exposure_data.ri_info`/`ri_scope` being chunked/merged (they are shared,
  read-only, across workers), and no shared-file side effects under `input_dir` that concurrent
  workers could race on. Models that need any of this can set
  `exposure_pre_analysis_multiprocessing=False`.

## Out of scope

- Chunking `ri_info`/`ri_scope` themselves, or supporting hooks that mutate them - reinsurance
  treaty structures aren't naturally partitioned by account the way location/account rows are.
- Distributing the pre-analysis step across multiple machines (only in-process multiprocessing,
  matching the existing keys/lookup step's model).

benhayes21:
## Summary

Closes #2156.

The `ExposurePreAnalysis` hook (`oasislmf/computation/hooks/pre_analysis.py`) always ran once,
serially, on the whole location+account exposure in a single process, unlike the keys/lookup
step and the ktools loss calculation, which are both already chunked/parallelised. For large
portfolios this makes the pre-analysis hook the long pole in the pipeline. This PR parallelises
it the same way the keys/lookup step is parallelised, reusing its "number of chunks / number of
processes" configuration style.

**Chunking strategy.** Unlike the keys/lookup step - a pure per-location function that can safely
split by individual `loc_id` (`oasislmf/lookup/factory.py`'s `location_producer`) - a pre-analysis
hook is an arbitrary user-supplied plugin that may need a whole account's locations together
(e.g. account-level exposure adjustments). So chunks are formed from unique `(PortNumber,
AccNumber)` groups instead, never splitting a single account's rows across two chunks. If the
location file lacks those columns (or `part_count` resolves to 1), the step falls back to the
existing single-process code path unchanged.

**New parameters**, added to `ExposurePreAnalysis.step_params`:

- `exposure_pre_analysis_multiprocessing`
- `exposure_pre_analysis_num_processes`
- `exposure_pre_analysis_num_chunks`

Each defaults to the resolved `lookup_multiprocessing`/`lookup_num_processes`/`lookup_num_chunks`
value for the run (see `ExposurePreAnalysis.run`'s `_resolve` helper), so the two steps
parallelise consistently out of the box while remaining independently tunable or disable-able
per stage.

**Implementation:**

- `oasislmf/computation/hooks/pre_analysis_multiproc.py` (new) - producer/worker/queue
  multiprocessing implementation, mirroring the fork-based pattern already proven in
  `oasislmf/lookup/factory.py` (same `with_error_queue` decorator, `billiard`-or-`multiprocessing`
  fallback, fork context, `tblib` traceback propagation across processes).
- `oasislmf/utils/parallel.py` (new) - `resolve_partition_count`, the partition-count/pool-size
  sizing math extracted out of `oasislmf.lookup.factory.KeyServerFactory.generate_key_files_multiproc`
  so both steps size chunks identically (no behaviour change to the lookup step).
- `oasislmf/computation/hooks/pre_analysis.py` - new params + dispatch logic in `run()`.
- `docs/source/explanation/pre-analysis-adjustments.rst` - documents the new params and the
  compatibility limits this implies for a pre-analysis hook (no cross-account state, no reliance
  on `ri_info`/`ri_scope` being chunked/merged, no shared-file side effects under `input_dir`).

**Side effect:** this also removes the leftover `print(kwargs)` / `print(_class(**kwargs))`
debug statements in `ExposurePreAnalysis.run()`, which constructed the user's hook class twice
per run and printed the full kwargs (including any secrets in `exposure_pre_analysis_setting`)
to stdout. Fixes #2151.

## Test plan

- [x] New chunking tests in `tests/model_preparation/test_exposure_pre_analysis.py`: an
      account-aware hook run with `exposure_pre_analysis_num_chunks` of 1/2/3 asserts (from
      inside the hook) that every chunk contains exactly one account's locations, and that the
      merged output matches the expected multiplied values in all cases; a separate test
      confirms `exposure_pre_analysis_multiprocessing=False` matches the same output.
- [x] Fixed a pre-existing test (`tests/computation/test_run_model.py::test_model_run__str2bool_called`)
      that asserted on the *last* call to a mocked `str2bool` - an ordering-fragile assertion
      that the new `exposure_pre_analysis_multiprocessing` param (also `str2bool`-typed)
      exposed; changed to `assert_any_call`, which preserves the test's actual intent.
- [x] `tests/model_preparation/test_exposure_pre_analysis.py`, `tests/model_preparation/test_lookup.py`,
      `tests/lookup/`, and `tests/computation/` (excluding two modules with an unrelated missing
      test dependency, `responses`) all pass: 224 passed.


benhayes21:
@sambles heads up for the release: this PR and #2060 both touch `ExposurePreAnalysis.run()` in `oasislmf/computation/hooks/pre_analysis.py` (#2060 adds `get_source_compression()`/`save_exposure_data()` around the two `exposure_data.save()` calls; this PR restructures the code between them to add multiprocessing). This branch was created before #2060 existed, so it doesn't yet account for those changes.

Whichever of the two you merge first, the other will need a rebase to reconcile the overlapping section of `run()` - they shouldn't be merged in parallel without one rebasing onto the other first. Flagging so you can sequence them for the release rather than merging both independently.

oasislmf/computation/hooks/pre_analysis_multiproc.py
```diff
@@ -0,0 +1,187 @@
...
+def exposure_producer(error_queue, loc_df, acc_df, part_count, group_cols, exposure_queue):
+    """Split loc_df/acc_df into part_count chunks and put them on exposure_queue.
+
+    If group_cols is given, chunks are formed from unique combinations of those columns
+    (e.g. ['PortNumber', 'AccNumber']) so that every location/account row belonging to the
+    same group ends up in the same chunk. Otherwise (no account grouping possible), chunks
+    are formed from unique 'loc_id' values, matching oasislmf.lookup.factory's location split.
+    """
+    if group_cols:
+        group_keys = loc_df[group_cols].drop_duplicates().reset_index(drop=True)
```
SkylordA: **Bug: accounts with no matching locations are silently dropped**

`group_keys` is derived only from `loc_df.drop_duplicates(group_cols)`, and `acc_df` is joined onto each chunk with an inner `merge`. Any account present in the account file but absent from the location file's groups (e.g. a genuinely locationless account, or one left over after filtering `location`/`location_numbers` without also filtering `account`/`account_numbers`) never lands in any chunk, and silently disappears from `exposure_data.account.dataframe` after `pd.concat`.

The identical input run with multiprocessing disabled keeps the account — this is a data-loss regression specific to the multiproc path, which is enabled by default.

Suggest deriving chunk keys from the union of `loc_df` and `acc_df` groups (or preserving unmatched account rows separately) rather than an inner join on location-only groups.

oasislmf/computation/hooks/pre_analysis_multiproc.py
```diff
@@ -0,0 +1,187 @@
...
+            break
+
+        chunk_exposure_data = copy.copy(exposure_data)
+        chunk_exposure_data.location = copy.copy(exposure_data.location)
+        chunk_exposure_data.location.dataframe = loc_part
+        if has_account:
+            chunk_exposure_data.account = copy.copy(exposure_data.account)
+            chunk_exposure_data.account.dataframe = acc_part
+
+        chunk_kwargs = dict(hook_kwargs)
```
SkylordA, line 70: **Bug: hook mutations to fields other than location/account are lost**

Only `chunk_exposure_data.location.dataframe` and `.account.dataframe` are shipped back through `result_queue` (lines 100-104) and merged back in `pre_analysis.py:163-165`. If a hook mutates any other `exposure_data` field — e.g. `ri_info`/`ri_scope` — that mutation happens only inside the forked worker and is never propagated to the parent.

This is a silent correctness break for any custom hook that touches reinsurance info/scope: the run completes with no error, but the saved exposure retains the original, unmodified `ri_info`/`ri_scope`. Worth at minimum a docs callout and ideally returning/merging the full chunked object, not just location/account.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -122,9 +130,42 @@ def run(self):
...
+                return own_value
+            kwargs_value = self.kwargs.get(lookup_key)
+            return kwargs_value if kwargs_value is not None else default
+
+        multiproc_enabled = _resolve(self.exposure_pre_analysis_multiprocessing, 'lookup_multiprocessing', True)
+        num_processes = _resolve(self.exposure_pre_analysis_num_processes, 'lookup_num_processes', -1)
+        num_chunks = _resolve(self.exposure_pre_analysis_num_chunks, 'lookup_num_chunks', -1)
+
+        group_cols = ['PortNumber', 'AccNumber']
+        can_group_by_account = all(col in exposure_data.location.dataframe.columns for col in group_cols)
```
SkylordA: **Bug: `can_group_by_account` doesn't validate the account dataframe**

This only checks `exposure_data.location.dataframe.columns` for `PortNumber`/`AccNumber`. If `exposure_data.account` exists but its dataframe lacks those columns (e.g. malformed input that survives because `check_oed=False`), `can_group_by_account` stays `True`, multiproc proceeds with `group_cols` set, and `exposure_producer`'s `acc_df.merge(key_part, on=group_cols)` raises a `KeyError`/`MergeError` — a new crash on input that runs fine single-process.

Suggest also requiring `all(col in exposure_data.account.dataframe.columns for col in group_cols)` when `exposure_data.account is not None`.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -122,9 +130,42 @@ def run(self):
...
+            multiproc_enabled = False
+        if can_group_by_account:
+            row_count = exposure_data.location.dataframe[group_cols].drop_duplicates().shape[0]
+        else:
+            row_count = exposure_data.location.dataframe.shape[0]
+        pool_count, part_count = resolve_partition_count(row_count, num_processes, num_chunks)
+
+        if multiproc_enabled and pool_count > 1:
+            self.logger.info(f'\nRunning pre-analysis across {pool_count} processes, {part_count} chunks')
+            location_df, account_df, _class_return = run_pre_analysis_multiproc(
```
SkylordA: **Correctness: `result["class"]` has an inconsistent shape**

On the single-process branch (line 168), `_class_return` is whatever the hook's `.run()` returns — a single object. On the multiproc branch, it's `class_returns`, a list of one return value per chunk. The switch is invisible to callers and depends purely on portfolio size (whether `resolve_partition_count` decides to chunk).

Any caller of the documented `OasisManager().exposure_pre_analysis(**kwargs)` entry point that inspects `result["class"]` will work on small portfolios and break (e.g. `AttributeError` from calling `.get()` on a list) once the same code runs against a large enough portfolio. Consider normalizing the shape (e.g. always return a list, or concatenate/aggregate) regardless of which path ran.

oasislmf/computation/hooks/pre_analysis_multiproc.py
```diff
@@ -0,0 +1,187 @@
...
+    """Split loc_df/acc_df into part_count chunks and put them on exposure_queue.
+
+    If group_cols is given, chunks are formed from unique combinations of those columns
+    (e.g. ['PortNumber', 'AccNumber']) so that every location/account row belonging to the
+    same group ends up in the same chunk. Otherwise (no account grouping possible), chunks
+    are formed from unique 'loc_id' values, matching oasislmf.lookup.factory's location split.
+    """
+    if group_cols:
+        group_keys = loc_df[group_cols].drop_duplicates().reset_index(drop=True)
+        key_parts = [group_keys.iloc[idx] for idx in np.array_split(np.arange(len(group_keys)), part_count)]
```
SkylordA, line 33: **Bug: empty chunks are dispatched to the hook when `num_chunks` > account count**

`np.array_split(np.arange(len(group_keys)), part_count)` yields empty index arrays once `part_count` exceeds the number of unique `(PortNumber, AccNumber)` groups, producing 0-row chunks that still get queued and run through `hook_cls(**chunk_kwargs).run()`.

Setting `exposure_pre_analysis_num_chunks` above the account count is a natural thing for a user to try ("force more parallelism") since it's a documented tunable — but a hook not defensively written for empty input (e.g. `.iloc[0]`, a geocoding call assuming ≥1 row) will crash. Not covered by the current tests (`num_chunks` is only tested ≤ account count). Suggest either skipping empty chunks before dispatch, or clamping `part_count` to the number of groups.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -122,9 +130,42 @@ def run(self):
...
+        num_chunks = _resolve(self.exposure_pre_analysis_num_chunks, 'lookup_num_chunks', -1)
+
+        group_cols = ['PortNumber', 'AccNumber']
+        can_group_by_account = all(col in exposure_data.location.dataframe.columns for col in group_cols)
+        if exposure_data.account is not None and not can_group_by_account:
+            # Without PortNumber/AccNumber on the location file, a location-only chunk split
+            # (below) could split a single account's rows across chunks - not safe to merge back.
+            multiproc_enabled = False
+        if can_group_by_account:
+            row_count = exposure_data.location.dataframe[group_cols].drop_duplicates().shape[0]
```
SkylordA: **Design issue: chunk sizing is based on account count, not location count — can disable multiprocessing for the main target case**

`row_count` here is the number of unique `(PortNumber, AccNumber)` pairs, not the number of location rows. Fed into `resolve_partition_count`'s `MIN_BLOC_SIZE=1000` floor, a portfolio with few accounts always collapses to `pool_count=1` regardless of how many location rows those accounts have.

Concretely: a portfolio with 5 accounts and 500,000 locations — exactly the "large portfolio, slow pre-analysis" case called out in this PR's own docs — yields `row_count=5`, so `bloc_size` floors at 1000, `part_count=1`, and multiprocessing silently never runs, with no warning to the user. Worth sizing off something that tracks actual work (e.g. location row count) rather than group count.

oasislmf/computation/hooks/pre_analysis_multiproc.py
```diff
@@ -0,0 +1,187 @@
...
+        else:
+            break
+
+        if res is None:
+            finished_workers += 1
+        else:
+            yield res
+
+
+def run_pre_analysis_multiproc(exposure_data, hook_cls, hook_kwargs, pool_count, part_count, group_cols):
```
SkylordA, line 98: **Nit: missing self-guard for `pool_count <= 1`**

`run_pre_analysis_multiproc` has no internal check for `pool_count <= 1` — it unconditionally spins up queues and processes. The guard only exists at the current call site (`pre_analysis.py:158`, `if multiproc_enabled and pool_count > 1`).

For comparison, `factory.py`'s `generate_key_files_multiproc` is self-defending (`if pool_count <= 1: return self.generate_key_files_singleproc(...)`). Any future caller of this function that forgets the check will spin up a full process pool for a single chunk. Suggest adding the same guard here.

oasislmf/computation/hooks/pre_analysis_multiproc.py
```diff
@@ -0,0 +1,187 @@
...
+
+from queue import Empty, Full
+
+# add pickling support for traceback object
+import tblib.pickling_support
+
+tblib.pickling_support.install()
+
+
+def with_error_queue(fct):
```
SkylordA: **Reuse: this duplicates ~150 lines of already-hardened machinery from `oasislmf/lookup/factory.py`**

`with_error_queue`, the put/get-with-retry poll loops, the `None`-sentinel relay protocol, the consumer generator, and the pool orchestration/teardown/error-reraise block are structurally identical to `factory.py`'s `location_producer`/`lookup_multiproc_worker`/`key_producer`/`generate_key_files_multiproc`. The PR correctly extracted `resolve_partition_count` into `utils/parallel.py` — the same treatment here (parameterizing the shared engine over `group_cols`/a per-chunk callable) would avoid two engines drifting independently. The missing `pool_count<=1` guard above is an early sign of that drift already starting.

Not necessarily a blocker, but would appreciate at least a follow-up issue tracked if not addressed now.

oasislmf/computation/hooks/pre_analysis.py
```diff
@@ -122,9 +130,42 @@ def run(self):
         self.logger.info('\nPre-analysis original files: {}'.format(
             json.dumps(original_files, indent=4)))
 
-        print(kwargs)
-        print(_class(**kwargs))
-        _class_return = _class(**kwargs).run()
+        def _resolve(own_value, lookup_key, default):
```
SkylordA: **Reuse: `_resolve()` reimplements the existing param-cascading mechanism**

`ComputationStep` already cascades defaults across chained steps via `chained_commands`/`get_params()` when a param name is reused (that's how `location`, `oasis_files_dir`, etc. already flow between `ExposurePreAnalysis`, `GenerateFiles`, `GenerateKeys`). This PR instead gives `ExposurePreAnalysis` differently-named params (`exposure_pre_analysis_multiprocessing` etc.) and bolts on `_resolve()` to manually fall back into `self.kwargs.get(lookup_key)`.

If `ExposurePreAnalysis` is ever invoked standalone (outside a `RunModel`/`GenerateOasisFiles` chain), `self.kwargs` never contains `lookup_multiprocessing`/etc., so this silently falls back to the hardcoded literals instead of genuinely inheriting a user's setting. It's also invisible in `--help`/the JSON schema since these aren't declared under the shared names in `step_params`. Declaring `lookup_multiprocessing`/`lookup_num_processes`/`lookup_num_chunks` directly in `step_params` would make `_resolve()` unnecessary.

SkylordA (changes requested):

benhayes21:
@SkylordA all 9 review comments should now be addressed - flagging so you can take a look and resolve the threads if you agree:

1. **Accounts with no matching locations silently dropped** - fixed: chunk keys are now derived from the union of `loc_df` and `acc_df` groups (see `_split_chunks` in `pre_analysis_multiproc.py`), not location groups alone, so an account-only group still gets its own chunk.
2. **Hook mutations to ri_info/ri_scope lost** - fixed: `_make_chunk_processor` now snapshots `ri_info`/`ri_scope` before running the hook and raises an `OasisException` immediately if a chunk mutates either, instead of silently discarding the change.
3. **`can_group_by_account` doesn't validate the account dataframe** - fixed: it now also requires `all(col in exposure_data.account.dataframe.columns for col in group_cols)` when `exposure_data.account is not None`.
4. **Empty chunks dispatched when `num_chunks` > group count** - fixed in two places: `resolve_partition_count` now clamps `part_count` to `row_count` generally (this also closes the same gap in the keys/lookup multiproc path, which shares this function), and `pre_analysis.py` additionally clamps to the actual `(PortNumber, AccNumber)` group count before dispatch.
5. **`result["class"]` inconsistent shape** - fixed: it's now always a list, regardless of whether multiprocessing ran.
6. **Chunk sizing based on account count, not location count** - fixed: partition sizing now uses the location row count (the real per-hook workload), so the "5 accounts, 500k locations" case from your example now chunks correctly.
7. **Missing `pool_count <= 1` self-guard** - fixed: the shared `run_multiproc` engine (see #8) now raises if called with `pool_count <= 1`, so every caller gets this guard for free.
8. **~150 duplicated lines from `factory.py`** - dealt with directly rather than just tracked: extracted a shared `oasislmf.utils.multiproc` engine (`chunk_producer`/`multiproc_worker`/`result_producer`/`run_multiproc`), parameterized over a per-worker `make_process_chunk(worker_id)` factory (so `factory.py` can still build a distinct lookup instance per worker) and an `on_results` callback. Both `pre_analysis_multiproc.py` and `lookup/factory.py` now only hold their own domain logic on top of it. Net line count went down despite adding new tests for the engine itself.
9. **`_resolve()` reimplements param-cascading** - fixed: removed the ad-hoc `exposure_pre_analysis_multiprocessing`/`num_processes`/`num_chunks` params and `_resolve()` entirely; `ExposurePreAnalysis` now declares `lookup_multiprocessing`/`lookup_num_processes`/`lookup_num_chunks` directly in `step_params` (matching `GenerateKeys`), so they cascade through `ComputationStep`'s existing mechanism whether chained or run standalone.

Also worth noting: the keys/lookup multiproc path (`generate_key_files_multiproc`) previously had no test coverage of its actual multiprocess execution (existing tests only asserted on mocked call args) - added a real end-to-end test for that too while extracting the shared engine.

Full test suite (engine unit tests, pre-analysis unit + integration tests, keys/lookup integration test, plus the broader `RunModel`/`GenerateFiles`/`RunExposure`/`GenerateKeys` regression suites) all green; `ruff`/`autopep8` clean.

oasislmf/computation/hooks/pre_analysis_multiproc.py
SkylordA: @benhayes21 

Thanks for the fix — the guard correctly catches a mutated `ri_info`/`ri_scope` when one already existed before the hook ran. Two gaps remain though:

1. The check is gated on `if unchunked_ri_info is not None` (`pre_analysis_multiproc.py:74`). If `ri_info`/`ri_scope` starts as `None` and the hook *creates* one under multiprocessing, that change is still silently discarded — same failure mode as the original bug, just narrower (only triggers when a hook adds RI data to an exposure that didn't have any). Worth checking both directions, e.g. comparing "was it None before vs. after" as well as the equality check.
2. If a hook sets `ri_info`/`ri_scope` to `None`, line 74 raises a bare `AttributeError: 'NoneType' object has no attribute 'dataframe'` rather than the intended `OasisException`. Should be handled explicitly.

Also, the docs (`pre-analysis-adjustments.rst:66-68`) currently say the run "raises an error as soon as a chunk is found to have changed either one" — that's not quite accurate given gap 1, worth softening or fixing before merge.

Leaving this thread open until the None-to-populated case is covered.

SkylordA (commented):
see existing comment thread

PR #2160, opened 2026-09-15
fix(losses): pre-build GUL structures for the engine that actually runs

### LINKED ISSUE #2161: Blended vulnerabilities: "Vulnerability_ids ... are missing from vulnerability.bin" when the getmodel/gulpy structures are built for a gulmc run
## Issue Description

Loss generation fails with

```
Exception: Vulnerability_ids {np.int32(...), np.int32(...)} are missing from file:///.../static/vulnerability.bin
```

on models that use vulnerability blending (models shipping `aggregate_vulnerability` + `weights`),
**including runs configured with `"gulmc": true`**.

Two things make the message hard to place:

1. It appears immediately after the `prepare_run_inputs` log line, but that function
   (`oasislmf/execution/bin.py`) performs no vulnerability validation at all — it is simply the
   step logged before the crash. The exception is raised by `get_vulns()` in
   `oasislmf/pytools/getmodel/manager.py`, reached from the shared-structure pre-build that runs
   a few lines later in `GenerateLossesDir.run()` (`oasislmf/computation/generate/losses.py`).
2. The ids listed are the aggregate/blended vulnerability ids. They are deliberately never
   present in `vulnerability.bin` — only their component ids are.

There are two structure builders, and only one of them understands blending:

* `oasislmf/pytools/gulmc/structure.py` reads `aggregate_vulnerability` and `weights` and expands
  each aggregate id into its component vulnerability ids before resolving them against
  `vulnerability.bin`. The components do exist, so it succeeds; weights are reapplied during
  sampling.
* `oasislmf/pytools/getmodel/structure.py` (the `modelpy | gulpy` path) takes each item's
  `vulnerability_id` as-is and has no knowledge of the aggregate files, so the lookup fails and
  this exception is raised.

There are two ways to end up in the second builder:

**(a) `gulmc` explicitly disabled.** Blending is a gulmc-only feature and the plain
`modelpy | gulpy` pipeline was never expected to support it, but it fails with this confusing
message rather than a clear one.

**(b) `gulmc` enabled, but not visible to the step that pre-builds the structures.**
`GenerateLossesDir` selects which structures to build from `getattr(self, 'gulmc', False)`, and
`gulmc` is not among its declared `step_params`. Undeclared params are silently dropped by
`ComputationStep`, so whenever that step is invoked on its own — as the chunked/distributed loss
workflow does (`generate_losses_dir`, then `generate_losses_partial` per chunk) — `gulmc`
resolves to `False` no matter how the run is configured, and the getmodel/gulpy structures are
pre-built. On a blended model that raises during preparation, before any chunk runs. Single-script
runs (`oasislmf model run`, `oasislmf model generate-losses`) are unaffected, because there the
same step runs as part of a class that does declare `gulmc`.

The same missing declarations drop `model_custom_gulcalc`, `peril_filter`, `dynamic_footprint`
and the df-engine options in that step, so structures are also pre-built unfiltered when
`--peril-filter` is used, and pre-built at all when a custom gulcalc binary is configured.

1. Take a model with blended vulnerabilities. `tests/assets/test_model_2` in this repo is one:
   `input/items.csv` references aggregate ids `100001` / `100003`, while `static/vulnerability.csv`
   contains only the component ids.
2. Build each structure set against that run directory:

```python
from oasislmf.pytools.gulmc.structure import create_gulmc_structure
from oasislmf.pytools.getmodel.structure import create_getmodel_structure

run_dir = "<copy of tests/assets/test_model_2>"
kwargs = dict(run_dir=run_dir, ignore_file_type=set(), peril_filter=[])

create_gulmc_structure(**kwargs, dynamic_footprint=False)   # OK
create_getmodel_structure(**kwargs)                         # raises
```

3. The second call raises the reported exception:

```
Exception: Vulnerability_ids {np.int32(100001), np.int32(100003)} are missing from
file:///.../repro/static/vulnerability.bin
```

The end-to-end form is a chunked run on a blended model with `"gulmc": true`: the
`generate_losses_dir` step fails with the same exception while preparing the run directory.

* Affected: **2.5.4, 2.5.5, 2.5.6, 2.5.7** (latest). Introduced by #1964, first released in 2.5.4.
* Not affected: 2.5.3 and earlier, and the 2.4.x / 2.3.x lines — they have no shared-structure
  pre-build step.
* Path (b) requires the chunked/distributed loss workflow, which calls `generate_losses_dir` as a
  standalone step. Path (a) reproduces anywhere.

## Example data / logs

```
... prepare_run_inputs ...
Exception: Vulnerability_ids {np.int32(100001), np.int32(100003)} are missing from
file:///.../static/vulnerability.bin
```

sstruzik:
## Problem

`GenerateLossesDir` decides which read-only `.npy` structures to pre-build for the run
(`gulmc_structure` vs `getmodel_structure` + `gulpy_structure`), but it never declared the
params that decision is made from — `gulmc`, `model_custom_gulcalc`, `peril_filter`,
`dynamic_footprint`, `model_df_engine`/`base_df_engine`. They were only declared on the
sub-steps that write the run script (`GenerateLossesPartial`, `GenerateLosses`), and read back
in the dir step through `getattr(self, 'gulmc', False)`, so an undeclared param silently
resolved to its `getattr` default.

That is harmless when the dir step runs as part of one of those sub-steps (the MDK's
`generate-losses` / `oasislmf model run`), because `self` is then an instance that does declare
them. It is not harmless in the **chunked workflow**, where the dir step is called on its own:

```python
# OasisPlatform src/model_execution_worker/distributed_tasks.py
params['analysis_settings'] = OasisManager().generate_losses_dir(**params)   # gulmc dropped here
...
OasisManager().generate_losses_partial(**chunk_params)                       # gulmc honoured here
```

`params` does carry `gulmc: True` (it comes from `_params_generate_oasis_losses`), but
`generate_losses_dir` discards it, takes the `not gulmc` branch and pre-builds
**getmodel + gulpy** structures. It then writes `output/analysis_settings.json`, which makes
each chunk's `_is_run_settings_stored` guard skip the dir step entirely — so the gulmc
structures are never built centrally and every chunk rebuilds them in-process, while the
unused getmodel/gulpy `.npy` files are tarred into `run_directory.tar.gz` and shipped to every
chunk worker.

Two further consequences of the same missing declarations:

* `model_custom_gulcalc` — with a custom GUL binary the dir step still built getmodel + gulpy
  structures that nothing reads.
* `peril_filter` — the MDK `--peril-filter` option was dropped, so structures were built
  unfiltered. Since `gulmc`/`getmodel`/`gulpy` all prefer a pre-built structure over their own
  runtime `peril_filter` argument, a chunked run with `--peril-filter` **and** `--gulmc False`
  silently ignored the filter. (With the default `gulmc True` the numbers were unaffected,
  because the mis-built structures were never loaded.)

## Fix

Declare the six params on `GenerateLossesDir`, where they are used, and read them as plain
attributes. The duplicate copies on `GenerateLossesPartial` / `GenerateLosses` are removed —
`get_params()` keeps the first definition of a name, so they were already dead and had started
to drift (two different help strings for `model_df_engine`).

Verified no change to the existing steps' resolved params: `GenerateLosses` and
`GenerateLossesPartial` are byte-identical before/after; `GenerateLossesDir` and
`GenerateLossesOutput` gain the six with the same defaults.

## Test

`test_losses__chunked_workflow__structures_built_for_the_engine_that_runs` calls
`generate_losses_dir` the way the chunked workflow does and asserts, through the same
`*_structure_exists` helpers the engines use at runtime, that the structures built match the
engine selected. It fails on `main` for the `gulmc=True` (default) and `model_custom_gulcalc`
cases and passes with this change.

## Affected releases

Introduced in #1964 (`1238449a5`, 2026-05-29), first shipped in **2.5.4**. Present in
**2.5.4, 2.5.5, 2.5.6 and 2.5.7** (latest). Clean in 2.5.3 and earlier, and in the 2.4.x /
2.3.x lines. OasisPlatform 2.5.4 and 2.5.5 pin affected oasislmf versions.

## Not in this PR

Two adjacent findings from the same review, left out to keep this focused:

* `get_getmodel_cmd` tests `if gulmc is True` (`execution/bash.py`) while the structure
  selection tests truthiness — a truthy non-`True` value yields `modelpy | gulmc`. The MDK path
  is protected by `str2bool`; direct `genbash`/`bash_params` callers are not.
* The complex-model gulcalc inference (`{model_supplier_id}_{model_name_id}_gulcalc` on PATH)
  exists only in `runner.run`, so the single-script flow honours it and the chunked flow does
  not — the same model config runs a different GUL engine in each.

* Fix the chunked loss workflow pre-building the wrong GUL shared structures: `generate-losses-dir`
  now honours `--gulmc`, `--model-custom-gulcalc`, `--peril-filter`, `--dynamic-footprint` and the
  df-engine options instead of silently falling back to their defaults.


oasislmf/computation/generate/losses.py
```diff
@@ -206,6 +206,19 @@ class GenerateLossesDir(GenerateLossesBase):
...
 
+        # GUL engine selection -- these decide which shared structures are pre-built here, so they
+        # must be declared on this step and not only on the sub-steps that generate the run script.
+        # The chunked workflow calls this step on its own (`generate_losses_dir`), and an undeclared
+        # param is silently dropped, which would pre-build the structures of the engine that is not
+        # going to run.
+        {'name': 'gulmc', 'default': True, 'type': str2bool, 'const': True, 'nargs': '?', 'help': 'use full Monte Carlo gulcalc python version'},
+        {'name': 'model_custom_gulcalc', 'default': None, 'help': 'Custom gulcalc binary name to call in the model losses step'},
+        {'name': 'peril_filter', 'default': [], 'nargs': '+', 'help': 'Peril specific run'},
+        {'name': 'dynamic_footprint', 'default': False, 'help': 'Dynamic Footprint'},
```
Ha-Ree: 'type': str2bool

Ha-Ree (changes requested):

oasislmf/computation/generate/losses.py
sstruzik: Done — deb3141b. Without a type the value was never coerced, so `--dynamic-footprint False` stayed a truthy string and turned the feature on; added `'const': True, 'nargs': '?'` too so it matches the other boolean params here.

Ha-Ree (approved):

SkylordA (approved):

PR #2163, opened 2026-09-16
Check and wait for pytools logging to complete

sambles:
#### Problem

  Chunked loss generation (`generate_losses_chunk`/`generate_losses_output` in the platform worker) would occasionally archive and merge pytool log files (`gulmc_*.log`, `summarypy_*.log`, `eltpy_*.log`, etc.) that were missing trailing lines compared to what eventually ended up on disk. This surfaced downstream as occasionally archive and merge pytool log files (`gulmc_*.log`, `summarypy_*.log`, `eltpy_*.log`, etc.) that were missing trailing lines compared to what eventually ended up on disk. This surfaced downstream as `check_complete()` reporting lost processes (e.g. `summarypy - 1 processes lost`) at the final output/merge stage, well after the actual truncation occurred.

#### Root causes

1. **Archiving raced against still-writing processes.** `run_analysis()` (`oasislmf/execution/runner.py`) waited only for the bash script's tracked child PID to exit (`proc.communicate()`). A pytool's internal worker/subprocess can be reparented and keep writing to its log file after that point - bash's own `wait $pid1 $pid2 ...` only reaps direct children it captured with `$!`, not orphaned descendants. The platform worker's `filestore.put()` archived the log directory immediately afterward, sometimes mid-write.

2. **Shared log directory across concurrent chunks.** Every chunk of the same analysis resolves to the same `model_run_dir` (keyed by `analysis_id` + `run_data_uuid`, not chunk index), and pytools defaulted to writing into a flat `./log` directory regardless of which chunk invoked them. Concurrently-running chunks on the same worker pod were therefore all writing into one shared directory, and each chunk's archive captured a point-in-time snapshot of that shared state rather than its own isolated logs.

3. **Bash's own completeness check ran in the wrong place.** `check_complete()` (which verifies every pytool log reached a "finish" marker) was only ever invoked in the single-script, non-chunked code path. For chunked runs it never ran per-chunk - incompleteness was only ever caught later, once all chunks' logs were merged for the output stage, making the original failure hard to localize.

#### Fix

- `redirect_logging()` now honors an `OASIS_PYTOOLS_LOG_DIR` environment variable override for where pytool log files are written.
- The generated bash script exports `OASIS_PYTOOLS_LOG_DIR=$LOG_DIR` in its header, where `$LOG_DIR` is already chunk-scoped (`log/<process_number>` for a chunk, `log/out` for the output stage). Each chunk/stage now writes its pytool logs into its own directory instead of a directory shared with concurrently-running siblings.
- `check_complete()`'s file search now uses `$LOG_DIR` instead of a hardcoded `log`.

##### Settle-before-archive check (`oasislmf/execution/runner.py`)

- Added `_wait_for_log_writers()`, called after the bash script's tracked process exits and before `run_analysis()`/`run_outputs()` return. It blocks until nothing appears to still be writing under the (now-isolated) log directory, combining two signals since neither is reliable alone:
  - An open-file-handle scan (`psutil`), which is precise and immune to a writer simply pausing between log lines, but can be silently defeated by `AccessDenied` under a restricted container/k8s security context.
  - File `(size, mtime)` stability, which needs no special permissions and works regardless of which process/host is writing, but on its own produces false positives during any legitimate pause between log lines.
  - The scan is limited to same-UID processes to avoid spurious `AccessDenied` noise from unrelated system processes.
  - If the process signal ever becomes unreliable (`AccessDenied`), the required settle window widens substantially rather than trusting an empty result.
  - If nothing settles within the timeout, it raises `OasisException` rather than silently proceeding to archive a possibly-truncated snapshot.

##### Explicit completeness check, moved to Python (`oasislmf/execution/bash.py`, `oasislmf/execution/runner.py`)

- `bash_wrapper()` gained a `run_check_complete` flag, decoupled from `process_number`, controlling whether the bash-side `check_complete()` footer is emitted.
- `genbash()` (used only by the single-script, single-machine `run()` path, where one script owns the entirety of `$LOG_DIR`) keeps `run_check_complete=True` - bash's own centralized check remains appropriate there.
- `run_analysis()` and `run_outputs()` (the distributed, one-script-per-chunk/stage paths) now pass `run_check_complete=False`, removing the bash-side check, and instead call a new Python-side `_check_pytool_logs_complete()` after `_wait_for_log_writers()` succeeds. This is a direct port of bash's "started vs finished" logic (scanning `<tool>_[0-9]*.log` files for a "finish" marker), but running only once writers have already been confirmed settled - so a worker that was merely slow, rather than genuinely lost, isn't misreported as incomplete.

#### Net effect

Each chunk (and the output/merge stage) now: waits for its own writers to genuinely finish, verifies completeness of its own isolated logs immediately, and fails loudly (`OasisException`) if either check doesn't pass - instead of silently archiving a truncated snapshot that would only surface as a failure much later, at final merge time, with the original chunk no longer identifiable.


oasislmf/execution/runner.py
```diff
@@ -184,6 +404,11 @@ def run_analysis(**params):
     monitor.start(proc.pid)
     stdout, _ = proc.communicate()
     monitor.stop()
+    logging.debug("run_analysis: bash script (pid=%s) exited with code %s, checking log completeness in %s",
+                  proc.pid, proc.returncode, monitor_dir)
+    check_start = time.time()
+    _ensure_pytool_logs_complete(monitor_dir)
```
Ha-Ree, line 410: Should this go after the `if proc.returncode != 0:` so you get the error on failure instead of the incomplete logs error?

oasislmf/execution/runner.py
```diff
@@ -1,15 +1,234 @@
...
+    """
+    writers = []
+    fully_inspected = True
+    own_uid = os.getuid()
+    for proc in psutil.process_iter(['pid', 'name', 'uids']):
+        uids = proc.info.get('uids')
+        if uids is not None and uids.real != own_uid:
+            continue
+        try:
+            # DIAGNOSTIC: is this a pytool process at all (by name OR cmdline,
```
Ha-Ree, line 57: Is this diagnostic comment supposed to still be here?

oasislmf/execution/runner.py
```diff
@@ -1,15 +1,234 @@
...
+                    writers.append((proc.pid, proc.info.get('name'), f.path))
+                    break
+        except psutil.AccessDenied:
+            fully_inspected = False
+        except (psutil.NoSuchProcess, psutil.ZombieProcess):
+            continue
+    return writers, fully_inspected
+
+
+def _wait_for_log_writers(log_dir, timeout=30, poll_interval=0.5, stable_checks=2, degraded_stable_seconds=10.0):
```
Ha-Ree, line 91: Just to make sure: no error raised but description of pr says
`If nothing settles within the timeout, it raises OasisException rather than silently proceeding to archive a possibly-truncated snapshot.`

oasislmf/execution/runner.py
```diff
@@ -1,15 +1,234 @@
...
+            # interpreter, e.g. 'python3', not the console-script name), and
+            # if so, what raw path does psutil report for its open files vs.
+            # log_dir? Helps tell "no real writer left" apart from "the path
+            # string just doesn't match" (symlinks/bind-mounts resolved
+            # differently by /proc than by os.path.abspath). Remove once
+            # open_writers is confirmed to ever populate on the real cluster.
+            is_pytool = proc.info.get('name') in MONITORED_TOOLS
+            if not is_pytool:
+                try:
+                    is_pytool = any(tool in ' '.join(proc.cmdline()) for tool in MONITORED_TOOLS)
```
Ha-Ree, line 68: Maybe worth using stephane's _TOOL_RE in resource_monitor (already also refers to MONITORED_TOOLS)

Ha-Ree (changes requested):

PR #1836, opened 2026-01-16
Feature Remove ktools

sambles:
### Refactored package and removed C++ binaries 
* Moved to using `pyproject.toml`  over `setup.py`  (and updated CI to match)
* Dropped non-ORD outputs as options
* Pytools is now the main kernel option, ktools binaries are not built into the package 
* Removed params that enabled pytools (now the only option) 
* Removed 'legacy_stream' option 
* Removed / Updated GenBash tests 
* renamed parameters,  anything using *ktools-<something>* is now *kernel-<something>*, so  from `--ktools-alloc-rule-il` to  `--kernel-alloc-rule-il`

- [x] remember to revert a2f60a0154c2ac44c5463c69dd7aaa79c889706f and dc409812f9e153c77d8f1ec6c350683d9b63119c

### Manual Piwind checks 
* https://github.com/OasisLMF/OasisLMF/actions/runs/21664914183/job/62457910755
* https://github.com/OasisLMF/OasisPiWind/actions/runs/21444146908

### Linked to 
* https://github.com/OasisLMF/OasisPiWind/pull/203
* https://github.com/OasisLMF/ODS_Tools/pull/219
* https://github.com/OasisLMF/OasisPlatform/pull/1317

> ~Difference in ORD output when switching to all pytools for a piwind run~  fixed with https://github.com/OasisLMF/OasisLMF/pull/1859
[defaults-run-ORD.zip](https://github.com/user-attachments/files/25071369/defaults-run-ORD.zip)
[pytools-only-run-ORD.zip](https://github.com/user-attachments/files/25071377/pytools-only-run-ORD.zip)
[update-pytools.zip - after fix](https://github.com/user-attachments/files/25130093/update-pytools.zip)

#### Manual Piwind check
CI failing because https://github.com/OasisLMF/OasisPlatform/pull/1317 not released yet
~https://github.com/OasisLMF/OasisPlatform/actions/runs/21796276874~

https://github.com/OasisLMF/OasisPlatform/actions/runs/21796567591

PR #2164, opened 2026-09-17
FileNotFoundError on GenerateLossesDir.run

### LINKED ISSUE #2162: GenerateLossesDir.run crashes with FileNotFoundError
Title: GenerateLossesDir.run crashes with FileNotFoundError on intermediate RI layer directories when net RI output is requested (only final inuring priority)

Environment

oasislmf 2.5.6
Python 3.12
Linux (Ubuntu 5.15 kernel)
Complex model run (oasislmf model run), ri_output: true, ri_summaries set, no ri_inuring_priorities key in analysis_settings.json
Summary

GenerateLossesDir.run in oasislmf/computation/generate/losses.py crashes with a FileNotFoundError when building summarypy structures for reinsurance, because it iterates over every physical RI_N input directory, but generate_summaryxref_files() in oasislmf/preparation/summaries.py only writes an fmsummaryxref file into the RI directories that correspond to actual reinsurance "output levels" (i.e. the last layer of each requested inuring priority — by default just the final one, when ri_inuring_priorities is not set in analysis_settings.json).

Traceback

  ...
  File ".../oasislmf/computation/generate/losses.py", line 399, in run
    create_summary_object_file(summary_dir, runtype)
  File ".../oasislmf/pytools/summary/manager.py", line 65, in create_summary_object_file
    summary_objects = get_summary_object(static_path, run_type)
  File ".../oasislmf/pytools/summary/manager.py", line 92, in get_summary_object
    summary_xref = load_as_ndarray(static_path, 'fmsummaryxref', fm_summary_xref_dtype)
  File ".../oasislmf/pytools/common/data.py", line 413, in load_as_ndarray
    with open(filepath) as file_in:
FileNotFoundError: [Errno 2] No such file or directory: '.../input/RI_4/fmsummaryxref.csv'
Root cause

In this run there are 13 RI layers (RI_1...RI_13) across 4 inuring priorities, mapped via ri_inuring_priority_output_levels.json as:

{"1": 4, "2": 7, "3": 10, "4": 13}
Since analysis_settings.json doesn't set ri_inuring_priorities, generate_summaryxref_files() (summaries.py, around line 755) only adds the max OED priority (4) to ri_inuring_priorities_oed, so ri_output_levels = {13} — only RI_13 gets an fmsummaryxref.bin/.csv written.

However, GenerateLossesDir.run (losses.py, around line 388–399) builds summary_dirs for RUNTYPE_REINSURANCE_LOSS from ri_dirs, which is just every directory on disk matching RI_\d+ (losses.py, around line 248) — all 13 layers, with no filtering by output level. It then calls create_summary_object_file() on each, which unconditionally requires an fmsummaryxref file to exist, crashing on the first intermediate (non-output-level) RI layer it hits.

Notably, oasislmf/execution/bash.py already has the correct concept of this distinction (get_ri_inuring_priorities(), used throughout to restrict summarypy/output commands to genuine output-level RI layers) — so the rest of the execution pipeline already treats intermediate RI layers correctly. GenerateLossesDir.run's structure-creation step appears to be the one place that wasn't updated to respect this.

Suggested fix

In GenerateLossesDir.run, filter ri_dirs to only the directories that actually have an fmsummaryxref.bin/.csv (or equivalently, recompute the same ri_output_levels logic used in generate_summaryxref_files) before building summary_dirs for RUNTYPE_REINSURANCE_LOSS:

if runtype == RUNTYPE_REINSURANCE_LOSS:
    summary_dirs = [
        os.path.join(self.model_run_dir, 'input', ri_sub_dir) for ri_sub_dir in ri_dirs
        if os.path.isfile(os.path.join(self.model_run_dir, 'input', ri_sub_dir, 'fmsummaryxref.bin'))
        or os.path.isfile(os.path.join(self.model_run_dir, 'input', ri_sub_dir, 'fmsummaryxref.csv'))
    ]
else:
    summary_dirs = [os.path.join(self.model_run_dir, 'input')]
Applying this locally against 2.5.6 resolves the crash for our regression test (13 RI layers / 4 inuring priorities, net output on the final priority only).

Steps to reproduce

Run a complex model with ri_output: true and ri_summaries configured, and an OED reinsurance structure with more than one RI layer per inuring priority (e.g. LOC/POL/ACC/SEL sub-layers), without setting ri_inuring_priorities in analysis_settings.json.
oasislmf model run fails during generate.losses.run → GenerateLossesDir.run with FileNotFoundError: ... fmsummaryxref.csv on the first non-output-level RI directory.

Ha-Ree:
Fixes #2162.

## Problem

`oasislmf model run` crashes with `FileNotFoundError: .../input/RI_N/fmsummaryxref.csv` when net RI
output is requested on a reinsurance structure that has more RI layers than reinsurance output
levels.

Two places disagreed about which `RI_N` directories are output levels:

- `generate_summaryxref_files` (`preparation/summaries.py`) writes `fmsummaryxref` only to the
  output-level layers — the last layer of each requested inuring priority. With no
  `ri_inuring_priorities` in the analysis settings, that is just the final one.
- `GenerateLossesDir.run` built `summary_dirs` from **every** directory on disk matching `RI_\d+`
  and called `create_summary_object_file` on each, which requires the xref unconditionally. It
  crashed on the first intermediate layer.

`execution/bash.py` already had the right concept — `get_ri_inuring_priorities()` restricts
summarypy and output commands to genuine output levels, and `num_reinsurance_iterations`
(the layer count) equals the max priority's output level, so bash.py and summaries.py agree.
`GenerateLossesDir.run` was the only place that had not been updated.

## Scope

Broader than the issue reports: multiple inuring priorities are not required. **Two RI layers at a
single priority is enough** — every layer below the final output level crashes. Two treaties at
`InuringPriority 1` with different `RiskLevel` reproduce it.

This is a 2.5.0 regression rather than one from #2006. The loop predates #2006 but sat behind
`--summarypy`, which defaulted to `False`. #1836 ("Remove ktools") made it unconditional, so the
crash became reachable by default in 2.5.0 and has shipped in 2.5.0–2.5.7.

### Why CI never caught it

Every RI fixture (`RI_RUN_SETTINGS`, `RI_ALL_OUTPUT_SETTINGS`) also sets `rl_output: True`, and
gross RL output takes the `ri_output_levels = ri_layers` branch, writing an xref into *all* RI
directories. On top of that `MIN_INF` and PiWind's `SourceReinsInfoOEDPiWind.csv` each declare a
single treaty, so there were no intermediate layers to trip over. The bug was invisible to the
whole suite.

## Changes

**`computation/generate/losses.py`**

- Filter RI summary directories to those that actually hold an `fmsummaryxref`, rather than
  re-deriving the output-level rule in a third place. Safe because this call is only a pre-compute:
  `load_summary_object` builds the structure on demand, and summarypy is never pointed at a
  non-output-level layer.
- Dedupe `ri_dirs`. It concatenates `os.listdir(oasis_files_dir) + os.listdir(model_run_dir)`, so
  re-running into an existing run directory built each layer's financial structure twice. Wasted
  work only — the layer count that reaches `num_reinsurance_iterations` comes from `ri_layers.json`
  via `_get_num_ri_layers()`, never from `len(ri_dirs)`.

**Tests**

- `MULTI_LAYER_INF` / `MULTI_LAYER_SCP`: an RI structure with more layers than output levels.
- `test_losses__run_ri__net_only__intermediate_layers_skipped`, driven by `RI_AAL_SETTINGS`
  (`ri_output` without `rl_output`) so it exercises the net-only path the existing fixtures mask.

## Testing

- The new test fails on unpatched `main` with the issue's exact traceback
  (`FileNotFoundError: .../input/RI_1/fmsummaryxref.csv`) and passes with the fix.
- `pytest tests/computation/test_generate_losses.py -k "run_ri or run_il or run_gul"` — 7 passed.
- `test_losses__chucked_workflow` fails locally both with and without this change
  (`oasis_exec_monitor.sh: command not found`) — environmental, unrelated.

## Follow-up, not addressed here

The pre-compute loop omits `RUNTYPE_REINSURANCE_GROSS_LOSS`, so an rl-only run builds its summary
structures lazily at run time. Correct, just uncached — and it cannot simply be added to the loop,
since `get_summary_object` rejects `'rl'` (bash.py passes `-t ri` for that runtype).


oasislmf/computation/generate/losses.py
```diff
@@ -381,7 +384,15 @@ def run(self):
                 summary_sets_id = np.sort([summary['id'] for summary in summaries if 'id' in summary])
                 if summary_sets_id.shape[0]:
                     if runtype == RUNTYPE_REINSURANCE_LOSS:
-                        summary_dirs = [os.path.join(self.model_run_dir, 'input', ri_sub_dir) for ri_sub_dir in ri_dirs]
+                        # Only the RI layers that are reinsurance output levels get an
+                        # fmsummaryxref; intermediate layers are computed but never summarised.
+                        summary_dirs = [
```
SkylordA: The `isfile` filter can't tell "not an output level this run" apart from a genuine upstream failure (e.g. `generate_summaryxref_files` erroring partway) or stale run-dir state — it silently skips pre-building the summary object either way, where the old code would crash loudly.

Also worth noting: `summaries.py:683-738` already knows exactly which RI dirs it wrote xref files into, so re-deriving that set here via disk probes risks the two computations quietly drifting apart over time. Could this reuse that set directly instead of inferring it from `isfile`?

oasislmf/computation/generate/losses.py
```diff
@@ -381,7 +384,15 @@ def run(self):
...
+                            ri_dir for ri_dir in (
+                                os.path.join(self.model_run_dir, 'input', ri_sub_dir) for ri_sub_dir in ri_dirs
+                            )
+                            if any(os.path.isfile(os.path.join(ri_dir, f"{SUMMARY_OUTPUT['il']}.{ext}"))
```
SkylordA: Minor: checking both `.bin` and `.csv` here is redundant (`.bin` is always written unconditionally by `generate_summaryxref_files`, `.csv` only conditionally), and the nested generator-in-list-comprehension is a bit awkward to read/extend.

oasislmf/computation/generate/losses.py
Ha-Ree: Agreed on both counts — swapped the disk probe for a shared derivation in `e3d13e501`.

`generate_summaryxref_files` now resolves its output levels through a new `get_ri_summaryxref_dirs(input_dir, analysis_settings, all_layers=...)` in `summaries.py`, and `GenerateLossesDir.run` calls the same function, so there is one definition rather than two that can drift. `all_layers=True` covers the gross-RL case where every layer gets an xref.

That also restores the loud failure: the pre-compute is now pointed at exactly the dirs that are supposed to hold an xref, so a partial `generate_summaryxref_files` or stale run-dir state hits `create_summary_object_file` and raises, instead of being filtered out as "not an output level".

oasislmf/computation/generate/losses.py
Ha-Ree: Correct, and the comprehension is gone entirely in `e3d13e501` — the RI directories now come back from `get_ri_summaryxref_dirs` rather than being inferred from the filesystem, so there is no extension check left to get wrong.

oasislmf/computation/generate/losses.py
```diff
@@ -381,7 +383,11 @@ def run(self):
                 summary_sets_id = np.sort([summary['id'] for summary in summaries if 'id' in summary])
                 if summary_sets_id.shape[0]:
                     if runtype == RUNTYPE_REINSURANCE_LOSS:
-                        summary_dirs = [os.path.join(self.model_run_dir, 'input', ri_sub_dir) for ri_sub_dir in ri_dirs]
+                        # Intermediate RI layers are computed but never summarised, so only the
+                        # output levels hold an fmsummaryxref.
+                        summary_dirs = get_ri_summaryxref_dirs(
```
SkylordA, line 388: `all_layers` isn't passed here, so this always resolves the net-output-level set. But `generate_summaryxref_files` (summaries.py:756) passes `all_layers=bool(rl_summaries)` — when `rl_output`+`rl_summaries` are also requested, it writes xref files to every RI layer, not just net output levels. So the write side and this pre-build call can resolve different dir sets again, just via a differing argument instead of differing logic.

Not a crash — `load_summary_object` falls back to an on-demand rebuild when the pre-build is missing — but it silently skips pre-building for the extra gross-RL layers, and nothing tests `ri_output` + `rl_summaries` together.

Should this pass `all_layers=bool(self.settings.get('rl_output') and self.settings.get('rl_summaries'))` to match?

SkylordA (approved):

vinulw (approved):

PR #2166, opened 2026-09-18
feat(lookup): add geog_lookup builtin for OED GeogScheme/GeogName (#2144)

### LINKED ISSUE #2144: Support for GeogScheme/GeogName columns in lookup builtins
The lookup builtin `merge` can be used to assign an areaperil id based on a static column such as `PostalCode` or `AreaCode` by simply providing a table with that column plus an `areaperil_id` column. However, it does not handle the case where the location information is in a `GeogNameX` column for a particular `GeogScheme`. For instance, a model might want to assign an area peril based on what3words but that value could be in any of the 30 `GeogNameX` columns so we cannot provide data in the model that is guaranteed to join correctly. To find the right column, we need to find the corresponding `GeogScheme` with the value `W3W`. I don't think that's possible with the existing builtin functions and requires a custom function.

There should be support for this situation in the lookup builtins since it's a key part of OED. Perhaps a more specialised "OED join" function that would take either an OED column name or a GeogScheme name.

PeterCotroneo:
## Summary

Adds a `geog_lookup` builtin step so the built-in lookup can assign an id from an OED
scheme-tagged location value. OED stores such values in a pair of columns
`GeogScheme{N}` / `GeogName{N}` (N = 1..30); the scheme of interest (e.g. `W3W`) can sit
in any slot, so there is no fixed column the existing `merge` step can join on. Closes #2144.

## What it does

`geog_lookup` scans the `GeogScheme{N}`/`GeogName{N}` pairs and copies the `GeogName{N}`
whose paired `GeogScheme{N}` matches the target scheme into a single output column (lowest
matching slot wins). A subsequent `merge` step then joins on that column. Example config:

```jsonc
"step_definition": {
  "resolve_w3w": {
    "type": "geog_lookup",
    "columns": ["GeogScheme1","GeogName1","GeogScheme2","GeogName2"],
    "parameters": { "geog_scheme": "W3W", "output_column": "w3w", "slots": 2 }
  },
  "areaperil": {
    "type": "merge",
    "columns": ["w3w"],
    "parameters": { "file_path": "%%KEYS_DATA_PATH%%/w3w_areaperil.csv", "id_columns": ["area_peril_id"] }
  }
},
"strategy": ["resolve_w3w", "areaperil"]
```

Options: `slots` (default 30), `case_insensitive` (default True, also trims whitespace),
`on_missing` ("null" leaves unmatched rows null, "error" raises). Empty/sparse scheme slots
(the common OED shape) are handled - the mask is NA-safe.

## Design note

Implemented as a small composable step that feeds the existing `merge`, rather than a bundled
"OED join" that resolves and joins in one step. This keeps `merge` as the single join
implementation and lets the resolved column feed any downstream step. Happy to switch to a
bundled `oed_merge` step (as the issue suggests) if preferred - it would wrap resolve-then-merge.

## Also included

A guard on `merge` for the no-shared-column case, which previously surfaced as an opaque
pandas `MergeError`. It now raises a clear `OasisException` pointing at `geog_lookup` - the
exact failure a user hits today when naively trying to merge on a scheme value.

## Tests

Ten tests in `tests/lookup/test_builtin.py`: slot-independence, first-match-wins, case/whitespace
matching, null vs error miss handling, sparse/NaN slots, sparse-absent slots, invalid-option
rejection, an end-to-end `geog_lookup` -> `merge` producing `area_peril_id`, and the merge guard.
Full `test_builtin.py` passes (59); `ruff check` clean.


carlfischerjba:
Elegant solution. Thanks.

oasislmf/lookup/builtin.py
```diff
@@ -1091,10 +1094,70 @@ def build_merge(self, file_path, id_columns=[], file_type='csv', **kwargs):
...
+            slots (int): highest ``GeogScheme{N}``/``GeogName{N}`` index to scan (default 30).
+            case_insensitive (bool): match the scheme ignoring case and surrounding whitespace.
+            on_missing (str): ``'null'`` to leave unmatched rows null, ``'error'`` to raise.
+        """
+        if on_missing not in ('null', 'error'):
+            raise OasisException(f"build_geog_lookup: on_missing must be 'null' or 'error', got {on_missing!r}")
+        target = geog_scheme.upper() if case_insensitive else geog_scheme
+
+        def geog_lookup(locations):
+            resolved = pd.Series(pd.NA, index=locations.index, dtype='object')
```
Ha-Ree: `process_locations` renames location columns to the spelling used in the step's columns list but we resolve columns regardless of capitalisation for ease of use (see `build_merge`'s `rename_map` above) which can be done here with
```
def geog_lookup(locations):
    col_map = {col.lower(): col for col in locations.columns}
....
    scheme_col, name_col = col_map.get(f"geogscheme{n}"), col_map.get(f"geogname{n}")
    if scheme_col is None or name_col is None:
        continue
```

Ha-Ree (changes requested):

oasislmf/lookup/builtin.py
PeterCotroneo: Cheers, @Ha-Ree, good catch. Switched to your lowercase column map and added a regression test. Fixed in 645b3e2.

Ha-Ree (approved):

